PR #34, opened 2018-03-29
Initial Structure

dpanayotov:
This is the initial structure of the Service Manager Core.
Currently /vendor is not submitted and `dep ensure` is required to build the project.


api/api_test.go
```diff
@@ -0,0 +1,76 @@
...
+					controllers = append(controllers, nil)
+					defaultAPI.RegisterControllers(controllers...)
+				}
+				Expect(nilControllerInSlice).To(Panic())
+			})
+		})
+
+		Context("With no brokers registered", func() {
+			It("Should have only one broker", func() {
+				Expect(len(defaultAPI.Controllers())).To(Equal(1))
```
KirilKabakchiev: I'd suggest to not check for hard-coded values in tests (like in this case that exactly 1 is present) - for example it might be cleaner to get the count in the begining of the test, add one item, then get the new count and validate that it has increased by one.

You may look at https://semaphoreci.com/community/tutorials/getting-started-with-bdd-in-go-using-ginkgo and search for `originalItemCount`

api/api_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"testing"
+
+	"github.com/Peripli/service-manager/rest"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestApi(t *testing.T) {
```
KirilKabakchiev: should spec, tests and mocks be in seperate files?

rest/errors_test.go
```diff
@@ -0,0 +1,167 @@
...
+	})
+
+	Describe("Error Handler Func", func() {
+
+		var returnedData error
+
+		var mockedAPIHandler rest.APIHandler
+
+		BeforeEach(func() {
+			returnedData = nil
```
KirilKabakchiev: do we need this?

rest/errors_test.go
```diff
@@ -0,0 +1,167 @@
...
+				Expect(string(mockedWriter.data)).To(BeEmpty())
+			})
+		})
+
+		Context("With API Handler returning an error", func() {
+			BeforeEach(func() {
+				returnedData = testError
+			})
+
+			It("Should write to Response Writer", func() {
```
KirilKabakchiev: this can be inlined in the test

rest/errors_test.go
```diff
@@ -0,0 +1,167 @@
+/*
+ *    Copyright 2018 The Service Manager Authors
```
KirilKabakchiev: do we need copyright on tests?

dpanayotov: don't know.. doesn't do any harm

api/api_test.go
dpanayotov: maybe.. seemed unnecessary to have a separate file for just this thing. When there are multiple tests in a directory or we have common methods, we can move it

PR #37, opened 2018-04-11
[WORK-IN-PROGREESS] Osb api + errors/env/db refactoring

KirilKabakchiev:
Moved to https://github.com/Peripli/service-manager/pull/46


storage/interfaces.go
```diff
@@ -16,11 +16,30 @@
...
+}
+
+type Config interface {
+	createFunc() Storage
+}
+
+// Broker interface for Broker db operations
+//go:generate counterfeiter . Broker
+type Broker interface {
+	Create(broker *types.Broker) error
```
dpanayotov: these should probably take a context

storage/registry.go
```diff
@@ -18,66 +18,57 @@ package storage
...
+//Use returning the storage seems more natural rather then doing .Get; .Get() introduces a hidden dependency as well)
+//This way we kind of provide a wrapper over the sqlx.db by keeping the same flow of usage while keeping an abstraction for Storage possible
+// Wrapping as follows
+// - postgres impl of storage _ imports the driver, we _import the postgres storage's package
+// Registering a storage "wraps" registering a driver and allows us to "use" the storage that wraps abstract opening a storage (in posgres it opens sqlx.db)
+// The idea is to allow abstraction of Storage while keeping the sqlx.db flow of usage as close as possible
+// uri is passed as argument as we don't want to obtain env variables in random places of the code as the config might not be provided as env vars)
+// It's better to abstract away the environment loading and use it to build a configuration that is passed to the server. The server than takes care
+// to split the configuration and pass segments to the components it initializes (to storage the server will pass uri)
+func Use(name string, uri string, ctx context.Context) (Storage, error) {
```
dpanayotov: I'm ok with this way. It would work with postgres, mongo, etc. but etcd can take more than 1 uri for different nodes
If we go the db/sql way we should register something that would provide you a Storage. Then on Use(..) you'd have storageProvider.Provide(uri?)  which would give you the new Storage. Not sure if we'd benefit from separating the opening from the closing of the storage

server/server.go
```diff
@@ -19,64 +19,82 @@ package server
...
-		ReadTimeout:  server.Configuration.RequestTimeout(),
+		Addr:         server.Configuration.Address,
+		WriteTimeout: server.Configuration.RequestTimeout,
+		ReadTimeout:  server.Configuration.RequestTimeout,
 	}
-	startServer(handler, server.Configuration.ShutdownTimeout())
+	startServer(handler, server.Configuration.ShutdownTimeout, ctx)
+}
+
+func moveRoutes(prefix string, fromRouter *mux.Router, toRouter *mux.Router) error {
```
dpanayotov: Couldn't find better way to embed another router's handlers at a path-prefix.

storage/registry.go
KirilKabakchiev: Right now
We regisrter a Storage that during registration is a "wrapper" for driver registration
Then we Use the Storage which during "usage/opening" is wrapper for sql/db opening

So we wrap both the "driver" and the "db" with our storage which seems fine to me - 

if we see a benefit we may split it so that we "wrap" the driver with storage and the "db.opening" with provider - but right now I can't think of a decent use case?

storage/interfaces.go
KirilKabakchiev: agreed, I was under the impression that sqlx doesn't support context but then I found https://github.com/jmoiron/sqlx/pull/270/files

server/server.go
KirilKabakchiev: maybe we can pick a less-weirdly sounding name

PR #40, opened 2018-04-12
Initial Storage

dpanayotov:
This PR is just to base REST APIs on something. Final storage structure will come aftewards


dpanayotov:
Initial Storage is part of #35 Further storage development will be part of #41

PR #46, opened 2018-04-16
Osb API

KirilKabakchiev:


api/osb/logic.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func (b *BusinessLogic) osbClient(request *http.Request) (osbc.Client, error) {
+	vars := mux.Vars(request)
+	brokerID, ok := vars[BrokerIDPathParam]
+	logrus.Debugf("Obtained path parameter [brokerID = %s] from mux vars", brokerID)
+	if !ok {
+		return nil, fmt.Errorf("Error creating OSB client: brokerID path parameter not found")
+	}
+	broker, err := b.brokerStorage.Find(request.Context(), brokerID)
```
pankrator: Is the context object really used?

KirilKabakchiev: no it's not mandatory - the idea was to make the interfaces pass it from now in case we need them in the future (so we don't have to change all the interfaces later).

api/osb/logic.go
```diff
@@ -0,0 +1,221 @@
...
+		return nil, err
+	}
+
+	return &broker.UpdateInstanceResponse{
+		UpdateInstanceResponse: *response,
+	}, nil
+}
+
+// ValidateBrokerAPIVersion implements pmorie/osb-broker-lib/pkg/broker/Interface.ValidateBrokerAPIVersion by
+// proxying the method to the underlying implementation to an underlying service broker
```
dpanayotov: it does not

api/osb/controller.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (c *Controller) osbHandler() http.Handler {
+	businessLogic := NewBusinessLogic(osbc.NewClient, c.BrokerStorage)
+
+	reg := prom.NewRegistry()
+	osbMetrics := metrics.New()
+	reg.MustRegister(osbMetrics)
+
+	api, err := osbrest.NewAPISurface(businessLogic, osbMetrics)
+	if err != nil {
```
dpanayotov: Should we check the error here? Even though the function declaration returns, the implementation doesn't. We might ignore it or check it just to be safe for future lib updates

KirilKabakchiev: i was thinking about this as well - they left it to return an error so that they don't have to change the method api in future updates of the library, in case the constructor becomes more complex.

I'd rather have the check (even though the code in the if statement is not testable) and be safe as I doubt we will be checking such minor details when upgrading dependency versions

api/osb/controller.go
```diff
@@ -0,0 +1,66 @@
...
+
+var _ rest.Controller = &Controller{}
+
+// Routes implements rest.Controller.Routes by providing the routes for the OSB API
+func (c *Controller) Routes() []rest.Route {
+	return []rest.Route{
+		{
+			Endpoint: rest.Endpoint{
+				Method: rest.AllMethods,
+				Path:   "/osb/{brokerID}",
```
dpanayotov: maybe put API_PREFIX somewhere in the api package and append it here (v1 for example)

PR #49, opened 2018-04-20
Safe environment configuration

NickyMateev:
# Safe environment configuration

Currently if any of the expected environment variables is omitted from the **application.yml** file, or if this file doesn't exist at all, the application startup will crash.

This PR suggests changes for ensuring safe loading of the environment configuration even if **application.yml** is not configured properly or is missing.


env/env.go
```diff
@@ -80,6 +80,10 @@ func (v *viperEnv) Set(key string, value interface{}) {
 	v.Viper.Set(key, value)
 }
 
+func (v *viperEnv) IsSet(key string) bool {
+	return v.Viper.IsSet(key)
```
KirilKabakchiev: There is a slight drawback to this approach - viper.isSet does a viper.find() which is not the fastest and it is already done during unmarshaling. 

what we do right now in main is 
```
config, err := server.NewConfiguration(env.Default())
```

env.Default uses application.yml file, so currently it's our "choice" to say that we want the file created. We can always do env.New(..) instead of default and do not provide configfile there if we are certain we will not need the file.

Ideally I would propose to drop IsSet method and add a check in viperEnv.Load() to check if the file exists and to SetConfigName and SetConfigPath only if it exists. Log proper stuff if it is missing

server/config.go
```diff
@@ -93,22 +94,22 @@ func NewConfiguration(env Environment) (*Config, error) {
 		return nil, err
 	}
 
-	if configSettings.Server.Port != 0 {
+	if env.IsSet("server.port") && configSettings.Server.Port != 0 {
```
KirilKabakchiev: I would suggest droping the IsSet method and change 
lines 92-96 (the configSettings in NewConfiguration) as follows : 
```
configSettings := &Settings{
		Server: &AppSettings{},
		Db:     &DbSettings{},
		Log:    &LogSettings{},
	}
```

The whole purpose of doing unmarshal is not not hardcode stuff like "server.port" in the code but use struct fields only

KirilKabakchiev:
can you please make the  diff show only newly added things that come from this change? It would make it a lot easier for review :)

PR #57, opened 2018-05-04
K8S build and deploy guide

pankrator:


deployment/k8s/sm-secret.yml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name:  sm-postgre-service-name
```
KirilKabakchiev: how about replacing the secret with the db name and the secret with the db credentials with one that contains the db uri and then set and env variable in the sm-deployment.yml with name SM_DB_URI and value the val from the secret. And then you don't need  the file cmd/k8s/env/env.go

deployment/k8s/postgre-pod.yml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
dpanayotov: use https://github.com/kubernetes/charts/tree/master/stable/postgresql

dpanayotov: On a second thought - should we even give the option to use a postgres in k8s? I mean it would be nice, but etcd is what sm should be using ultimately.

KirilKabakchiev: I think for now we could allow it so that we can validate our deployment works there.

When we have time we may implement a db layer and “queries” for etcd and swap the postgres from the sm helm installation

PR #59, opened 2018-05-14
Info Endpoint

dpanayotov:
Initial implementation of https://github.com/Peripli/specification/pull/27


sm/sm.go
```diff
@@ -45,7 +46,9 @@ func NewServer(ctx context.Context, serverEnv server.Environment) (*server.Serve
 	if err != nil {
 		return nil, fmt.Errorf("Error using storage: %v", err)
 	}
+
 	defaultAPI := api.Default(storage)
+	defaultAPI.RegisterControllers(info.NewController(serverEnv))
```
georgifarashev: Shouldn't this be returned by *api.Default*?

georgifarashev:
Can this be tested with a simple integration test?

dpanayotov:
Not sure if inside the controller is the place to validate that `token_issuer_url` actually has openid provided configuration, but generally a validation should be present

sm/sm.go
dpanayotov: I didn't want to overload the api with both a storage and an environment. But yes, you're correct that this is a core part of the API and should be in the default. No going around passing the environment for now

Gopkg.toml
```diff
@@ -6,7 +6,7 @@
 
 [[constraint]]
   name = "github.com/gorilla/mux"
-  version = "1.6.1"
+  version = "=1.6.1"
```
dpanayotov: Backwards-incompatible change is introduced in mux 1.6.2 - https://github.com/gorilla/mux/pull/338 
which prevents the router from starting due to https://github.com/pmorie/osb-broker-lib/blob/master/pkg/server/server.go#L63

We should either stick to 1.6.1 or ignore the error for the `healthz` endpoint

KirilKabakchiev: can you add this as a comment above the constraint in the toml file?

PR #60, opened 2018-05-21
Authenticator

dpanayotov:
This provides just the mechanism to decide whether authentication is valid or not. Filter implementation is not in the scope of this issue


authentication/oidc/odic.go
```diff
@@ -0,0 +1,109 @@
...
+
+	var p providerJSON
+	err = json.Unmarshal(body, &p)
+	if err != nil {
+		return nil, err
+	}
+
+	keySet := oidc.NewRemoteKeySet(ctx, p.JWKSURL)
+	cfg := &oidc.Config{
+		ClientID: options.ClientID,
```
pankrator: Don't we need client_secret also?

authentication/oidc/oidc.go
```diff
@@ -0,0 +1,126 @@
...
+	req, err := http.NewRequest(http.MethodGet, wellKnown, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var readConfigFunc DoRequestFunc
+	if options.ReadConfigurationFunc != nil {
+		readConfigFunc = options.ReadConfigurationFunc
+	} else {
+		readConfigFunc = http.DefaultClient.Do
```
pankrator: Default client does not have request timeouts set. We should go against the option to use it by default

dpanayotov: I think it's nice to have defaults. However, I strongly agree the default shouldn't be used in production.

pankrator: Actually, if a request with context that has timeouts set is provided it will be sufficient

PR #61, opened 2018-05-31
Add search by multiple names for brokers and platforms

pankrator:


storage/postgres/broker.go
```diff
@@ -96,6 +96,29 @@ func (bs *brokerStorage) GetAll() ([]types.Broker, error) {
 	return brokers, nil
 }
 
+func (bs *brokerStorage) GetByName(names []string) ([]types.Broker, error) {
```
georgifarashev: The code for platforms and brokers seems almost the same. Is it possible to extract the common part? Maybe in the controllers as well if possible.

storage/postgres/platform.go
```diff
@@ -78,6 +78,29 @@ func (ps *platformStorage) GetAll() ([]types.Platform, error) {
...
 
+func (ps *platformStorage) GetByName(names []string) ([]types.Platform, error) {
+	platformDTOs := []Platform{}
+
+	whereClause := ""
+	for index, name := range names {
+		if index != 0 {
+			whereClause += " OR "
+		}
+		whereClause += fmt.Sprintf("name='%s'", name)
```
georgifarashev: Is this SQL statement safe? The input is from the user so maybe using placeholders is a better idea?

dzahariev:
Not needed for now.

PR #71, opened 2018-06-15
Enable Service Manager environment and configuration to work with pflags

KirilKabakchiev:
Currently Service Manager can be configured via configuration file or env variables.
This PR allows service manager initial start up to be configured via pflags/flags as well.
The proiority is manual Set > pflag value > env var > config file > pflag default value.

This PR also simplifies the configuration structure.

- [x] Initial implementation
- [x] Refactoring
- [x] Unit tests
- [x] godoc/linter
- [x] Ready for Review


server/server.go
```diff
@@ -24,40 +24,50 @@ import (
...
 // Run starts the server awaiting for incoming requests
-func (server *Server) Run(ctx context.Context) {
+func (s *Server) Run(ctx context.Context) {
 	handler := &http.Server{
-		Handler:      server.Router,
-		Addr:         server.Configuration.Address,
-		WriteTimeout: server.Configuration.RequestTimeout,
-		ReadTimeout:  server.Configuration.RequestTimeout,
+		Handler:      s.Router,
+		Addr:         s.Config.Host + ":" + strconv.Itoa(s.Config.Port),
```
pankrator: Why not port being a string at first place?

Issue #67, opened 2018-06-13
Add recovery middleware to web request flows


KirilKabakchiev:
https://github.com/urfave/negroni/blob/master/recovery.go 
https://github.com/codahale/http-handlers/blob/master/recovery/recovery.go

PR #76, opened 2018-06-20
Filters & Plugins

### LINKED ISSUE #82: Filters and plugins need to have names/identifiers for debug purposes
During filters registration it is shown what number of filters matched a given endpoint (method + path).

It will be more useful to have filter's name/identifier instead of how much of them were registered.
Even better will be to know which filter did not match any endpoint.

dotchev:
TODO:
- [x] Finish GoDoc (should fix metalinter issues)
- [x] Provide plugin example

Fixes #82


pkg/filter/filter.go
```diff
@@ -0,0 +1,67 @@
...
+	// Methods match route method.
+	// If nil, matches any method.
+	//
+	// NOTE: This will work as long as each route handles a single method.
+	// If a route handles multiple methods,
+	// the filter could be called for methods which are not listed here.
+	Methods []string
+
+	// PathPattern matches route path (as registered in mux).
+	// This is a glob pattern so you can use * and **.
```
dzahariev: global

rest/api.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Filters contains the registered filters
+	Filters []filter.Filter
+}
+
+// RegisterControllers registers a set of controllers
+func (api *API) RegisterControllers(controllers ...Controller) {
+	for _, controller := range controllers {
+		if controller == nil {
+			logrus.Panicln("Cannot add nil controllers")
```
dzahariev: Probably you should call "continue" here to avoid nil controller to be added.

rest/api.go
```diff
@@ -0,0 +1,80 @@
...
+// RegisterFilters registers a set of filters
+func (api *API) RegisterFilters(filters ...filter.Filter) {
+	api.Filters = append(api.Filters, filters...)
+}
+
+// RegisterPlugins registers a set of plugins
+func (api *API) RegisterPlugins(plugins ...plugin.Plugin) {
+	for _, plug := range plugins {
+		if plug == nil {
+			logrus.Panicln("Cannot add nil plugins")
```
dzahariev: The same here. 
If this plugin is nil, it will be logged, but will get a NPE later on line 51

docs/plugins.md
```diff
@@ -0,0 +1,109 @@
+# Plugins
+
+Plugins provide means to intercept different OSB calls (provision, deprovision, bind, etc.).
```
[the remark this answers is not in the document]
pankrator: Currently it is the OSB calls. Filters can intercept all calls

pkg/filter/filter.go
pankrator: ?

rest/api.go
pankrator: `Panicnl` will panic and stop the server.

rest/api.go
pankrator: `Panicnl` will panic and stop the server.

rest/api.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		match := false
+		register := func(method string, pathPattern string, middleware filter.Middleware) {
```
dpanayotov: extract register func to return always true and set match = register(...) everywhere

pkg/filter/filter.go
```diff
@@ -0,0 +1,67 @@
...
+// It can also terminate the request processing by not invoking next.
+type Middleware func(req *Request, next Handler) (*Response, error)
+
+// RouteMatcher defines the criteria to match against registered routes.
+//
+// NOTE: This does not match against HTTP requests at runtime but registered
+// routes at startup. This improves performance but has some limitations.
+type RouteMatcher struct {
+	// Methods match route method.
+	// If nil, matches any method.
```
dpanayotov: Not sure if nil is good match. I'd rather have "*" as a match because then you explicitly state you want all methods. Otherwise you could've forgot to put any methods altogether

rest/filters.go
```diff
@@ -0,0 +1,43 @@
...
+// MatchFilters returns all filters that mach the provided endpoint's path and method
+func MatchFilters(endpoint *Endpoint, filters []filter.Filter) []filter.Filter {
+	matches := []filter.Filter{}
+	for _, filter := range filters {
+		if matchPath(endpoint.Path, filter.PathPattern) &&
+			matchMethod(endpoint.Method, filter.Methods) {
+			matches = append(matches, filter)
+		}
+	}
+	logrus.Debugf("%d filters for endpoint %v", len(matches), endpoint)
```
dpanayotov: Count is not very useful here. The filter names would be more useful for debugging

pkg/filter/filter.go
```diff
@@ -0,0 +1,67 @@
...
+	//
+	// NOTE: This will work as long as each route handles a single method.
+	// If a route handles multiple methods,
+	// the filter could be called for methods which are not listed here.
+	Methods []string
+
+	// PathPattern matches route path (as registered in mux).
+	// This is a glob pattern so you can use * and **.
+	// ** matches also / but * does not.
+	// If empty, matches any path.
```
dpanayotov: Again, empty should be treated as invalid IMO.

rest/filters.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	return pat.Match(endpointPath) || pat.Match(endpointPath+"/")
+}
+
+func matchMethod(method string, methods []string) bool {
+	if methods == nil {
+		return true
+	}
+	for _, m := range methods {
+		if m == method {
```
dpanayotov: lowercase both

rest/util.go
```diff
@@ -35,16 +34,24 @@ func SendJSON(writer http.ResponseWriter, code int, value interface{}) error {
 	return encoder.Encode(value)
 }
 
+// NewJSONResponse turns plain object into a byte array representing JSON value and wraps it in filter.Response
+func NewJSONResponse(code int, value interface{}) (*filter.Response, error) {
+	headers := http.Header{}
+	headers.Add("Content-Type", "application/json")
+	body, err := json.Marshal(value)
```
dpanayotov: we should panic all internal errors during the requests and have a recovery filter which converts a panic to a 500 response (improvement in another PR)

sm/sm.go
```diff
@@ -22,15 +22,25 @@ import (
...
 	"github.com/Peripli/service-manager/api"
+	"github.com/Peripli/service-manager/rest"
 	"github.com/Peripli/service-manager/server"
 	"github.com/Peripli/service-manager/storage"
 	"github.com/Peripli/service-manager/storage/postgres"
 	"github.com/sirupsen/logrus"
 )
 
+// Parameters contains context, environment for configuring a Service Manager server and optional extensions API
+type Parameters struct {
```
dpanayotov: Settings or Options

dpanayotov: Package name "app" would be better

rest/util.go
KirilKabakchiev: https://github.com/Peripli/service-manager/issues/67 - I've created an improvement for that sometime ago

api/platform/platform_controller.go
```diff
@@ -58,21 +54,21 @@ func checkPlatformMandatoryProperties(platform *types.Platform) error {
...
 
-	platform, errDecode := getPlatformFromRequest(req)
+	platform, errDecode := getPlatformFromRequest(request)
 	if errDecode != nil {
-		return errDecode
+		return nil, errDecode
 	}
 	if errMandatoryProperties := checkPlatformMandatoryProperties(platform); errMandatoryProperties != nil {
-		return types.NewErrorResponse(errMandatoryProperties, http.StatusBadRequest, "BadRequest")
+		return nil, filter.NewErrorResponse(errMandatoryProperties, http.StatusBadRequest, "BadRequest")
```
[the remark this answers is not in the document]
pankrator: When we return an error it will be converted to `Response` from the `HandleError` method in errors.go
I don't think we should construct a response object here.
Do you have a specific concern about this?

pkg/filter/filter.go
```diff
@@ -0,0 +1,67 @@
+package filter
```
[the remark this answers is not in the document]
pankrator: They are here, because they will be imported from other projects (ones that contain plugins) separately from the other symbols from `rest` package.

rest/filters.go
pankrator: Currently we don't have any naming on filters. Should we add names or identifiers?

sm/sm.go
pankrator: There are already too much Settings and Options. Parameters makes it clear that it contains optional function parameters.

rest/filters.go
pankrator: Added as an issue

pkg/filter/filter.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// Request contains the original http.Request, path parameters and the raw body
+// Request.Request.Body should not be used as it would be already processed by internal implementation
+type Request struct {
+	// Request is the original http.Request
+	*http.Request
+
+	// PathParams contains the URL path parameters
+	PathParams map[string]string
```
dpanayotov: This should probably be read-only. Maybe an issue can be opened for improvement.

api/broker/broker_controller.go
```diff
@@ -74,21 +73,22 @@ func validateBroker(broker *types.Broker) error {
 	return validateBrokerCredentials(broker.Credentials)
 }
 
-func (c *Controller) createBroker(response http.ResponseWriter, request *http.Request) error {
+func (c *Controller) createBroker(request *filter.Request) (*filter.Response, error) {
```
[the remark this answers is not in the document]
dotchev: yes, we should think of a better package name here, probably rest.*

pkg/filter/filter.go
dpanayotov: why would a path param be changed?

rest/filters.go
dotchev: ok, added filter name

rest/util.go
dotchev: The result is the same - status 500

PR #81, opened 2018-06-27
Respect PORT when running in CF

KirilKabakchiev:
Currently SM does not respect CF's PORT env variable and instead starts the process on a port of its own desire (defaults to 8080, picked up from application.yml). If CF picks a different port for the SM process, this leads to CF health check failure. 

This seems to have appeared after switching from running a docker container to using the go buildpack.

### DIFF
--- a/main.go
+++ b/main.go
@@ -26,7 +26,7 @@ func main() {
 	if err != nil {
 		panic(fmt.Sprintf("error loading environment: %s", err))
 	}
-	if err := cf.SetEnvValues(env); err != nil {
+	if err := cf.SetCFOverrides(env); err != nil {
 		panic(fmt.Sprintf("error setting CF environment values: %s", err))
 	}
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -26,7 +26,7 @@ func main() {
 	if err != nil {
 		panic(fmt.Sprintf("error loading environment: %s", err))
 	}
-	if err := cf.SetEnvValues(env); err != nil {
+	if err := cf.SetCFOverrides(env); err != nil {
```
pankrator: Why bind implementation to CF?

KirilKabakchiev: The "binding" is there from the time we removed the two main.go-s ( it's not a exactly a binding, but we override some properties if VCAP_APPLICATION is found)

Option two would be to have two main methods or additional flag to specify mode in which the service works

PR #84, opened 2018-07-02
Pkg libraries to be reused in proxies

KirilKabakchiev:
- [x] Initial implementation
- [ ] Tests


config/config.go
```diff
@@ -44,20 +36,19 @@ type Settings struct {
...
 
-// File describes the name, path and the format of the file to be used to load the configuration in the env
-type File struct {
-	Name     string
-	Location string
-	Format   string
+// AddPFlags adds the SM config flags to the provided flag set
+func AddPFlags(set *pflag.FlagSet) {
+	env.CreatePFlags(set, DefaultSettings())
+	env.CreatePFlags(set, struct{ File env.File }{File: env.DefaultConfigFile()})
```
pankrator: Why not use `CreatePFlagsForConfigFile` from env package?

pkg/log/log.go
```diff
@@ -29,12 +33,15 @@ type Settings struct {
 func SetupLogging(settings Settings) {
 	level, err := logrus.ParseLevel(settings.Level)
 	if err != nil {
-		panic("Could not parse log level configuration")
+		panic(fmt.Sprintf("Could not parse log level configuration: %s", err.Error()))
 	}
 	logrus.SetLevel(level)
-	if settings.Format == "json" {
+	switch settings.Format {
```
pankrator: may be create a map with key - the type of format, and value - logrus formatter

PR #85, opened 2018-07-03
Add recovery handler

georgifarashev:
# Pull Request Template

## Motivation
https://github.com/Peripli/service-manager/issues/67

## Approach
[Gorilla RecoveryHandler](https://godoc.org/github.com/gorilla/handlers#RecoveryHandler) is used

#### Pull Request status

For example:

- [x] Initial implementation
- [ ] Tests


pankrator:
One more thing.
Can you add test with the same controller, but attach also a filter that panics. Not sure if perfectly valid, but I think it would be nice to have such a test.

PR #91, opened 2018-07-11
Add healthcheck API

georgifarashev:
# Add healthcheck API

## Motivation
Provide a monitoring endpoint that can be checked regularly if the SM is up and running.

## Approach
Provide endpoint at `/monitor/health`

#### Pull Request status

For example:

- [x] Initial implementation
- [x] Refactoring
- [x] Unit tests
- [x] Integration tests


api/healthcheck/healthcheck_controller.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/Peripli/service-manager/rest"
+	"github.com/Peripli/service-manager/storage"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	apiVersion = "v1"
+	root       = "monitor/health"
+	// URL is the path of the healthcheck endpoint
+	URL = "/" + apiVersion + "/" + root
```
pankrator: Can we make URL private?

api/healthcheck/healthcheck_controller.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// Controller platform controller
+type Controller struct {
+	Storage storage.Storage
+}
+
+// Routes returns slice of routes which handle healthcheck operation
+func (c *Controller) Routes() []rest.Route {
```
NickyMateev: I think it might be better for consistency's sake to have this extracted in a `healthcheck.go` file, because every other API we've defined so far is split like so - `someAPI.go` and `someAPI_controllers.go`.

api/healthcheck/healthcheck_controller.go
georgifarashev: It is used in the `healthcheck_test.go:54`

PR #94, opened 2018-07-24
Remove credentials table

georgifarashev:
# Remove credentials table

- [x] Initial implementation
- [x] Refactoring
- [n/a] Unit tests
- [n/a] Integration tests

### DIFF
--- a/api/platform/platform_controller.go
+++ b/api/platform/platform_controller.go
@@ -99,6 +99,7 @@ func (c *Controller) getPlatform(request *web.Request) (*web.Response, error) {
 	if err = common.HandleNotFoundError(err, "platform", platformID); err != nil {
 		return nil, err
 	}
+	platform.Credentials = nil
 	return rest.NewJSONResponse(http.StatusOK, platform)
 }
 
@@ -109,9 +110,15 @@ func (c *Controller) getAllPlatforms(request *web.Request) (*web.Response, error
 	if err != nil {
 		return nil, err
 	}
-	platformsResponse := map[string][]types.Platform{"platforms": platforms}
+	for _, platform := range platforms {
+		platform.Credentials = nil
+	}
 
-	return rest.NewJSONResponse(http.StatusOK, &platformsResponse)
+	return rest.NewJSONResponse(http.StatusOK, struct {
+		Platforms []*types.Platform `json:"platforms"`
+	}{
+		Platforms: platforms,
+	})
 }
 
 // deletePlatform handler for DELETE /v1/platforms/:platform_id
@@ -136,6 +143,7 @@ func (c *Controller) patchPlatform(request *web.Request) (*web.Response, error) 
 		return nil, errDecode
 	}
 	newPlatform.ID = platformID
+	newPlatform.CreatedAt = time.Time{}
 	newPlatform.UpdatedAt = time.Now().UTC()
 	platformStorage := c.PlatformStorage
 	err := platformStorage.Update(newPlatform)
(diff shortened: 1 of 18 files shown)


georgifarashev:
This change will brake existing deployments as DB migrations are not backwards compatible.

storage/postgres/migrations/001_initial_schema.up.sql
```diff
@@ -1,40 +1,26 @@
...
     broker_url text NOT NULL,
-    created_at timestamp DEFAULT current_timestamp,
-    updated_at timestamp DEFAULT current_timestamp,
-    credentials_id integer NOT NULL,
-    FOREIGN KEY (credentials_id) REFERENCES credentials
+    created_at timestamp DEFAULT current_timestamp NOT NULL,
+    updated_at timestamp DEFAULT current_timestamp NOT NULL,
+    username varchar(255) NOT NULL,
+    password varchar(500) NOT NULL,
+    catalog json
```
georgifarashev: maybe we should add not null here too

storage/postgres/migrations/001_initial_schema.up.sql
```diff
@@ -1,40 +1,26 @@
...
     name varchar(255) NOT NULL UNIQUE,
     description text,
     broker_url text NOT NULL,
-    created_at timestamp DEFAULT current_timestamp,
-    updated_at timestamp DEFAULT current_timestamp,
-    credentials_id integer NOT NULL,
-    FOREIGN KEY (credentials_id) REFERENCES credentials
+    created_at timestamp DEFAULT current_timestamp NOT NULL,
+    updated_at timestamp DEFAULT current_timestamp NOT NULL,
+    username varchar(255) NOT NULL,
```
NickyMateev: Username should be UNIQUE just like with the platforms table.

api/platform/platform_controller.go
```diff
@@ -136,6 +143,7 @@ func (c *Controller) patchPlatform(request *web.Request) (*web.Response, error)
 		return nil, errDecode
 	}
 	newPlatform.ID = platformID
+	newPlatform.CreatedAt = time.Time{}
 	newPlatform.UpdatedAt = time.Now().UTC()
 	platformStorage := c.PlatformStorage
```
NickyMateev: Why extract the storage to a variable? I'd rather use ```c.PlatformStorage``` directly on the next line for the sake of consistency throughout the file.

api/platform/platform_controller.go
```diff
@@ -136,6 +143,7 @@ func (c *Controller) patchPlatform(request *web.Request) (*web.Response, error)
 		return nil, errDecode
 	}
 	newPlatform.ID = platformID
+	newPlatform.CreatedAt = time.Time{}
```
NickyMateev: Won't this override the original creation time?

georgifarashev: No. This will set `CreatedAt` to the zero value of `time.Time` which means it will be skipped in the update query. The idea here is to prevent users from modifying the `CreatedAt` field.

api/platform/platform_controller.go
georgifarashev: It is used in line 157 as well

api/platform/platform_controller.go
NickyMateev: Got it 👍

storage/postgres/migrations/001_initial_schema.up.sql
georgifarashev: It should be possible to register two brokers with the same username.

api/platform/platform_controller.go
KirilKabakchiev: this is not very reliable though - at some point we would want to be able to update certain fields to their default values (e.x. clean up description)

PR #100, opened 2018-07-30
Fix k8s deployment

georgifarashev:
# Fix k8s deployment

- [x] Initial implementation
- [n/a] Refactoring
- [n/a] Unit tests
- [n/a] Integration tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,10 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -o /mai
 ########################################################
 # Build the runtime container
 ########################################################
-FROM scratch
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -23,7 +23,10 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -o /mai
 ########################################################
 # Build the runtime container
 ########################################################
-FROM scratch
+FROM alpine:3.7 AS package_step
+
+RUN apk update && apk add ca-certificates
```
KirilKabakchiev: Is this currently needed?

georgifarashev: Currently not but it might be useful in the future.

PR #102, opened 2018-07-31
Proxy last operation call

dpanayotov:
Motivation:
Async service instance creation from platforms calls last_operation endpoint of the broker

Steps:
- [x] Implementation
- [x] Tests

### DIFF
--- a/.gometalinter.json
+++ b/.gometalinter.json
@@ -5,5 +5,6 @@
     "Deadline": "120s",
     "Aggregate": true,
     "Sort": ["severity", "path", "line", "message"],
-    "Checkstyle": false
+    "Checkstyle": false,
+    "Cyclo": 12
 }
--- a/test/common/test_context.go
+++ b/test/common/test_context.go
@@ -94,7 +94,6 @@ type TestContext struct {
 	SM          *httpexpect.Expect
 	SMWithOAuth *httpexpect.Expect
 	SMWithBasic *httpexpect.Expect
-	SMServer    *httptest.Server
 
 	brokers map[string]*Broker
 }
@@ -123,10 +122,9 @@ func (ctx *TestContext) Cleanup() {
 		return
 	}
 
-	if ctx.SMServer != nil {
+	if ctx.SMWithOAuth != nil {
 		RemoveAllBrokers(ctx.SMWithOAuth)
 		RemoveAllPlatforms(ctx.SMWithOAuth)
-		ctx.SMServer.Close()
 	}
 
 	for _, broker := range ctx.brokers {
--- a/test/osb_test/osb_test.go
+++ b/test/osb_test/osb_test.go
@@ -211,18 +211,24 @@ var _ = Describe("Service Manager OSB API", func() {
 	})
 
 	Describe("Get Last Operation", func() {
-		Context("For service instance", func() {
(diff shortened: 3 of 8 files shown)


test/common/test_context.go
```diff
@@ -94,7 +94,6 @@ type TestContext struct {
 	SM          *httpexpect.Expect
 	SMWithOAuth *httpexpect.Expect
 	SMWithBasic *httpexpect.Expect
-	SMServer    *httptest.Server
```
pankrator: Why remove SMServer?

.gometalinter.json
```diff
@@ -5,5 +5,6 @@
     "Deadline": "120s",
     "Aggregate": true,
     "Sort": ["severity", "path", "line", "message"],
-    "Checkstyle": false
+    "Checkstyle": false,
+    "Cyclo": 12
```
pankrator: Better skip function cyclo check. I will try to find the comment that needs to be added to disable the check

test/common/test_context.go
pankrator: The reference to the httptest.Server is lost during the construction of testContext, than no one is responsible to close it.

test/osb_test/osb_test.go
```diff
@@ -209,4 +209,26 @@ var _ = Describe("Service Manager OSB API", func() {
...
+		assertLastOperationHasState := func(url string) {
+			resp := ctx.SMWithBasic.GET(url).WithHeader("X-Broker-API-Version", "oidc_authn.13").
+				Expect().Status(http.StatusOK).JSON().Object()
+
+			resp.ContainsKey("state")
+		}
+
+		Context("For service instance", func() {
+			It("should return state", func(){
+				assertLastOperationHasState(validBroker.OSBURL+ "/v2/service_instances/iid/last_operation")
```
pankrator: Can you also add test requests with appropriate query parameters like described here: https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#parameters

test/plugin_test/plugin_test.go
```diff
@@ -172,6 +172,8 @@ var _ = Describe("Service Manager Plugins", func() {
 			{"bind", "PUT", "/v2/service_instances/1234/service_bindings/111"},
 			{"unbind", "DELETE", "/v2/service_instances/1234/service_bindings/111"},
 			{"fetchBinding", "GET", "/v2/service_instances/1234/service_bindings/111"},
+			{"pollInstance", "GET","/v2/service_instances/1234/last_operation"},
```
pankrator: should we verify the plugin execution is triggered when query parameters are used

.gometalinter.json
pankrator: https://github.com/alecthomas/gometalinter#comment-directives

test/plugin_test/plugin_test.go
dpanayotov: Good point... This is an issue for all endpoints currently. Not sure how to handle that. It would be easiest to forward all calls `/v1/osb/{brokerId}/**` regardless of whether they're OSB. After all, a broker might implement its own API that's not OSB but still can be called by someone

test/common/test_context.go
pankrator: It is created on line 88 (https://github.com/Peripli/service-manager/blob/e4d08df3954ad01b38f837fa7854ec0f61c1e4d4/test/common/test_context.go#L88) and never closed.

PR #104, opened 2018-08-03
Configure server host

dotchev:
On Mac you get a nasty popup each time
you start it unless it listens on localhost


dpanayotov:
I don't remember if it was with docker on cloud foundry or plain app on cloud foundry but it doesn't work there

dotchev:
By default nothing changes. This takes effect only if you set the server host in the config. Otherwise it is empty, so SM will listen on all interfaces.

PR #107, opened 2018-08-07
Lock when generating encryption key

### LINKED ISSUE #106: Failure to start when generating encryption key
When service-manager runs on more than 1 node, generating encryption key for the first time might fail. This happens because all nodes get that there is no encryption key and all try to generate one. However, if one generates before the others, others will fail because only 1 encryption key can be set.

dpanayotov:
Fixes #106 

Opened #116 for test


pkg/sm/encryption.go
```diff
@@ -0,0 +1,75 @@
...
+func waitForLeader(keyFetcher security.KeyFetcher) error {
+	var encryptionKey []byte
+	for i := 0; i < 5; i++ {
+		logrus.Debugf("Waiting for leader to generate encryption key...")
+		encryptionKey, err := keyFetcher.GetEncryptionKey()
+		if err != nil {
+			return err
+		}
+		if len(encryptionKey) != 0 {
+			return nil
```
dzahariev: I think here you should return the key instead null

pkg/sm/sm.go
```diff
@@ -155,6 +155,9 @@ func (smb *ServiceManagerBuilder) Build() *ServiceManager {
 }
 
 func initializeSecureStorage(secureStorage storage.Security) error {
+	if err := secureStorage.Lock(); err != nil {
+		return err
```
pankrator: We don't want to return an error if database is locked. I think it should be:
1) Try to lock database.
2) If it is already locked, then wait some time (2 seconds for example)
3) Try to lock again
4) Repeat several times(5 for example)
5) If it couldn't acquire the lock, then panic
6) If lock is acquired then check for the encryption key. If it is already there then do nothing, otherwise generate one

pankrator: It appears that advisory locks block until a lock can be acquired: 
https://medium.com/@codeflows/application-level-locking-with-postgres-advisory-locks-c29759eeb7a7

test/main_test.go
```diff
@@ -0,0 +1,85 @@
...
+		session.Kill()
+		session2.Kill()
+		os.Unsetenv("API_TOKEN_ISSUER_URL")
+	})
+	Context("when more than one instance is started", func() {
+		It("should wait for lock", func() {
+			if os.Getenv("CLEAN_ENV") == "" {
+				Skip("Environment is not clean to validate behavior of parallel encryption key generation")
+			} else {
+				command := exec.Command(binaryPath)
```
KirilKabakchiev: pass port as flag here so both can use different ports

storage/postgres/security.go
```diff
@@ -25,9 +25,40 @@ import (
...
 type securityStorage struct {
 	db            *sqlx.DB
 	encryptionKey []byte
+	isLocked      bool
+}
+
+// Lock acquires a database lock so that only one process can manipulate the encryption key.
+// Returns an error if the process has already acquired the lock
+func (s *securityStorage) Lock() error {
+	if s.isLocked {
```
KirilKabakchiev: add a mutex here to safeguard from future possible raceconditions

storage/postgres/security.go
```diff
@@ -25,9 +25,40 @@ import (
...
+	if err != nil {
+		return err
+	}
+	s.isLocked = true
+	return nil
+}
+
+// Unlock releases the database lock.
+func (s *securityStorage) Unlock() error {
+	if !s.isLocked {
```
KirilKabakchiev: add a mutex here to safeguard from future possible raceconditions

storage/postgres/security.go
```diff
@@ -17,17 +17,53 @@
...
 	db            *sqlx.DB
 	encryptionKey []byte
+	isLocked      bool
+	mutex         *sync.Mutex
+}
+
+// Lock acquires a database lock so that only one process can manipulate the encryption key.
+// Returns an error if the process has already acquired the lock
+func (s *securityStorage) Lock(ctx context.Context) error {
+	s.mutex.Lock()
```
KirilKabakchiev: is this the correct lock ( e.x. this or RWLock )

dpanayotov: RWMutex provides only R(W)Lock and R(W)Unlock for these explicit situations. RWMutext.Lock/Unlock is the same as Mutex.Lock/Unlock
Since this is both reading and writing the value, there's no difference between RWMutex.Lock and Mutex.Lock

PR #108, opened 2018-08-07
Filter ordering

dpanayotov:
Added functionality:

* `RegisterFilterBefore(name, filter)` - Registers the specified filter before the one with that name. If there is no filter with such name panics. Option is to add `RegisterFilterBefore` for methods that return an error and `RegisterFilterBeforeOrDie` that panic on error as the standard library takes such approach.
* `RegisterFilterAfter(name, filter)` - Registers the specified filter after the one with that name. If there is no filter with such name panics.
* `ReplaceFilter(name, filter)` - Registers the specified filter in the places of the one with that name preserving the order. If there is no filter with such name panics.
* `RemoveFilter(name)` - Deletes the filter from the chain, preserving the order. If there is no filter with such name panics.
* `RegisterFilters(filters ...Filter)` now panics if a filter with the same name has already been registered. In addition, filters containing `:` in their names are forbidden as this is a name separator for the plugins.
* `RegisterPlugins(plugins ...Plugin)` now panics if a plugin with the same name has already been registered

# Note 
If a filter is registered either before, after or in the place of another this is happening in a global filter chain, regardless of the routes matched by the different filters. This gives the following side effect:
Let's have the following filters with matchers:
```
FilterA matches /v1 and /v2
FilterB matches /v2 and /v3
FilterC matches /v1 and /v3
FilterD matches /**
```
And let the filter chain be `[FilterA, FilterB, FilterC]`
`api.RegisterFilterAfter("FilterA", FilterD)` and making calls to `/v1`, `/v2` and `/v3` will yield the following chains:
```
/v1: FilterA, FilterD, FilterC
/v2: FilterA, FilterD, FilterB
/v3: FilterD, FilterC
```
Even though `FilterA` does not match `/v3`, `FilterD` does and is registered at the position at which `FilterA` would have been had it matched `/v3`

### DIFF
--- a/pkg/sm/sm.go
+++ b/pkg/sm/sm.go
@@ -117,24 +117,9 @@ func New(ctx context.Context, env env.Environment) *ServiceManagerBuilder {
 	}
 }
 
-// ServiceManager  struct
-type ServiceManager struct {
-	ctx    context.Context
-	Server *server.Server
-}
-
 // Run starts the Service Manager
 func (sm *ServiceManager) Run() {
 	sm.Server.Run(sm.ctx)
-}
-
-// ServiceManagerBuilder type is an extension point that allows adding additional filters, plugins and
-// controllers before running ServiceManager.
-type ServiceManagerBuilder struct {
-	*web.API
-	ctx    context.Context
-	cancel context.CancelFunc
-	cfg    server.Settings
 }
 
 // Build builds the Service Manager
@@ -147,11 +132,6 @@ func (smb *ServiceManagerBuilder) Build() *ServiceManager {
 		ctx:    smb.ctx,
 		Server: srv,
 	}
-}
-
-// Run starts the Service Manager
-func (sm *ServiceManager) Run() {
-	sm.Server.Run(sm.ctx)
 }
 
 func initializeSecureStorage(ctx context.Context, secureStorage storage.Security) error {
--- a/pkg/sm/sm.go
+++ b/pkg/sm/sm.go
@@ -58,6 +58,7 @@ type ServiceManager struct {
 // DefaultEnv creates a default environment that can be used to boot up a Service Manager
 func DefaultEnv(additionalPFlags ...func(set *pflag.FlagSet)) env.Environment {
 	set := env.EmptyFlagSet()
+
 	config.AddPFlags(set)
 	for _, addFlags := range additionalPFlags {
 		addFlags(set)
@@ -117,11 +118,6 @@ func New(ctx context.Context, env env.Environment) *ServiceManagerBuilder {
 	}
 }
 
-// Run starts the Service Manager
-func (sm *ServiceManager) Run() {
-	sm.Server.Run(sm.ctx)
-}
-
 // Build builds the Service Manager
 func (smb *ServiceManagerBuilder) Build() *ServiceManager {
 	// setup server and add relevant global middleware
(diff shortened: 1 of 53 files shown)


pkg/sm/sm.go
```diff
@@ -121,25 +121,10 @@ func (sm *ServiceManager) Run() {
 // ServiceManagerBuilder type is an extension point that allows adding additional filters, plugins and
 // controllers before running ServiceManager.
 type ServiceManagerBuilder struct {
+	*web.API
```
KirilKabakchiev: maybe embed without the star

pkg/web/api.go
```diff
@@ -82,15 +84,158 @@ func (api *API) RegisterFilters(filters ...Filter) {
...
+	for i := range api.Filters {
+		registeredFilter := api.Filters[i]
+		if registeredFilter.Name() == filterName {
+			return i
+		}
+	}
+	return -1
+}
+
+func (api *API) findFilterPositionOrDie(filterName string) int {
```
KirilKabakchiev: im not sure if panicing or returning error is better here. maybe both?

pkg/web/api.go
```diff
@@ -82,15 +84,158 @@ func (api *API) RegisterFilters(filters ...Filter) {
...
+
+func (api *API) decomposePluginOrDie(plugin Plugin) []Filter {
+	pluginSegments := api.decomposePlugin(plugin)
+	if len(pluginSegments) == 0 {
+		logrus.Panicf("%T does not implement any plugin operation", plugin)
+	}
+	return pluginSegments
+}
+
+func (api *API) ReplacePlugin(name string, plugin Plugin) {
```
KirilKabakchiev: Replace only if after decomposing the plugin you get segments that are the same (same count and same type and same order) as the segments of the plugin that is being replaced

pkg/web/api.go
```diff
@@ -82,15 +84,158 @@ func (api *API) RegisterFilters(filters ...Filter) {
 	api.Filters = append(api.Filters, filters...)
```
KirilKabakchiev: should check that name is valid:
1. not empty
2. does not already exist
3. does not contain : delimiter because we will use it as a special symbol

pkg/web/api.go
```diff
@@ -82,15 +84,158 @@ func (api *API) RegisterFilters(filters ...Filter) {
...
+	logrus.Debugf("Registering filter %s before %s", filter.Name(), beforeFilterName)
+	api.registerFilterRelatively(beforeFilterName, filter, func(beforeFilterPosition int) int {
+		if beforeFilterPosition == 0 {
+			return 0
+		}
+		return beforeFilterPosition - 1
+	})
+}
+
+// RegisterFilterAfter adds the given filter after the filter with the specified name
```
KirilKabakchiev: RouteMatchersValue []FilterMatcher of pluginsegment - make it single filter matcher

pkg/sm/sm.go
dpanayotov: it would require modification down a few levels. Not sure if that's ok. Pointer would be ok I think

pkg/web/api.go
```diff
@@ -79,19 +80,115 @@ func (api *API) RegisterControllers(controllers ...Controller) {
...
+	if slice.StringsAnyEquals(newFilterNames, "") {
+		logrus.Panicf("Filters cannot have empty names")
+	}
+	registeredFilterNames := api.filterNames(api.Filters)
+	commonFilterNames := slice.StringsIntersection(registeredFilterNames, newFilterNames)
+	if len(commonFilterNames) > 0 {
+		logrus.Panicf("Filters %s are already registered", commonFilterNames)
+	}
+	if slice.StringsAnySubstring(newFilterNames, ":") {
+		logrus.Panic("Cannot register filters with : in their names")
```
KirilKabakchiev: would it make sense to do the validation in way so that we may include the invalid name in the panic?

PR #113, opened 2018-08-10
Reverse proxy

pankrator:
# Pull Request Template

## Prerequisites
* Reverse proxy needs to be used in other Service Manager components like the SB proxy

## Approach
Extract reverse proxy in separate package

- [x] Initial implementation
- [x] Refactoring
- [] Unit tests

### DIFF
--- a/api/osb/osb_controller.go
+++ b/api/osb/osb_controller.go
@@ -17,15 +17,12 @@
 package osb
 
 import (
-	"bytes"
 	"fmt"
-	"io/ioutil"
 	"net/http"
-	"net/http/httptest"
-	"net/http/httputil"
 	"net/url"
 	"regexp"
 
+	"github.com/Peripli/service-manager/pkg/proxy"
 	"github.com/Peripli/service-manager/pkg/types"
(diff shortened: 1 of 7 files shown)


api/osb/osb_controller.go
```diff
@@ -53,48 +52,41 @@ func (c *Controller) handler(request *web.Request) (*web.Response, error) {
...
 		return nil, err
 	}
-	modifiedRequest.SetBasicAuth(username, string(plaintextPassword))
-	modifiedRequest.URL.Scheme = target.Scheme
-	modifiedRequest.URL.Host = target.Host
-	modifiedRequest.Body = ioutil.NopCloser(bytes.NewReader(request.Body))
-	modifiedRequest.ContentLength = int64(len(request.Body))
+
+	proxier := proxy.NewReverseProxy(proxy.Options{
+		Transport: http.DefaultTransport,
```
pankrator: Should not use default Transport here. Can't think of better way at the moment. Perhaps configure it outside of the controller and pass it to the controller?

PR #119, opened 2018-08-18
Trim broker URL before persisting

NickyMateev:
# Trim broker URL before persisting

This pull request suggests trimming broker URLs and thus removing any trailing forward slashes before persisting any broker.

The need for this PR arises from the fact that in a Kubernetes environment (possibly in other platforms in the future as well) the Service Catalog fails to retrieve broker catalogs for brokers that have a URL with trailing forward slashes("/").


NickyMateev:
Closed due to the fact that #120 fixes the real issue.

PR #123, opened 2018-08-22
Filter blacklisting

NickyMateev:
# Filter blacklisting

## Motivation
This PR proposes adding a filter blacklist which is used to exclude filters by name when the Service Manager is being setup.

This would ease local development and allow for more flexibility.

#### Pull Request status

- [x] Initial implementation
- [x] Fix old tests
- [x] Unit tests

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -40,6 +40,12 @@ type Security struct {
 	EncryptionKey string `mapstructure:"encryption_key"`
 }
 
+// Filters is the configuration used for filter settings
+type Filters struct {
+	// Blacklist holds the names of filters to be excluded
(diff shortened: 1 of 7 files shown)


NickyMateev:
One thing that bothers me in this PR is that the **web.MatchInArray(...)** function would be better suited to be in the _pkg/util_ package, but then a circular dependency happens.

Suggestions about mitigating this would be appreciated!

NickyMateev:
Another thing to note is that the authentication filters still get created even if they are blacklisted - they are just not attached afterwards.
This could be annoying, because a UAA will need to be present to create the Bearer filter, despite the fact the filter itself won't be attached afterwards..

PR #128, opened 2018-08-28
Refactor component tests

dotchev:
# Refactor component tests

## Motivation
Make test context reusable so it can be used by custom service manager wrappers.


dotchev:
This is true for the whole `common` package. Probably we should rename it.

PR #135, opened 2018-09-14
Reusable log interceptor

dotchev:


test/testutil/log_interceptor.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// Levels defines which log levels to capture
+func (*LogInterceptor) Levels() []logrus.Level {
+	return logrus.AllLevels
+}
+
+// Fire is called for each log entry
+func (hook *LogInterceptor) Fire(entry *logrus.Entry) error {
+	str, _ := entry.String()
```
KirilKabakchiev: prolly needs synchrozation

dotchev: logrus fires hooks synchronously, see https://github.com/sirupsen/logrus/blob/master/entry.go#L143

PR #138, opened 2018-09-20
Remove redundant Server validation

NickyMateev:
Validation is redundant. The check has already been made here: [pkg/sm/sm.go][1]

[1]: https://github.com/Peripli/service-manager/blob/master/pkg/sm/sm.go#L83


KirilKabakchiev:
discussed - server is in pkg and if used standalone, validating config before running makes sure everything needed is there before actually running - closing for now

PR #141, opened 2018-09-27
Circuit Breaker and Retry

dpanayotov:
does it make sense to set a circuit breaker as a filter and cut all requests to an endpoint?


dpanayotov:
closing due to inactivity

PR #142, opened 2018-09-28
Aggregated health checks

dpanayotov:
Allows defining multiple health check indicators and aggregates all registered in one health status. If one of the indicators is DOWN, the whole status is DOWN. All indicators must be UP in order for the aggregated health status to be UP
Example of unavailable database:
```json
{  
   "status":"DOWN",
   "details":{  
      "storage":{  
         "status":"DOWN",
         "details":{  
            "error":{  
               "Op":"dial",
               "Net":"tcp",
               "Source":null,
               "Addr":{  
                  "IP":"192.168.99.100",
                  "Port":5555,
                  "Zone":""
               },
               "Err":{  
                  "Syscall":"connectex",
                  "Err":10061
               }
            }
         }
      }
   }
}
```

Example of up:
```json
{  
   "status":"UP",
   "details":{  
      "storage":{  
         "status":"UP"
     }
   }
}
```


pkg/health/composite_indicator.go
```diff
@@ -0,0 +1,45 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package health
+
+// CompositeIndicator aggregates multiple health indicators and provides one detailed health
+type CompositeIndicator struct {
```
KirilKabakchiev: I am not sure about the composite indicator - if one registers two composite indicators that each has a different aggregator, he should be able to register a top level agregator too, which aggregates the result of the most outer idicators. And this might result in unnecessay complexity

How about we have in the registry just a slice of indicators and one agregation policy? this way the consumer can define everything and it will be simpler

api/healthcheck/healthcheck_controller.go
```diff
@@ -19,45 +19,30 @@ package healthcheck
...
 // healthCheck handler for GET /v1/monitor/health
 func (c *Controller) healthCheck(r *web.Request) (*web.Response, error) {
 	ctx := r.Context()
 	logger := log.C(ctx)
-	logger.Debug("Performing health check...")
-
-	if err := c.Storage.Ping(); err != nil {
-		logger.Debugf("storage.Ping failed: %s", err)
-		return util.NewJSONResponse(http.StatusServiceUnavailable, statusStorageFailureResponse)
+	logger.Debug("Performing healthResult check...")
```
KirilKabakchiev: the indicators are named - should we log the names of the indicators used in the health check?

pkg/health/default_aggregator.go
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package health
+
+// DefaultAggregator aggregates the healths by constructing a new Health based on the given
+// where the overall health status is negative if one of the healths is negative and positive if all are positive
+type DefaultAggregator struct {
```
KirilKabakchiev: just naming but - how about aggregation policy instead of aggregator?

storage/healthcheck.go
```diff
@@ -0,0 +1,39 @@
...
+func (i *HealthIndicator) Name() string {
+	return "storage"
+}
+
+// Health returns the health of the storage component
+func (i *HealthIndicator) Health() *health.Health {
+	err := i.Storage.Ping()
+	health := health.New()
+	if err != nil {
+		return health.WithError(err)
```
KirilKabakchiev: can we add some human readable info here, because the error returned is not very understandable

pkg/health/types.go
```diff
@@ -0,0 +1,110 @@
...
+// WithDetails adds the given details to the health
+func (h *Health) WithDetails(details map[string]interface{}) *Health {
+	for k, v := range details {
+		h.Details[k] = v
+	}
+	return h
+}
+
+// Indicator is an interface to provide the health of a component
+type Indicator interface {
```
KirilKabakchiev: should we define a ping indicator that just returns 200 ok and install it by default? something like https://sourcegraph.com/github.com/kubernetes/apiserver@52b58255fbaacc9fbeeefeca6543121f8508f22e/-/blob/pkg/server/config.go#L258

api/healthcheck/composite_indicator.go
```diff
@@ -0,0 +1,47 @@
...
+func newCompositeIndicator(indicators []health.Indicator, aggregator health.AggregationPolicy) health.Indicator {
+	return &compositeIndicator{
+		aggregationPolicy: aggregator,
+		indicators:        indicators,
+	}
+}
+
+// Name returns the name of the compositeIndicator
+func (i *compositeIndicator) Name() string {
+	return "CompositeHealthIndicator"
```
KirilKabakchiev: I guess it would be more useful if the name of the internal indictator is actually composed on the names of the indicators slice. (especially since we now log it in the Controller)

pkg/health/registry.go
```diff
@@ -0,0 +1,56 @@
...
+
+type defaultRegistry struct {
+	indicators        []Indicator
+	aggregationPolicy AggregationPolicy
+	indicatorsMutex   sync.Mutex
+	aggregatorMutex   sync.Mutex
+}
+
+func (p *defaultRegistry) RegisterHealthAggregationPolicy(aggregator AggregationPolicy) {
+	p.aggregatorMutex.Lock()
```
KirilKabakchiev: do we actually need this locking? we don't have post-start hooks so

pkg/health/registry_test.go
```diff
@@ -0,0 +1,84 @@
...
+			postAddIndicators := registry.HealthIndicators()
+			for _, indicator := range preAddIndicators {
+				Expect(postAddIndicators).To(ContainElement(indicator))
+			}
+			Expect(postAddIndicators).To(ContainElement(newIndicator))
+		})
+	})
+})
+
+type testAggregationPolicy struct {
```
KirilKabakchiev: you could reuse the generated fake

pkg/web/api.go
```diff
@@ -32,6 +33,9 @@ type API struct {
 
 	// Filters contains the registered filters
 	Filters []Filter
+
+	// Registry is the health indicators registry for this API
+	health.Registry
```
KirilKabakchiev: would it make sense to add a little segement to the documentation (extensions) to specify the possiblity to add health indicators?

storage/healthcheck.go
```diff
@@ -0,0 +1,39 @@
...
+import "github.com/Peripli/service-manager/pkg/health"
+
+// HealthIndicator returns a new indicator for the storage
+type HealthIndicator struct {
+	Storage Storage
+}
+
+// Name returns the name of the storage component
+func (i *HealthIndicator) Name() string {
+	return "storage"
```
KirilKabakchiev: StorageHealthIndicator (if these will be part of logs, more complete names will be more useful)

pkg/health/ping.go
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package health
+
+// pingIndicator is a default indicator that always returns up
+type pingIndicator struct {
+}
+
+func (*pingIndicator) Name() string {
+	return "ping"
```
KirilKabakchiev: PingHealthIndicator (if these will be part of logs, more complete names will be more useful)

storage/healthcheck.go
dpanayotov: this will be in the /v1/monitor/health so the json then would have 
```json
{  
   "status":"UP",
   "details":{  
      "StorageHealthIndicator":{  
         "status":"UP
     }
  }
}
```

pkg/health/registry_test.go
dpanayotov: cyclic dependency. healthfakes imports health :(

pkg/health/registry.go
dpanayotov: I thought a goroutine might register a health indicator, but then it won't be taken by the controller, so these are unnecessary

pkg/sm/sm.go
```diff
@@ -120,6 +122,8 @@ func New(ctx context.Context, env env.Environment) *ServiceManagerBuilder {
 // Build builds the Service Manager
 func (smb *ServiceManagerBuilder) Build() *ServiceManager {
 	// setup server and add relevant global middleware
+	smb.installHealth()
```
dpanayotov: this is a slight issue, because the sbproxy doesn't use ServiceManagerBuilder, but SBProxyBuilder, so it can't get the default behaviour. Basically this same code should be copied in SBProxyBuilder...

KirilKabakchiev: Thats correct - we discussed it also in terms of the race conditions fix - its not ideal - probably in a future pull request it wouls make sense to extract a generic application that can be built and contains all the duplicated stuff

PR #143, opened 2018-09-28
Release connection on ping

dpanayotov:


storage/postgres/storage_state.go
```diff
@@ -60,10 +61,14 @@ func (state *storageState) checkDB() error {
 	if state.cachedStateIsValid() {
 		return state.storageError
 	}
-	if _, err := state.db.Query("SELECT 1"); err != nil {
+	rows, err := state.db.Query("SELECT 1")
+	if err != nil {
 		state.storageError = err
 	} else {
 		state.storageError = nil
```
KirilKabakchiev: do we need this line?

georgifarashev: If the previous state was invalid and an error had occurred `state.stogageError` will not be `nil`. If the state is ok again this error should be cleared.

PR #146, opened 2018-10-05
Add makefile

dpanayotov:


Makefile
```diff
@@ -0,0 +1,64 @@
...
+ARCH     ?= amd64
+
+ifeq ($(PLATFORM),windows)
+FILE_EXT=.exe
+else
+FILE_EXT=
+endif
+
+GO_BUILD = env CGO_ENABLED=0 GOOS=$(PLATFORM) GOARCH=$(ARCH) \
+           go build
```
KirilKabakchiev: might be useful to add the posibility to include custom flags here when running `make build`

Makefile
```diff
@@ -0,0 +1,76 @@
...
+
+build: .init dep service-manager
+
+dep:
```
KirilKabakchiev: just an idea (prolly not for this PR) - but for windows users it would prolly make sense to include dep-docker or something - idea is to build in container and mount the directory.

Makefile
```diff
@@ -0,0 +1,76 @@
...
+ifeq ($(PLATFORM),windows)
+FILE_EXT=.exe
+else
+FILE_EXT=
+endif
+
+GO_BUILD = env CGO_ENABLED=0 GOOS=$(PLATFORM) GOARCH=$(ARCH) \
+           go build
+
+build: .init dep service-manager
```
KirilKabakchiev: build: .init dep-vendor service-manager

Makefile
```diff
@@ -0,0 +1,76 @@
+#    Copyright 2018 The Service Manager Authors
```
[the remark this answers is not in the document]
dpanayotov: I think it's best to have separate makefiles for every project and maybe reference them from here. `make everything` would build sm, sbproxy-cf, sbproxy-k8s.

Makefile
dpanayotov: It'd be best to act depending on the current state:
if there's no vendor - dep-vendor
If Gopkg.lock is empty - dep

Makefile
```diff
@@ -0,0 +1,76 @@
...
+BUILD_LDFLAGS =
+
+#  GOFLAGS -  extra "go build" flags to use - e.g. -v   (for verbose)
+GO_BUILD = env CGO_ENABLED=0 GOOS=$(PLATFORM) GOARCH=$(ARCH) \
+           go build $(GOFLAGS) -ldflags '-s -w $(BUILD_LDFLAGS)'
+
+build: .init dep-vendor service-manager
+
+dep-check:
+	@which dep 2>/dev/null || go get -u -v github.com/golang/dep/cmd/dep
```
dzahariev: Dep installation is better to be a prerequisite. 
Make check for dep and print message for missing tool instead installing it.

Makefile
```diff
@@ -0,0 +1,76 @@
+# Copyright 2018 The Service Manager Authors
```
KirilKabakchiev: Seems like most Makefiles put .PHONY on targets that do not create files (that are basically commands to be executed) and that should always be ran (even if a file with the target's name is present).

Maybe it would make sense to do the same here

dzahariev: https://stackoverflow.com/questions/2145590/what-is-the-purpose-of-phony-in-a-makefile

Makefile
```diff
@@ -0,0 +1,76 @@
...
+$(BINDIR)/service-manager: .init .
+	 $(GO_BUILD) -o $@ $(PROJECT_PKG)
+
+# init creates the bin dir
+.init: $(BINDIR)
+
+$(BINDIR):
+	mkdir -p $@
+
+test: build
```
pankrator: We can make it like in service catalog: https://github.com/kubernetes-incubator/service-catalog/blob/master/Makefile#L263

This will allow for single package test executions

KirilKabakchiev: let's do it in separate PR. I will merge this one as docs uses it.

PR #149, opened 2018-10-08
Add install docs

georgifarashev:


docs/install/sm.md
```diff
@@ -1 +1,94 @@
-Two sections (SM on CF and SM on K8S)
\ No newline at end of file
+# Run Service Manager
```
KirilKabakchiev: TODO: add dep ensures / git clones

Mention to not use go get

docs/install/sm.md
```diff
@@ -1 +1,94 @@
...
+
+## Prerequisites
+
+* You need to have an OAuth server to be used by the Service Manager. In CF this could be the CF UAA.
```
dpanayotov: Can we add a short note that an easy way to get one is with Keycloak? With docker-compose for example - https://github.com/jboss-dockerfiles/keycloak/tree/master/docker-compose-examples you need to define only 1 client in the realm that can be imported from the given json configuration and then change the issuer_url to `http://localhost:8080/auth/realms/master` where `localhost:8080` is the URL of keycloak
[sm.json.txt](https://github.com/Peripli/service-manager/files/2459033/sm.json.txt)

dpanayotov: Also, this oauth server must support [OpenID Connect Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html)

docs/install/cli.md
```diff
@@ -0,0 +1,62 @@
+# Installing the smctl
+
+To use the Service Manager CLI you need to download and install it first:
+
+## Approach 1: Get the latest Service Manager CLI release
+
+You can get started with the CLI by simply downloading the latest release from [HERE][2].
+
+## Approach 2: Manual installation
```
dpanayotov: this doesn't work

docs/install/k8s-proxy.md
```diff
@@ -0,0 +1,37 @@
+# Install the service-broker-proxy-k8s
+
+## Prerequisites
+
+* `tiller` is installed and configured in the Kubernetes cluster.
```
dpanayotov: Maybe only service-catalog with a link to their installation guid which includes tiller and helm 
- https://github.com/kubernetes-incubator/service-catalog/blob/master/docs/install.md

docs/install/sm.md
KirilKabakchiev: Bump - add these

docs/install/cf-proxy.md
```diff
@@ -0,0 +1,41 @@
...
+
+In addition you can change other configurations like log level and log format.
+You can also use the `application.yml` file which has lower priority than the Environment variables.
+
+## Push
+
+Execute:
+
+```sh
+cf push -f manifest.yml
```
dpanayotov: must run `dep ensure --vendor-only` before

docs/install/cli.md
georgifarashev: I copied this from the sm cli docs. Will remove it

docs/install/sm.md
georgifarashev: > Can we add a short note that an easy way to get one is with Keycloak? With docker-compose for example - https://github.com/jboss-dockerfiles/keycloak/tree/master/docker-compose-examples you need to define only 1 client in the realm that can be imported from the given json configuration and then change the issuer_url to `http://localhost:8080/auth/realms/master` where `localhost:8080` is the URL of keycloak
> [sm.json.txt](https://github.com/Peripli/service-manager/files/2459033/sm.json.txt)

Isn't it better for this to be in the devguide?

docs/install/sm.md
georgifarashev: > Bump - add these

What should be added?

KirilKabakchiev: Dep ensures/ git clones - I can see you've already added those :)

PR #158, opened 2018-10-12
Fix setting correlation id header

dpanayotov:
X-Request-ID is also taken as a correlation id - https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Common_non-standard_request_fields 

golang's `Header.Set` uses [CanonicalMIMEHeader](https://golang.org/pkg/net/textproto/#CanonicalMIMEHeaderKey) which changes the capitalisation of ID. Also per https://tools.ietf.org/html/rfc7230#section-3.2 header keys are case-insensitive.

### DIFF
--- a/pkg/log/request.go
+++ b/pkg/log/request.go
@@ -24,7 +24,7 @@ import (
 )
 
 // CorrelationIDHeaders are the headers whose values will be taken as a correlation id for incoming requests
-var CorrelationIDHeaders = []string{"X-Correlation-ID", "X-CorrelationID", "X-ForRequest-ID"}
+var CorrelationIDHeaders = []string{"X-Correlation-ID", "X-CorrelationID", "X-ForRequest-ID", "X-Request-ID"}
 
 // CorrelationIDForRequest returns checks the http headers for any of the supported correlation id headers.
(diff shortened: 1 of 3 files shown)


pkg/log/request.go
```diff
@@ -24,13 +24,13 @@ import (
...
 // CorrelationIDHeaders are the headers whose values will be taken as a correlation id for incoming requests
-var CorrelationIDHeaders = []string{"X-Correlation-ID", "X-CorrelationID", "X-ForRequest-ID"}
+var CorrelationIDHeaders = []string{"X-Correlation-ID", "X-CorrelationID", "X-ForRequest-ID", "X-Request-ID"}
 
 // CorrelationIDForRequest returns checks the http headers for any of the supported correlation id headers.
 // The first that matches is taken as the correlation id. If none exists a new one is generated.
 func CorrelationIDForRequest(request *http.Request) string {
 	for key, val := range request.Header {
-		if slice.StringsAnyEquals(CorrelationIDHeaders, key) {
+		if slice.StringsAnyEqualsIgnoreCase(CorrelationIDHeaders, key) {
```
KirilKabakchiev: does it make sense to add a simple test(s) to validate the header case doesn't matter?

PR #165, opened 2018-10-23
Transform credentials after fetching catalog

### LINKED ISSUE #164: Cannot update service-broker credentials
If the service broker has updated it's credentials for authentication you must be able to update these credentials in the service-manager, but currently encryption of the credentials happens before fetching the catalog, so the call is made with encrypted credentials

https://github.com/Peripli/specification/blob/master/api.md#updating-a-service-broker

https://github.com/Peripli/service-manager/blob/ab6edc41606bd0987980398ae02c7b54a1ecc620/api/broker/broker_controller.go#L153

dpanayotov:
Fixes #164

### DIFF
--- a/api/broker/broker_controller.go
+++ b/api/broker/broker_controller.go
@@ -150,12 +150,12 @@ func (c *Controller) patchBroker(r *web.Request) (*web.Response, error) {
 		return nil, err
 	}
 
-	if err := transformBrokerCredentials(ctx, broker, c.Encrypter.Encrypt); err != nil {
+	catalog, err := c.getBrokerCatalog(ctx, broker)
+	if err != nil {
 		return nil, err
 	}
 
-	catalog, err := c.getBrokerCatalog(ctx, broker)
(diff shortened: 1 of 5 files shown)


api/broker/broker_controller.go
```diff
@@ -150,12 +150,12 @@ func (c *Controller) patchBroker(r *web.Request) (*web.Response, error) {
 		return nil, err
 	}
 
-	if err := transformBrokerCredentials(ctx, broker, c.Encrypter.Encrypt); err != nil {
+	catalog, err := c.getBrokerCatalog(ctx, broker)
```
KirilKabakchiev: would it make sense to add a test in the `test` folder  /semi-integration ?

There should be already a fake broker setup present there and also there is util for `				common.VerifyBrokerCatalogEndpointInvoked(brokerServer, 1)`

The test should just do a patch with new credentials and verify broker catalog is fetched successfully

PR #172, opened 2018-11-01
Improve broker tests

georgifarashev:
# Improve broker tests

## Motivation

Currently there are multiple implementations of test broker servers which brings confusion and unnecessary code duplication. 

## Approach

This PR introduces a single unified test broker server implementation to be used in all broker related tests.

### DIFF
--- a/test/common/broker.go
+++ b/test/common/broker.go
@@ -235,10 +235,12 @@ func (b *BrokerServer) defaultBindingLastOpHandler(rw http.ResponseWriter, req *
 
 func SetResponse(rw http.ResponseWriter, status int, message interface{}) {
 	rw.Header().Set("Content-Type", "application/json")
-	rw.WriteHeader(status)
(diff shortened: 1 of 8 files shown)


test/common/broker.go
```diff
@@ -1,73 +1,246 @@
...
+func (b *BrokerServer) defaultBindingLastOpHandler(rw http.ResponseWriter, req *http.Request) {
+	SetResponse(rw, http.StatusOK, Object{
+		"state": "succeeded",
+	})
 }
 
-func (b *Broker) Called() bool {
-	return b.Request != nil
+func SetResponse(rw http.ResponseWriter, status int, message interface{}) {
+	rw.Header().Set("Content-Type", "application/json")
```
KirilKabakchiev: you could remove this and in the if else reuse the pkg/util.go/WriteJSON method - also I am not sure if in general we need the extra test util method SetResponse

PR #175, opened 2018-11-06
Add util for strings suffix check

pankrator:
# Pull Request Template

## Motivation
Needed for suffix checks

## Pull Request status

* [x] Initial implementation

### DIFF
--- a/pkg/util/slice/strings.go
+++ b/pkg/util/slice/strings.go
@@ -63,3 +63,13 @@ func StringsAnyEquals(stringSlice []string, str string) bool {
 	}
 	return false
 }
+
+// StringsAnySuffix returns true if any of the strings in the slice have the given suffix.
+func StringsAnySuffix(stringSlice []string, suffix string) bool {
+	for _, v := range stringSlice {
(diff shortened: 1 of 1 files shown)


dpanayotov:
Why is this needed? Do you import it in some other project? If so, reference it

pankrator:
What do you mean by referencing it?

pkg/util/slice/strings.go
```diff
@@ -63,3 +63,13 @@ func StringsAnyEquals(stringSlice []string, str string) bool {
 	}
 	return false
 }
+
+// StringsAnySuffix returns true if any of the strings in the slice have the given suffix.
+func StringsAnySuffix(stringSlice []string, suffix string) bool {
```
KirilKabakchiev: what's the reason for adding this util (e.g. where is it needed). Also add a small test for it

dpanayotov: That's what I meant

pankrator: It is meant for internal usage.

PR #176, opened 2018-11-06
Trim bearer prefix no matter the case

pankrator:
## Motivation
if Authorization header is passed and bearer token is used, it should not matter in what case is the key word "bearer" 

## Pull Request status

* [x] Initial implementation

### DIFF
--- a/api/filters/authn/oauth/authenticator.go
+++ b/api/filters/authn/oauth/authenticator.go
@@ -96,13 +96,13 @@ func newOIDCConfig(options *options) *goidc.Config {
 // Authenticate returns information about the user by obtaining it from the bearer token, or an error if security is unsuccessful
 func (a *oauthAuthenticator) Authenticate(request *http.Request) (*web.User, security.Decision, error) {
 	authorizationHeader := request.Header.Get("Authorization")
-	if authorizationHeader == "" || !strings.HasPrefix(strings.ToLower(authorizationHeader), "bearer") {
+	if authorizationHeader == "" || !strings.HasPrefix(strings.ToLower(authorizationHeader), "bearer ") {
 		return nil, security.Abstain, nil
 	}
 	if a.Verifier == nil {
 		return nil, security.Abstain, errors.New("authenticator is not configured")
 	}
-	token := strings.TrimPrefix(authorizationHeader, "Bearer ")
+	token := authorizationHeader[len("Bearer ")-1:]
 	if token == "" {
 		return nil, security.Deny, nil
 	}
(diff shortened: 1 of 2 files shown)


dpanayotov:
Maybe a test that validates that the case doesn't matter

api/filters/authn/oauth/authenticator_test.go
```diff
@@ -313,90 +311,135 @@ var _ = Describe("OIDC Authenticator", func() {
...
 
-				BeforeEach(func() {
-					verifier = &securityfakes.FakeTokenVerifier{}
-					authenticator = &oauthAuthenticator{Verifier: verifier}
-
-					request.Header.Set("Authorization", "Bearer token")
-				})
-
-				Context("when verifier returns an error", func() {
+				Context("when Bearer is starts with uppercase", func() {
```
KirilKabakchiev: remove is

api/filters/authn/oauth/authenticator_test.go
```diff
@@ -313,90 +311,135 @@ var _ = Describe("OIDC Authenticator", func() {
...
+						request.Header.Set("Authorization", "bearer token")
+					})
 
-						claims := struct {
-							Abc string
-						}{}
-						err = user.Claims(&claims)
-						Expect(err).To(BeNil())
-						Expect(claims.Abc).To(Equal("xyz"))
+					Context("when returned token is valid", func() {
```
KirilKabakchiev: this seems pretty similar (duplicates) the test above. I don't insist but if you feel like, you could parameterize the tests (so you could cover all 3 scenarios for both lowercase and uppercase and avoid the code dup

api/filters/authn/oauth/authenticator.go
```diff
@@ -96,13 +96,13 @@ func newOIDCConfig(options *options) *goidc.Config {
...
 	authorizationHeader := request.Header.Get("Authorization")
-	if authorizationHeader == "" || !strings.HasPrefix(strings.ToLower(authorizationHeader), "bearer") {
+	if authorizationHeader == "" || !strings.HasPrefix(strings.ToLower(authorizationHeader), "bearer ") {
 		return nil, security.Abstain, nil
 	}
 	if a.Verifier == nil {
 		return nil, security.Abstain, errors.New("authenticator is not configured")
 	}
-	token := strings.TrimPrefix(authorizationHeader, "Bearer ")
+	token := strings.TrimSpace(authorizationHeader[len("Bearer "):])
```
KirilKabakchiev: not sure about this one, but as far as i remember the authentication token value (oauth2 spec) should be ignorecased. Meaning both Bearer and bearer should be expected here.

pankrator: It doesn't matter what the case is. I use the length of the "Bearer " string.

api/filters/authn/oauth/authenticator_test.go
pankrator: Good catch, I forgot that I rearranged the tests.

api/filters/authn/oauth/authenticator.go
KirilKabakchiev: Yeah with this change you added len, my bad - for some reason i was looking at the trimpredix version

PR #178, opened 2018-11-09
SM ReverseProxy: Improve logging and error messaging

KirilKabakchiev:
* Add an ErrorHandler to reverseproxy to properly add meaningful message when SB is not reachable
* Add tests to verify that SM proxying proxies all query parameters properly
* Improve logging of SM proxying logic


KirilKabakchiev:
wrong pr

PR #182, opened 2018-11-14
Limit request body size and header size with configuration

pankrator:
## Motivation
SM should be able to configure the maximum body request size and header sizes

## Pull Request status

* [x] Limit the incoming request body size
* [x] Limit the incoming request headers size
* [ ] Limit the response body size from brokers
* [ ] Unit tests

Currently there is no easy way to test request headers size, as this requires to start real http.Server, rather than a httptest.Server. Tested it with Postman and it rejects request with large header values.

### DIFF
--- a/application.yml
+++ b/application.yml
@@ -2,14 +2,15 @@ server:
   request_timeout: 4000ms
   shutdown_timeout: 4000ms
   port: 8080
+  # max_body_bytes: 4000
+  # max_header_bytes: 1000
 log:
   level: debug
   format: text
 storage:
+#  name: sm-postgres
   uri: postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable
   encryption_key: ejHjRNHbS0NaqARSRvnweVV9zcmhQEa8
-# storage:
-#  name: sm-postgres
 api:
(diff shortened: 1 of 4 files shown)


application.yml
```diff
@@ -2,16 +2,18 @@ server:
...
   format: text
 storage:
   uri: postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable
-# storage:
+#storage:
 #  name: sm-postgres
 api:
-  token_issuer_url: https://uaa.dev.cfdev.sh
-  client_id: sm
+  token_issuer_url: https://uaa.cf.sap.hana.ondemand.com
```
NickyMateev: Will **token_issuer_url** and **client_id** remain like so or is this just your local configuration?

pkg/server/server.go
```diff
@@ -45,6 +51,8 @@ func DefaultSettings() *Settings {
 		Port:            8080,
 		RequestTimeout:  time.Second * 3,
 		ShutdownTimeout: time.Second * 3,
+		MaxBodyBytes:    4 * kb,
```
NickyMateev: We should probably bump the MaxBodyBytes up to 0.5MB or 1MB

PR #184, opened 2018-11-16
Allow storage insecure connection

dpanayotov:
Add option to skip ssl check on storage connection


storage/interfaces.go
```diff
@@ -37,6 +37,7 @@ type Settings struct {
 	URI           string
 	MigrationsURL string `mapstructure:"migrations_url"`
 	EncryptionKey string `mapstructure:"encryption_key"`
+	SkipSSL       bool   `mapstructure:"skip_ssl"`
```
KirilKabakchiev: how about naming it skip_ssl_validation so that we are consistent with the api skip ssl validation?

storage/interfaces.go
```diff
@@ -45,6 +46,7 @@ func DefaultSettings() *Settings {
 		URI:           "",
 		MigrationsURL: fmt.Sprintf("file://%s/postgres/migrations", basepath),
 		EncryptionKey: "",
+		SkipSSL:       false,
```
KirilKabakchiev: I know that this is the default value, but would it make sense to add it to application.yml like we did with the api.skip_ssl_validation so that developers are aware they can change it?

Issue #185, opened 2018-11-22
SM throws ErrNotBody allowed error on APIs returning 204 No Content


NickyMateev:
# SM throws ErrNotBody error on APIs returning 204 No Content

Exact place in code where error is thrown: https://github.com/Peripli/service-manager/blob/master/api/http_handler.go#L76

This is because our api.NewJSONResponse function returns a body with a length > 0 no matter what value is passed to it.

PR #186, opened 2018-11-22
Return empty response body for APIs returning 204 No Content

NickyMateev:
# Return empty response body for APIs returning 204 No Content

Closes: #185

### DIFF
--- a/pkg/util/api_test.go
+++ b/pkg/util/api_test.go
@@ -225,16 +225,33 @@ var _ = Describe("Utils test", func() {
 		})
 
 		It("builds a web.Response containing an empty response value when return code is 204 No Content", func() {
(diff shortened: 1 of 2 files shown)


pkg/util/api_test.go
```diff
@@ -223,6 +223,16 @@ var _ = Describe("Utils test", func() {
 			Expect(response.Body).Should(MatchJSON(testTypeValid))
 			Expect(response.Header.Get("Content-Type")).To(Equal("application/json"))
 		})
+
+		It("builds a web.Response containing an empty response value when return code is 204 No Content", func() {
```
KirilKabakchiev: could you add tests for all scenarios

PR #188, opened 2018-11-23
Add flag descriptions

dpanayotov:
All configuration flags should have description tag.
Nested struct configuration descriptions are built by appending the descriptions.

Example:
```golang
type AConfig struct{
   SomeConfig BConfig `mapstructure:"first" description:"description for bconfig. "`
   CConfig `mapstructure:"second" description:"description for embedded - "`
}

type BConfig struct {
   Name string `description:"this is the name"`
}

type CConfig struct {
   Number int `description:"this is embedded"`
}
```

and this will result in:

```console
--first.name  "description for bconfig. this is the name"
--second.number "description for embbeded - this is embedded"
```

Run `go run main.go --help` to see all flags with their descriptions

There is a limitation now that flags for slice of configurations is not supported.


pkg/env/env.go
```diff
@@ -73,11 +75,23 @@ func EmptyFlagSet() *pflag.FlagSet {
...
-	traverseFields(value, "", properties)
-	for bindingName, defaultValue := range properties {
-		if set.Lookup(bindingName) == nil {
-			set.String(bindingName, cast.ToString(defaultValue), fmt.Sprintf("commandline argument for %s", bindingName))
+	parameters, descriptions := buildParametersAndDescriptions(value)
+
+	descriptionsCount := len(descriptions)
+	parametersCount := len(parameters)
+	if descriptionsCount != parametersCount {
+		log.D().Warnf("Unexpected number of descriptions found for %s: %d. "+
```
dpanayotov: ideally it would tell you what the missing description is

pkg/env/helpers.go
```diff
@@ -0,0 +1,125 @@
...
+			buffer += name + "."
+			buildDescriptionTreeWithParameters(field.Value(), tree.Children[i], buffer, result)
+			buffer = buffer[0:strings.LastIndex(buffer, name)]
+		}
+	}
+}
+
+func isValidField(field *structs.Field) bool {
+	kind := field.Kind()
+	return field.IsExported() && kind != reflect.Slice && kind != reflect.Interface && kind != reflect.Func
```
KirilKabakchiev: Shouldnt a slice be valid field ? E.x. Would a slice with clientids that is part of the config work now ?

Issue #167, opened 2018-10-28
Adopt OSB API v2.14


KirilKabakchiev:
Service Manager currently implements OSB v2.13. v2.14 introduces some new stuff (including new API endpoints). Service Manager should be up-to-date with the latest OSB version from https://github.com/openservicebrokerapi/servicebroker.

PR #190, opened 2018-11-28
Add adapt credentials endpoint

c0d1ngm0nk3y:
# Pull Request Template

## Prerequisites

* This relates to #167 and implements the adapt_credentials endpoint.

## Motivation

One step closer to implementing the issue #167 

## Approach

It forwards the additional endpoint adapt_credentials to the service broker.

## Pull Request status

* [x] Initial implementation
* [x] Unit tests

## Third-party code

n/a

## Work in Progress label

n/a

### DIFF
--- a/test/osb_test/osb_test.go
+++ b/test/osb_test/osb_test.go
@@ -462,6 +462,31 @@ var _ = Describe("Service Manager OSB API", func() {
 
 			})
 		})
+
+		Context("when call to missing service broker", func() {
+			It("should fail", func() {
+				assertMissingBrokerError(
+					ctx.SMWithBasic.POST(missingBrokerURL+"/v2/service_instances/iid/service_bindings/bid/adapt_credentials").WithHeader("X-Broker-API-Version", "oidc_authn.13").WithJSON(&object{}))
+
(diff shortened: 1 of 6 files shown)


c0d1ngm0nk3y:
@dzahariev as discussed with @loewenstein

loewenstein:
FYI @fmui

test/osb_test/osb_test.go
```diff
@@ -445,6 +446,24 @@ var _ = Describe("Service Manager OSB API", func() {
 		})
 	})
 
+	Describe("Post Binding Adapt Credentials", func() {
```
KirilKabakchiev: Would it make sense to add the other cases as well? Currently we have call to working broker and call to broken broker. We might add the cases for verifying query params are preserved as well as calls to missing broker and stopped broker.

c0d1ngm0nk3y: @KirilKabakchiev Could you check if the added tests are sufficient from your point of view?

KirilKabakchiev: looks good, thanks

PR #193, opened 2018-11-29
Fix error status codes in broker API and add additional tests

KirilKabakchiev:
Bump coverage and fix issues with response codes during create broker/ update broker catalog reconcilation


storage/postgres/abstract.go
```diff
@@ -69,7 +69,9 @@ func create(ctx context.Context, db namedExecerContext, table string, dto interf
 		strings.Join(set, ", :"),
 	)
 	log.C(ctx).Debugf("Executing query %s", query)
+	log.C(ctx).Debugf(">>>>>>VALUES: %v", dto)
```
pankrator: Can you remove this line

KirilKabakchiev: done

PR #194, opened 2018-11-30
Fetch Catalog from SM DB

KirilKabakchiev:


api/osb/osb_controller.go
```diff
@@ -35,42 +37,97 @@ import (
...
+func (c *controller) proxyHandler(r *web.Request) (*web.Response, error) {
+	return c.handler(r, c.proxy)(r)
+}
+
+func (c *controller) catalogHandler(r *web.Request) (*web.Response, error) {
+	return c.handler(r, c.catalog)(r)
 }
 
-func (c *controller) handler(r *web.Request) (*web.Response, error) {
+func (c *controller) handler(r *web.Request, f func(r *web.Request, logger *logrus.Entry, id string) (*web.Response, error)) func(request *web.Request) (*web.Response, error) {
```
NickyMateev: Why does this **handler(...)** function receive a *web.Request* argument?
Shouldn't the below returned function make use of the **request** argument rather than this **r** variable?

api/api.go
```diff
@@ -41,20 +41,6 @@ import (
 	osbc "github.com/pmorie/go-open-service-broker-client/v2"
 )
 
-// Security is the configuration used for the encryption of data
```
NickyMateev: Why is this deleted?

KirilKabakchiev: It seems its not used

api/osb/catalog_fetcher.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// FetchCatalog implements osb.CatalogFetcher and fetches the catalog for the broker with the specified broker id from SM DB
+func (scf *StorageCatalogFetcher) FetchCatalog(ctx context.Context, brokerID string) (*types.ServiceOfferings, error) {
+	catalog, err := scf.CatalogStorage.ListWithServicePlansByBrokerID(ctx, brokerID)
+	if err != nil {
+		return nil, err
+	}
+
+	// SM generates its own ids for the services and plans - currently for the platform we want to provide the original catalog id
```
georgifarashev: Idea for another PR: Is it possible to just change the names of our ids and set the catalog ids as just ID?

api/broker/broker_controller.go
```diff
@@ -173,7 +173,7 @@ func (c *Controller) listBrokers(r *web.Request) (*web.Response, error) {
 		for _, broker := range brokers {
 			offerings, err := c.Repository.ServiceOffering().ListWithServicePlansByBrokerID(ctx, broker.ID)
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("error getting catalog for broker with id %s from SM DB", broker.ID)
```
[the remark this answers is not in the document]
KirilKabakchiev: osb_controller line 111 relies on the ErrNotFoundInStorage error

PR #196, opened 2018-12-04
Visibility Labels

NickyMateev:
## Motivation

SM will allow filtering on resources based on a provided **label query**. In order to achieve that we need a utility which translates the query to SQL clauses.

This pull request sets the base for using field and label queries as well as labeling resources.
The first resource, part of this PR, is the Visibilities. All other resources will follow with a generalization for a Label, LabeledEntity and Entity interfaces which would help construct more flexible queries 

## Pull Request status

* [x] Labels API and Go implementation
* [x] FieldQuery and LabelQuery API
* [x] Create visibility with labels
* [x] List visibility by labels and fields
* [x] Fix list by label to join all labels of a visibility instead of only the queried label
* [x] Adjust operator matching
* [x] Provide means to plug selection criteria based on external state, rather than only from query
* [x] Update visibility's labels
* [x] Delete all visibilities by field query
* [x] Encode, decode, escape key/values in field query and label query
* [x] Refactoring
* [x] Unit tests

### Create Visibility
POST /v1/visibilities
Body:
```json
{
	"platform_id": "3559758d-4bd2-44fb-a4b7-805ced0c72c1",
	"service_plan_id": "4406c895-1166-4966-a0f5-646dc40a198f",
	"labels": {
		"org_id": "org_id_value",
		"cluster_id": "cluster_id_value"
	}
}
```

### List Visibilities
GET /v1/visibilities
Response:
```json
{ 
    "visibilities": [
         {
	   "platform_id": "3559758d-4bd2-44fb-a4b7-805ced0c72c1",
	   "service_plan_id": "4406c895-1166-4966-a0f5-646dc40a198f",
	   "labels": {
		"org_id": "org_id_value",
		"cluster_id": "cluster_id_value"
	     }
         }
    ]
}
```
### Delete visibilities
```
DELETE /v1/visibilities?fieldQuery=platform_id = 3559758d-4bd2-44fb-a4b7-805ced0c72c1
```

#### Queries:
- Operators
Supported operator for queries are:
```
* Equals (=):
- Works by testing whether the left operand's value and the right operand are equal
- Example: platform_id = my_platform_id
* Not Equals (!=)
- Works by testing whether the left operand's value and the right operand are not equal
- Example: platform_id != my_platform_id
* Greater Than (gt)
- Works by testing whether the left operand's value is greater than the right operand. Works only for numerical values
- Example: id gt 5
* Less Than (lt)
- Works by testing whether the left operand's value is less than the right operand. Works only for numerical values
- Example: id lt 5
* In (in)
- Works by testing whether the left operand's value is contained in the right operand. Works only for list values of the right operand contained in square braces.
- Example: id in [5|6|7]
* Not In (notin)
- Works by testing whether the left operand's value is NOT contained in the right operand. Works only for list values of the right operand contained in square braces.
- Example: id notin [1|2|3]
* Equals Or Nil (eqornil)
- Works by testing whether the left operand's value is equal to the right operand OR the left operand's value is NULL
- Example: platform_id eqornil my_platform_id
```
- Field Query
```
GET /v1/visibilities?fieldQuery=platform_id = my_platform_id
```

- Label Query
```
GET /v1/visibilities?labelQuery=org_id = my_org_id
```

- Mixed Query

```
GET /v1/visibilities?fieldQuery=platform_id = my_platform_id&labelQuery=org_id = my_org_id
```

### Authorization:
When listing all visibilities with platform authorization you will get in return all visibilities ONLY for your platform and all PUBLIC visibilities

In other words, the following two are equivalent:
```
GET /v1/visibilities
Authorization: Basic <my_platform_credentials>
```
and
```
GET /v1/visibilities?fieldQuery=platform_id eqornil my_platform_id
Authorization: Bearer ...
```

Public visibilities (plans that are made public to everyone) have platform_id set to empty

## Update Visibility and its labels

```
PATCH /v1/visibilities/:visibility_id
```
Body:
```json
{
	"platform_id": "my-platform-id",
	"labels": [
		{
			"op": "add",
			"key": "org_id",
			"values": ["my-org-id"]
		},
		{
			"op": "add_values",
			"key": "cluster_id",
			"values": ["my-cluster-id"]
		},
		{
			"op": "remove",
			"key": "my-key"
		},
		{
			"op": "remove_values",
			"key": "other-key",
			"values": ["value1", "value2"]
		}
	]
}
```


KirilKabakchiev:
How do we handle multiple values behind rhe same key ? E.g. should the api reaponse contain an array of values for each label key ?

storage/postgres/abstract.go
```diff
@@ -86,21 +89,91 @@ func create(ctx context.Context, db pgDB, table string, dto interface{}) (string
...
+	if err != nil {
+		return err
+	}
+	return db.SelectContext(ctx, entity, sqlQuery, queryParams...)
+}
+
+func deleteAllByFieldCriteria(ctx context.Context, extContext sqlx.ExtContext, table string, dto interface{}, criteria ...query.Criterion) error {
+	for _, criterion := range criteria {
+		if criterion.Type != query.FieldQuery {
+			return &query.UnsupportedQuery{Message: "conditional delete is only supported for field queries"}
```
KirilKabakchiev: Here is one idea to support delete by labelquery and fieldquery and no query - all deletes will use subquery

1. DELETE FROM vis WHERE id IN (SELECT visibility_id FROM vis) - no criterias means to where clause

2. DELETE FROM vis WHERE id IN (SELECT visibility_id FROM vis_labels WHERE key = org_id AND val IN [1,2,3,4]) - labelquery means buildQueryWithParams func builds us a where clause for labels

3. DELETE FROM vis WHERE id IN (SELECT visibility_id FROM vis WHERE field1 = value AND field2 IN [v1,v2,v3,v4] - fieldquery means we use vis table instead of vis_labels in the subquery and we enter the else of buildQueryWithParams func when building the where clause

4.  DELETE FROM vis WHERE id IN (SELECT from 2.) OR id IN (SELECT from 3.) - in case both fieldquery and labelquery - append them with or/union

And another way would be to have a switch with fallthrough in the deleteAllByFieldCriteria and in both cases we would use buildqueryWithParams to build the where clause but in the labelQuery case the where clause would be prepended with WHERE id IN (SELECT id from <table> <whereclause-from-before>) or with AND id IN (SELECT id from <table> <whereclause-from-before>) if we entered the first switch case (the fieldquery one)

So the statement would look like DELETE FROM vis WHERE field=val AND field2 IN [...] AND id IN (SELECT id from <table> <whereclause-from-before>)

storage/postgres/visibility.go
```diff
@@ -29,7 +35,30 @@ type visibilityStorage struct {
 func (vs *visibilityStorage) Create(ctx context.Context, visibility *types.Visibility) (string, error) {
 	v := &Visibility{}
 	v.FromDTO(visibility)
-	return create(ctx, vs.db, visibilityTable, v)
+	id, err := create(ctx, vs.db, visibilityTable, v)
+	if err != nil {
+		return "", err
+	}
+	return id, vs.createLabels(ctx, id, visibility.Labels)
```
pankrator: are the 2 queries executed in transaction?

pankrator: It might be somewhere else, I just cannot find it

dpanayotov: Yes, they are - https://github.com/Peripli/service-manager/blob/query-interpreter/api/visibility/visibility_controller.go#L58

pkg/query/selection.go
```diff
@@ -0,0 +1,245 @@
...
+	// OpenBracket is the token that denotes the end of a multivariate operand
+	CloseBracket string = "]"
+)
+
+// CriterionType is a type of criteria to be applied when querying
+type CriterionType string
+
+const (
+	// FieldQuery denotes that the query should be executed on the entity's fields
+	FieldQuery CriterionType = "fieldQuery"
```
KirilKabakchiev: It would make sense to support fieldquery out of the box/by default for new apis - e.g. without actually supporting labels. I think that currently one would need to add labels support to the resource which would introduce field querying, too - which is a bit strange

test/visibility_test/visibility_test.go
```diff
@@ -17,9 +17,14 @@
 package visibility_test
 
 import (
+	"fmt"
```
KirilKabakchiev: a general question regarding testing labeling and field querying - every type will support field querying and most types will support labeling - so we would probably need to copy the same 1k lines of tests for each resource type ? Would it make sense to put labeling and field querying tests in a separate package and parameterize them so that we can just add the new api type when to want to add tests for either labeling or fieldquerying?

pkg/query/selection.go
dpanayotov: Not necessarily. If you don't implement createLabels and updateLabels, you can add 1 line of code that checks if the criteria passed to List contains LabelQuery and you can return an error. Everything else would work fine

test/visibility_test/visibility_test.go
dpanayotov: i agree, but it would be better if it comes in a new PR where all entities begin to support labels

storage/postgres/abstract.go
dpanayotov: It would be nice to support this, but I'm not sure if we can't leave it in a new PR

storage/postgres/visibility.go
```diff
@@ -29,62 +37,128 @@ type visibilityStorage struct {
...
 }
 
-func (vs *visibilityStorage) List(ctx context.Context) ([]*types.Visibility, error) {
-	var visibilities []Visibility
-	err := list(ctx, vs.db, visibilityTable, map[string][]string{}, &visibilities)
-	if err != nil || len(visibilities) == 0 {
-		return []*types.Visibility{}, err
+func (vs *visibilityStorage) createLabels(ctx context.Context, visibilityID string, labels []*types.VisibilityLabel) error {
+	for _, label := range labels {
+		label.ServiceVisibilityID = visibilityID
```
KirilKabakchiev: when creating labels , createdat/updatedat , ServiceVisibilityID and the label id are all set in different places across the code and this is kind of confusing

pkg/query/selection.go
```diff
@@ -0,0 +1,245 @@
...
+	// NotEqualsOperator takes two operands and tests if they are not equal
+	NotEqualsOperator Operator = "!="
+	// GreaterThanOperator takes two operands and tests if the left is greater than the right
+	GreaterThanOperator Operator = "gt"
+	// LessThanOperator takes two operands and tests if the left is lesser than the right
+	LessThanOperator Operator = "lt"
+	// InOperator takes two operands and tests if the left is contained in the right
+	InOperator Operator = "in"
+	// NotInOperator takes two operands and tests if the left is not contained in the right
+	NotInOperator Operator = "notin"
```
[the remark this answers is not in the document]
dpanayotov: It would, because the matching takes into account the decoded spaces around the operator. As long as you pass a query with the pattern "left_operand operator [right_operand|...]" it would be okay. The whitespaces around the operator are the ones setting its position

pkg/query/selection.go
```diff
@@ -0,0 +1,245 @@
...
+	if c.Operator.IsNumeric() && !isNumeric(c.RightOp[0]) {
+		return &UnsupportedQueryError{Message: fmt.Sprintf("%s is numeric operator, but the right operand is not numeric", c.Operator)}
+	}
+	if slice.StringsAnyEquals(c.RightOp, "") {
+		return &UnsupportedQueryError{Message: "right operand must have value"}
+	}
+	return nil
+}
+
+type criteria []Criterion
```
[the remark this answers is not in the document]
dpanayotov: only for the add function. can be removed

dpanayotov:
`visibilities?fieldQuery=platform_id+eqornil+%3Deee1ff71-78c0-461b-bcce-8a2b830d126c` will fail, because %3D will be decoded to whitespace which is invalid. One option is to re-encode the query and replace all encodings of special symbols with their actual representation (these are `[ ] , = `)

pkg/query/selection.go
```diff
@@ -195,16 +197,26 @@ func BuildCriteriaFromRequest(request *web.Request) ([]Criterion, error) {
...
+	escapeReplacer          = strings.NewReplacer(escapedOpenBracket, OpenBracket, escapedCloseBracket, CloseBracket, escapedSeparator, string(Separator), escapedOperandSeparator, OperandSeparator, "%3D", "=")
+)
+
 func process(input string, criteriaType CriterionType) ([]Criterion, error) {
 	var c []Criterion
 	if input == "" {
 		return c, nil
 	}
+	escapedInput := url.QueryEscape(input)
+	escapedInput = escapeReplacer.Replace(escapedInput)
```
KirilKabakchiev: not sure if this works as initally desired. This look especially confusing when for example one tries querying say service_plans by description that equals Shared fake Server, 5tb persistent disk, 40 max concurrent connections. 100 async.  which would be the similar to the example description in the osb spec

pkg/types/labels.go
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package types
+
+import "time"
+
+type Label struct {
```
KirilKabakchiev: We should consider if we want to do some restrictions regarding what symbols can be contained in label keys and values and whatever we decide should be introduced with this change so that we can avoid having label migrations afterwards

api/filters/selection_criteria.go
```diff
@@ -0,0 +1,66 @@
...
+type SelectionCriteria struct {
+}
+
+// Name implements the web.Filter interface and returns the identifier of the filter.
+func (*SelectionCriteria) Name() string {
+	return CriteriaFilterName
+}
+
+// Run represents the selection criteria middleware function that processes the request and configures the request-scoped selection criteria.
+func (l *SelectionCriteria) Run(req *web.Request, next web.Handler) (*web.Response, error) {
```
KirilKabakchiev: Are we going to support field querying on all entity fields because right now this filter will try to build criterions for everything that has been provided in the query. We might or might not want to introduce a "per entity" field query whitelist that specifies on what can an entity/resource be field-queried. (that is unless we can come up with parsing logic that works for all cases

pkg/query/selection.go
```diff
@@ -0,0 +1,299 @@
...
+	NotInOperator, GreaterThanOperator, LessThanOperator, EqualsOrNilOperator}
+
+const (
+	// OpenBracket is the token that denotes the beginning of a multivariate operand
+	OpenBracket string = "["
+	// OpenBracket is the token that denotes the end of a multivariate operand
+	CloseBracket string = "]"
+	// Separator is the separator between field and label queries
+	Separator        rune   = ','
+	OperandSeparator string = "+"
```
KirilKabakchiev: godoc (consistency)

pkg/query/selection.go
```diff
@@ -0,0 +1,299 @@
...
+	if len(c.RightOp) > 1 && !c.Operator.IsMultiVariate() {
+		return fmt.Errorf("multiple values received for single value operation")
+	}
+	if c.Operator.IsNullable() && c.Type != FieldQuery {
+		return &UnsupportedQueryError{"nullable operations are supported only for field queries"}
+	}
+	if c.Operator.IsNumeric() && !isNumeric(c.RightOp[0]) {
+		return &UnsupportedQueryError{Message: fmt.Sprintf("%s is numeric operator, but the right operand is not numeric", c.Operator)}
+	}
+	//if slice.StringsAnyEquals(c.RightOp, "") {
```
KirilKabakchiev: clean up or uncomment, also add a test for the case  (per op?) in case one is missing

pkg/query/selection_test.go
```diff
@@ -0,0 +1,171 @@
...
+			})
+			Specify("Field query with duplicate key", func() {
+				var err error
+				ctx, err = AddCriteria(ctx, validCriterion)
+				Expect(err).ToNot(HaveOccurred())
+				addInvalidCriterion(ByField(EqualsOrNilOperator, validCriterion.LeftOp, "right op"))
+			})
+		})
+
+		Context("Valid", func() {
```
KirilKabakchiev: would it make sense to extend these tests to include cases for all operators?

pkg/query/selection_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+		Context("When there is an invalid field query", func() {
+			It("Should return an error", func() {
+				criteriaFromRequest, err := buildCriteria("http://localhost:8080/v1/visibilities?fieldQuery=leftop+lt+rightop")
+				Expect(err).To(HaveOccurred())
+				Expect(criteriaFromRequest).To(BeNil())
+			})
+		})
+
+		Context("When passing multivariate query", func() {
```
KirilKabakchiev: Same as above  - would it make sense to somehow parameterize threse tests to cover all operations that are defined

storage/interfaces.go
```diff
@@ -214,20 +216,14 @@ type Visibility interface {
...
-	List(ctx context.Context) ([]*types.Visibility, error)
+	List(ctx context.Context, criteria ...query.Criterion) ([]*types.Visibility, error)
 
 	// Delete deletes a visibility from SM DB
-	Delete(ctx context.Context, id string) error
+	Delete(ctx context.Context, criteria ...query.Criterion) error
 
 	// Update updates a visibility from SM DB
-	Update(ctx context.Context, visibility *types.Visibility) error
+	Update(ctx context.Context, visibility *types.Visibility, labelChanges ...*query.LabelChange) error
```
KirilKabakchiev: not sure if this is very user friendly - in case we want to introduce label changes only we would need to first go Get to obtain the resource and then use this method

storage/postgres/abstract.go
```diff
@@ -86,21 +89,100 @@ func create(ctx context.Context, db pgDB, table string, dto interface{}) (string
...
+	if labelsEntity == nil {
+		baseQuery = constructBaseQueryForEntity(baseTableName)
+	} else {
+		baseQuery = constructBaseQueryForLabelable(labelsEntity, baseTableName)
+	}
+	sqlQuery, queryParams, err := buildQueryWithParams(db, baseQuery, baseTableName, labelsEntity, criteria)
+	if err != nil {
+		return nil, err
+	}
+	log.C(ctx).Debugf("Executing query %s", sqlQuery)
```
KirilKabakchiev: Log the query and the parameters as well

storage/postgres/labels.go
```diff
@@ -0,0 +1,210 @@
...
+		for _, option := range labelCriteria {
+			rightOpBindVar, rightOpQueryValue := buildRightOp(option)
+			sqlOperation := translateOperationToSQLEquivalent(option.Operator)
+			labelQueries = append(labelQueries, fmt.Sprintf("%[1]s.key = ? AND %[1]s.val %[2]s %s", labelTableName, sqlOperation, rightOpBindVar))
+			queryParams = append(queryParams, option.LeftOp, rightOpQueryValue)
+		}
+		labelSubQuery += strings.Join(labelQueries, " AND ")
+		labelSubQuery += "))"
+
+		sqlQuery = strings.Replace(sqlQuery, "LEFT JOIN", "JOIN "+labelSubQuery, 1)
```
KirilKabakchiev: the replace is a bit strange

storage/postgres/migrations/005_visibility_labels.up.sql
```diff
@@ -0,0 +1,14 @@
+BEGIN;
+
+CREATE TABLE visibility_labels
+(
+  id            varchar(100) PRIMARY KEY,
+  key           varchar(255) NOT NULL,
+  val           varchar(255) NOT NULL,
+  visibility_id varchar(100) REFERENCES visibilities (id) ON DELETE CASCADE,
```
KirilKabakchiev: NOT NULL for visiblity_id ?

storage/postgres/types.go
```diff
@@ -121,6 +129,79 @@ type Visibility struct {
 	UpdatedAt     time.Time      `db:"updated_at"`
 }
 
+type Labelable interface {
```
KirilKabakchiev: maybe some godoc

storage/postgres/types.go
```diff
@@ -121,6 +129,79 @@ type Visibility struct {
...
+	Label() (labelTableName string, referenceColumnName string, primaryColumnName string)
+}
+
+type visibilityLabels []*VisibilityLabel
+
+func (vls visibilityLabels) Validate() error {
```
KirilKabakchiev: implement on pointer or not for all methods (in order to be consistent)

pkg/query/selection.go
```diff
@@ -0,0 +1,299 @@
...
+	Separator        rune   = ','
+	OperandSeparator string = "+"
+)
+
+// CriterionType is a type of criteria to be applied when querying
+type CriterionType string
+
+const (
+	// FieldQuery denotes that the query should be executed on the entity's fields
+	FieldQuery CriterionType = "fieldQuery"
```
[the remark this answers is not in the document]
dpanayotov: I think not providing it and providing empty value is the same

pkg/query/selection.go
```diff
@@ -0,0 +1,299 @@
...
+	// LabelQuery denotes that the query should be executed on the entity's labels
+	LabelQuery CriterionType = "labelQuery"
```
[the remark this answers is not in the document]
dpanayotov: as above

storage/interfaces.go
dpanayotov: That's what the initial idea of the labels was - they are not updatable by themselves, but only via the entity they're labelling. Moreover, you might have the same label (key-value) labelling different visibilities, so the only way to update the label for the first visibility is to either know the label's id or the associated visibility

pkg/query/selection.go
```diff
@@ -0,0 +1,299 @@
...
+)
+
+// IsMultiVariate returns true if the operator requires right operand with multiple values
+func (op Operator) IsMultiVariate() bool {
+	return op == InOperator || op == NotInOperator
+}
+
+// IsNullable returns true if the operator can check if the left operand is nil
+func (op Operator) IsNullable() bool {
+	return op == EqualsOrNilOperator
```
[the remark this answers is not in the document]
dpanayotov: the nil part will be ignored, so it will have the same effect as EqualsOperator. I don't think this is enough of a reason to separate into different operators

storage/postgres/labels.go
dpanayotov: I don't understand. Change from LEFT JOIN to JOIN and append the remaining query exactly 1 times.

pkg/query/update.go
```diff
@@ -0,0 +1,92 @@
...
+	if err := json.Unmarshal([]byte(labelChangesBytes), &labelChanges); err != nil {
+		return nil, err
+	}
+	for _, v := range labelChanges {
+		if v.Operation == RemoveLabelOperation {
+			v.Values = nil
+		}
+	}
+	for _, change := range labelChanges {
+		if err := change.Validate(); err != nil {
```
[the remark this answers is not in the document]
dpanayotov: In order to hide this, there should be a constructor for example. But then, a constructor doesn't make sense with exported fields IMO (there's ByField and ByLabel which could also return an error)

storage/postgres/labels.go
```diff
@@ -0,0 +1,210 @@
...
+func removeLabel(ctx context.Context, execer sqlx.ExtContext, labelable Labelable, referenceID, labelKey string, labelValues ...string) error {
+	labelTableName, referenceColumnName, _ := labelable.Label()
+	baseQuery := fmt.Sprintf("DELETE FROM %s WHERE key=? AND %s=?", labelTableName, referenceColumnName)
+	labelValuesCount := len(labelValues)
+	// remove all labels with this key
+	if labelValuesCount == 0 {
+		return executeNew(ctx, execer, baseQuery, []interface{}{labelKey, referenceID})
+	}
+	args := []interface{}{labelKey, referenceID}
+	if labelValuesCount == 1 {
```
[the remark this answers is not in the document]
dpanayotov: leftover

pkg/types/labels.go
dpanayotov: If an entity doesn't support labels, the storage won't implement the additional logic regarding labels.

KirilKabakchiev: I was thinking for some general validation that would apply for all label types later on - if we can think of any right now, we should probably implement it now before we have stored labels

storage/postgres/labels.go
KirilKabakchiev: I understand the query but what I mean that reconstructing the query seems strange - it kind of breaks the flow of appending stuff by switching the join type - its not a problem just looks strange

pkg/query/update.go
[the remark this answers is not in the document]
KirilKabakchiev: we need the validation logic in the storage layer for the internal apis

pkg/query/selection.go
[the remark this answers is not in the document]
KirilKabakchiev: eqornil and eq would work similarly for labels but its documented that they work differently which is confusing, someone might expect that eqornil works as specified since it can be used for labels - i dont insist on separating but it just looks cleaner and less confusing (and its just the operators and not the querying logic itself)

api/visibility/visibility_controller.go
```diff
@@ -111,13 +146,23 @@ func (c *Controller) patchVisibility(r *web.Request) (*web.Response, error) {
...
-	visibility, err := c.VisibilityStorage.Get(ctx, visibilityID)
+	visibility, err := c.Repository.Visibility().Get(ctx, visibilityID)
 	if err != nil {
-		return nil, util.HandleStorageError(err, "visibility", visibilityID)
+		return nil, util.HandleStorageError(err, "visibility")
 	}
 
 	createdAt := visibility.CreatedAt
 
+	changes, err := query.LabelChangesFromJSON(r.Body)
```
KirilKabakchiev: why unmarshal the labels and the visibility itself in two parts ?

api/filters/selection_criteria.go
dpanayotov: Could be a part of a new PR. 
```golang
type Person struct{
    Name QueryableString `json:"name"`
    Age QueryableInt `son:"
}
```

pkg/types/labels.go
dpanayotov: at best it can be extracted in this struct, but every other label that embeds it, should call validate

storage/postgres/abstract.go
dpanayotov: Not sure if it's okay to log the parameters

api/visibility/visibility_controller.go
dpanayotov: because that's not labels, but label changes

pkg/types/visibility.go
```diff
@@ -33,11 +33,17 @@ type Visibilities struct {
...
-	PlatformID    string    `json:"platform_id"`
-	ServicePlanID string    `json:"service_plan_id"`
-	CreatedAt     time.Time `json:"created_at"`
-	UpdatedAt     time.Time `json:"updated_at"`
+	ID            string             `json:"id"`
+	PlatformID    string             `json:"platform_id"`
+	ServicePlanID string             `json:"service_plan_id"`
+	CreatedAt     time.Time          `json:"created_at"`
+	UpdatedAt     time.Time          `json:"updated_at"`
+	Labels        []*VisibilityLabel `json:"labels,omitempty"`
```
KirilKabakchiev: would it make sense for this to be a map instead ? since we have uniqueness on keys and possibly , when one reads the visibilities with labels from db and if the purpose is processing something based on labels , you would probably be interested in particular label (labelkey)

dpanayotov: something like `type Labels map[string][]string`? Marshalling and unmarshalling would be a hassle then probably, because you'd have to define a helper struct.

PR #197, opened 2018-12-04
Visibilities

KirilKabakchiev:
* CRUD APIs on visibilities and add tests for visibility apis
* database layer for visibilities
* fix a bug during catalog reconcilation during update broker op - new plans for new services were not stored in sm db and add more tests for catalog reconcilation
* add Returning id on insert sql queries and improve errors returned from sql queries when a constraint violation happen (now returns bad request instead of internal server error)
* rework authn user context and include platform details in context during basic authn
* secure visibiity apis: list api for visibilities returns returns a filtered list based on the platform id if the authn is basic; list api returns all visibilities for token authn


api/visibility/visibility_controller.go
```diff
@@ -0,0 +1,140 @@
...
+	currentTime := time.Now().UTC()
+	visibility.CreatedAt = currentTime
+	visibility.UpdatedAt = currentTime
+
+	id, err := c.VisibilityStorage.Create(ctx, visibility)
+	if err != nil {
+		return nil, util.HandleStorageError(err, "visibility", visibility.ID)
+	}
+
+	logger.Errorf("id is %s", id)
```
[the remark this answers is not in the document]
KirilKabakchiev: left overs, removed it

api/visibility/visibility_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	log.C(ctx).Debugf("Updating visibility  with id %s", visibilityID)
+
+	visibility, err := c.VisibilityStorage.Get(ctx, visibilityID)
+	if err != nil {
+		return nil, util.HandleStorageError(err, "visibility", visibilityID)
+	}
+
+	createdAt := visibility.CreatedAt
+
+	if err := util.BytesToObject(r.Body, visibility); err != nil {
```
dpanayotov: this looks like a PUT, not a PATCH

application.yml
```diff
@@ -13,6 +13,6 @@ storage:
   encryption_key: ejHjRNHbS0NaqARSRvnweVV9zcmhQEa8
   skip_ssl_validation: false
 api:
-  token_issuer_url: http://localhost:8080/uaa
-  client_id: sm
+  token_issuer_url: https://uaa.cf.sap.hana.ondemand.com
```
dpanayotov: revert whole file

storage/interfaces.go
```diff
@@ -195,13 +199,34 @@ type ServicePlan interface {
...
 
-	// Update updates a service service_plan from SM DB
+	// Update updates a service plan from SM DB
 	Update(ctx context.Context, servicePlan *types.ServicePlan) error
 }
 
+// Visibility interface for Visibility db operations
+type Visibility interface {
+	// Create stores a visibility in SM DB
+	Create(ctx context.Context, servicePlan *types.Visibility) (string, error)
```
dpanayotov: servicePlan -> visibility

storage/interfaces.go
```diff
@@ -195,13 +199,34 @@ type ServicePlan interface {
...
+	ListByPlatformID(ctx context.Context, platformID string) ([]*types.Visibility, error)
+
+	// List retrieves all visibilities from SM DB
+	List(ctx context.Context) ([]*types.Visibility, error)
+
+	// Delete deletes a visibility from SM DB
+	Delete(ctx context.Context, id string) error
+
+	// Update updates a visibility from SM DB
+	Update(ctx context.Context, servicePlan *types.Visibility) error
```
dpanayotov: same

test/visibility_test/visibility_test.go
```diff
@@ -0,0 +1,527 @@
...
+				ctx.SMWithOAuth.POST("/v1/visibilities").
+					WithJSON(common.Object{
+						"service_plan_id": existingPlanIDs[1],
+					}).
+					Expect().Status(http.StatusCreated).JSON().Object().ContainsMap(common.Object{
+					"service_plan_id": existingPlanIDs[1],
+				})
+			})
+
+			It("returns 400 if visibilities for the plan exist", func() {
```
dpanayotov: how do you make a plan public then? DELETE request for N visibilities with N platform_ids and service_plan and then a new POST?

test/visibility_test/visibility_test.go
```diff
@@ -0,0 +1,527 @@
...
+				ctx.SMWithOAuth.GET("/v1/visibilities/"+existingVisibilityID).
+					Expect().
+					Status(http.StatusOK).JSON().Object().
+					ContainsKey("created_at").
+					ValueNotEqual("created_at", createdAt)
+			})
+		})
+
+		Context("when updated_at is in the body", func() {
+			It("should not update created_at", func() {
```
dpanayotov: "should not update updated_at"

Also, maybe add a check that updated_at is changed, but is not the one provided in the body

api/visibility/visibility_controller.go
KirilKabakchiev: it takes whatever is currently stored and ontop of it overrides whatever is provided in the request, keeping whatever is not provided in the request as it was.That's patching, right?

test/visibility_test/visibility_test.go
KirilKabakchiev: you would need to clean up all visibilities related to the plan before you can create a public visibility.

I am up for alternative suggestions as well.

test/visibility_test/visibility_test.go
KirilKabakchiev: the check is there - line 436 - ValueNotEqual("updated_at", updatedAt) - or do you have something else in mind ?

dpanayotov: I meant GET /visibilities/<visibility_id> and get the updated_at field before the update. Update the visibility. execute GET/visibilities/<visibility_id> and validate that the updated_at is different that the one before the update.

test/visibility_test/visibility_test.go
dpanayotov: Should be noted in specification

api/visibility/visibility_controller.go
KirilKabakchiev: I will add a test for partial updates

test/visibility_test/visibility_test.go
KirilKabakchiev: problem is no time passes between the requests, so we need to sleep which im not sure is very good option

```
updatedAt := "2015-01-01T00:00:00Z"

				currentUpdateAtValue := ctx.SMWithOAuth.GET("/v1/visibilities/" + existingVisibilityID).
					Expect().
					Status(http.StatusOK).JSON().Object().Value("updated_at").String().Raw()

				time.Sleep(1 * time.Second) // without this, the test would fail
				ctx.SMWithOAuth.PATCH("/v1/visibilities/"+existingVisibilityID).
					WithJSON(common.Object{
						"updated_at": updatedAt,
					}).
					Expect().
					Status(http.StatusOK).JSON().Object().
					ContainsKey("updated_at").
					ValueNotEqual("updated_at", updatedAt).
					ValueNotEqual("updated_at", currentUpdateAtValue)

				ctx.SMWithOAuth.GET("/v1/visibilities/"+existingVisibilityID).
					Expect().
					Status(http.StatusOK).JSON().Object().
					ContainsKey("updated_at").
					ValueNotEqual("updated_at", updatedAt).
					ValueNotEqual("updated_at", currentUpdateAtValue)
			})
```

PR #198, opened 2018-12-10
Free Plans Visibilities Filter

KirilKabakchiev:
The filter handles creation and deletion of public visibilities for free plans. For the relevant scenarios that filter covers, please have a look at the tests.

Currently the filter is enabled by default which means that SM would create one public visibility for each free plan by default. (unless the filter is disabled). Whoever processes the visibilities (one candidate would be the SM proxy) should decide how to handle the public visibilities in the corresponding platform)


api/filters/free_plans.go
```diff
@@ -0,0 +1,133 @@
...
+				if err != nil {
+					return err
+				}
+				for _, visibility := range visibilitiesForPlan {
+					if isFree {
+						if visibility.PlatformID == "" {
+							hasPublicVisibility = true
+							continue
+						} else {
+							if err := vRepository.Delete(ctx, visibility.ID); err != nil {
```
dpanayotov: maybe a test that shows that if there are visibilities for this public plan, they are deleted and only one visibility is left?

KirilKabakchiev: you cant have visibilities for paid plans directly

Its more like if the plan was paid and there were visibilities for it and then it was made public, the paid visibilities were deleted and a public visibility was added

Issue #203, opened 2018-12-17
SM osb endpoint returns invalid catalog


georgifarashev:
# `v1/osb/<broker_ID>/v2/catalog` endpoint returns invalid catalog

**The issue:**
When requesting the catalog of a broker through service manager the returned result is an invalid catalog.
More specifically if the catalog does not specify `service.tags` then the endpoint returns json with `service.tags=null`.
The original catalog does not contain `service.tags`.
This behavior can be observed for:

* `service.requires`
* `service.metadata`
* `service.tags`
* `service.plan.metadata`

Example catalog:
```json
{
    "services": [
        {
            "created_at": "2018-12-12T12:51:45Z",
            "updated_at": "2018-12-12T12:51:45Z",
            "id": "1f400825-1234-5678-8013-dacf63fcd648",
            "name": "dummy1",
            "description": "Dummy broker",
            "bindable": true,
            "instances_retrievable": false,
            "bindings_retrievable": false,
            "plan_updateable": false,
            "catalog_id": "1f400825-1234-5678-8013-dacf63fcd648",
            "catalog_name": "dummy1",
            "tags": null,
            "requires": null,
            "metadata": null,
            "broker_id": "ce667a38-aec6-48b0-874c-ed96b8cd46be",
            "plans": [
                {
                    "created_at": "2018-12-12T12:51:45Z",
                    "updated_at": "2018-12-12T12:51:45Z",
                    "id": "1f400825-1434-5278-9913-dfcf63fcd647",
                    "name": "dummy1plan1",
                    "description": "dummy 1 example plan 1",
                    "catalog_id": "1f400825-1434-5278-9913-dfcf63fcd647",
                    "catalog_name": "dummy1plan1",
                    "free": false,
                    "bindable": true,
                    "plan_updateable": false,
                    "metadata": null,
                    "schemas": {},
                    "service_offering_id": "157bfa96-8425-45d6-9ceb-9f7e98ce58de"
                } ]
        } ]
}
```

PR #204, opened 2018-12-18
Fix broken Catalog when some properties are missing

georgifarashev:
# Fix broken Catalog when some properties are missing

Fixes: #203


storage/postgres/types.go
```diff
@@ -293,3 +293,20 @@ func (v *Visibility) FromDTO(visibility *types.Visibility) {
 		v.PlatformID.Valid = true
 	}
 }
+
+func getJSONText(item json.RawMessage) sqlxtypes.JSONText {
+	if len(item) == len("null") && string(item) == "null" {
```
dpanayotov: can't the `len(item) == len("null")` check be omitted?

georgifarashev: If the `item` is very large this will prevent it from an unnecessary unmarshalling.

PR #206, opened 2018-12-21
New configuration flag descriptions

dpanayotov:
All configuration flags should have description tag.
Nested struct configuration descriptions are built by appending the descriptions.

Example:
```golang
type AConfig struct{
   SomeConfig BConfig `mapstructure:"first" description:"description for bconfig. "`
   CConfig `mapstructure:"second" description:"description for embedded - "`
}

type BConfig struct {
   Name string `description:"this is the name"`
}

type CConfig struct {
   Number int `description:"this is embedded"`
}
```

and this will result in:

```console
--first.name  "description for bconfig. this is the name"
--second.number "description for embbeded - this is embedded"
```

Run `go run main.go --help` to see all flags with their descriptions

There is a limitation now that flags for slice of configurations is not supported.

### DIFF
--- a/pkg/env/helpers.go
+++ b/pkg/env/helpers.go
@@ -74,6 +74,9 @@ func buildDescriptionPaths(root *descriptionTree, path []*descriptionTree) []str
 		for _, node := range path {
 			res += node.Value
 		}
+		if res == "" {
+			res = "external configuration or no description provided"
+		}
 		if res != "" {
 			result = append(result, res)
 		}
@@ -97,7 +100,8 @@ func buildDescriptionTreeWithParameters(value interface{}, tree *descriptionTree
 		return
 	}
 	s := structs.New(value)
-	for i, field := range s.Fields() {
+	k := 0
+	for _, field := range s.Fields() {
 		if isValidField(field) {
 			var name string
 			if field.Tag("mapstructure") != "" {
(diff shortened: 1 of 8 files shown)


pkg/env/helpers.go
```diff
@@ -0,0 +1,130 @@
...
+		return
+	}
+	t.Children = append(t.Children, tree)
+}
+
+type configurationParameter struct {
+	Name         string
+	DefaultValue interface{}
+}
+
```
NickyMateev: Does it make sense to refactor the next two functions like so:

```go
func buildParametersAndDescriptions(value interface{}) ([]configurationParameter, []string) {
	tree := &descriptionTree{}
	return buildParameters(tree, value), buildDescriptionsFromTree(tree)
}

func buildParameters(tree *descriptionTree, value interface{}) []configurationParameter {
	var parameters []configurationParameter
	buildDescriptionTreeWithParameters(value, tree, "", &parameters)
	return parameters
}
```

This way we can eliminate some code duplication. Then all clients can use the **buildParametersAndDescriptions** and if they don't need the descriptions, they can simply ignore the second variable - like here: https://github.com/Peripli/service-manager/blob/067e7c84d83885e084d04be62980f18de31ea53a/pkg/env/env.go#L114

PR #211, opened 2019-01-14
Remove catalog query parameter from list brokers

KirilKabakchiev:
This should wait as delivery of k8s proxy is closed.

The `?catalog=true` query paramater in `/v1/service_brokers` was a workaround to set visibilities in CF proxy and is no longer required. This PR removes it.


KirilKabakchiev:
@pankrator  validated that this should not be breaking for the k8s proxy so we might merge it I think

PR #212, opened 2019-01-14
Disallow duplicate label querying

KirilKabakchiev:
Currently label querying allows `duplicate label query keys` and disallows `duplicate field query keys` which is rather inconsistent. This PR disallows duplicate label querying. (e.x. labelQuery=key-op-val|key-op-val is disallowed and  labelQuery=key-op-val|key-op-val2  is not allowed)

### DIFF
--- a/pkg/query/selection.go
+++ b/pkg/query/selection.go
@@ -19,7 +19,6 @@ package query
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/query/selection.go
```diff
@@ -157,8 +158,23 @@ func mergeCriteria(c1 []Criterion, c2 []Criterion) ([]Criterion, error) {
 	}
 
 	for _, newCriterion := range c2 {
+
+		// disallow duplicate label queries
+		if newCriterion.Type == LabelQuery {
```
dpanayotov: this should be by key IMO.
`key1 = val1|key1 = val2` => `key1=val1 or key1=val2` which is equal to `key1 in [val1,val2]`

KirilKabakchiev: Ok - initially i thought we could allow key1 = val1|key1 = val2 because it works but if you prefer we can disallow - I've added a second commit disallowing dup keys

KirilKabakchiev: on a second thought `key1 = val1|key1 = val2` should actually mean that key1 is equal to both val1 and val2 at the same time (since values are an array, it should be mean that values contains both val1 and val2).

What do you think?

PR #214, opened 2019-01-14
Enable fieldQuerying on all APIs, parameterize API tests

KirilKabakchiev:
- enables field querying on all APIs
- parameterizes list,get,delete and deletelist API tests and label and field querying tests
- some refactoring 
- fix a bug during label and field querying that disallows comparing json pg to string (::text casting)
- fix a bug during label and field querying that disallows arrays as right hand side arguments to `in` and `notin` operators ( `||` is used as separator between right operands of multi value operators now)
- allow all read ops to work with basic auth (consistency)
- expose Storage from SMBuilder to be accessible when SM is used as library


test/broker_test/broker_test.go
```diff
@@ -36,1310 +37,1114 @@ func TestBrokers(t *testing.T) {
...
+							JSON().Object().
+							Keys().Contains("error", "description")
+					})
+				})
 
-					ctx.SMWithOAuth.GET("/v1/service_offerings").
-						Expect().
-						Status(http.StatusOK).
-						JSON().Path("$.service_offerings[*].id").Array().NotContains(serviceOfferingID)
+				Context("when request body is not valid JSON", func() {
```
dpanayotov: isn't this a generic test case for all resources?

test/common/application.yml
```diff
@@ -3,8 +3,8 @@ server:
   shutdown_timeout: 4000ms
   port: 1234
 log:
-  level: debug
-  format: text
+  level: error
+  format: json
```
dpanayotov: should this be text?

test/common/common.go
```diff
@@ -39,6 +45,139 @@ import (
 type Object = map[string]interface{}
 type Array = []interface{}
 
+func RmNonNumericArgs(obj Object) Object {
```
dpanayotov: Rm -> Remove?

test/common/common.go
```diff
@@ -39,6 +45,139 @@ import (
...
+				}
+			}
+		} else if !isNumeric(v) {
+			delete(o, k)
+		}
+	}
+	return o
+}
+
+func RmNumericArgs(obj Object) Object {
```
dpanayotov: These two along with RmNonJson are the same with only 1 bool comparison

test/broker_test/broker_test.go
```diff
@@ -36,1310 +37,1114 @@ func TestBrokers(t *testing.T) {
...
-				assertInvocationCount(brokerServer.CatalogEndpointRequests, 1)
-			})
-		})
+var _ = test.DescribeTestsFor(test.TestCase{
+	API:            "/v1/service_brokers",
+	SupportsLabels: false,
+	SupportedOps: []test.Op{
+		test.Get, test.List, test.Delete, test.DeleteList,
+	},
+	RndResourceBlueprint:                      blueprint(true),
```
dpanayotov: rnd can be omitted IMO

test/list.go
```diff
@@ -0,0 +1,498 @@
...
+			Expect().
+			Status(http.StatusOK).JSON().Object()
+		result.ContainsKey("labels")
+		r := result.Raw()
+		return r
+	}
+
+	By(fmt.Sprintf("Attempting to create a random resource of %s with mandatory fields only", t.API))
+	rWithMandatoryFields = t.RndResourceWithoutNullableFieldsBlueprint(ctx)
+	for i := 0; i < 4; i++ {
```
dpanayotov: 4? does it have to be 4 or it could be anything?

test/platform_test/platform_test.go
```diff
@@ -20,363 +20,283 @@ import (
...
-
-	Describe("List", func() {
-		Context("With no platforms", func() {
-			It("returns empty array", func() {
-				ctx.SMWithOAuth.GET("/v1/platforms").
-					Expect().
-					Status(http.StatusOK).
-					JSON().Object().Value("platforms").Array().Empty()
+			Describe("POST", func() {
+				Context("With invalid content type", func() {
```
dpanayotov: this could be generic test also

pkg/types/broker.go
```diff
@@ -28,7 +28,7 @@ import (
 
 // Brokers struct
 type Brokers struct {
-	Brokers []*Broker `json:"brokers"`
```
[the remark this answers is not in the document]
KirilKabakchiev: Adressed

test/broker_test/broker_test.go
KirilKabakchiev: Post and patch can also be generic but they will come next week probably

test/common/application.yml
KirilKabakchiev: Will revert

test/list.go
KirilKabakchiev: Just a random number greater than 2 will suffice

test/platform_test/platform_test.go
KirilKabakchiev: Post and patch i didnt have time for so we could do them later

KirilKabakchiev:
coverage drops because tests for free plans filter are  disabled as it is temporary unregistered

c0d1ngm0nk3y:
Isn't this PR a *bit* large? Judging on the description, this PR fixes/changes a couple of things at once. Would it make sense to split this PR into several PRs?

PR #222, opened 2019-01-28
handle error from client response

KirilKabakchiev:
When handing client errors, we don't want to propagate the client error to the api consumer but rather log it and return a general error

### DIFF
--- a/pkg/util/errors.go
+++ b/pkg/util/errors.go
@@ -69,7 +69,11 @@ func HandleResponseError(response *http.Response) error {
 	}
 
 	err = fmt.Errorf("StatusCode: %d Body: %s", response.StatusCode, body)
-	log.C(response.Request.Context()).Errorf("Call to client failed with: %s", err)
+	if response.Request != nil {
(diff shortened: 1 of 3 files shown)


pkg/util/errors.go
```diff
@@ -63,22 +63,14 @@ func WriteError(err error, writer http.ResponseWriter) {
...
 		return fmt.Errorf("error processing response body of resp with status code %d: %s", response.StatusCode, err)
 	}
 
-	if err := BytesToObject(body, httpErr); err != nil || httpErr.Description == "" {
-		logger.Debugf("Failure response with status code %d is not an HTTPError. Error converting body: %v. Default err will be returned.", response.StatusCode, err)
-		return fmt.Errorf("StatusCode: %d Body: %s", response.StatusCode, body)
-	}
-	return httpErr
+	err = fmt.Errorf("StatusCode: %d Body: %s", response.StatusCode, body)
+	log.D().Errorf("Call to client failed with: %s", err)
```
dpanayotov: the `response.Request` should be populated if this is a client request so this can be logged with context

PR #223, opened 2019-01-28
Broker API response now returns "service_brokers" instead of "brokers"

KirilKabakchiev:
Broker API response now returns "service_brokers" instead of "brokers"

* Requires adoption in proxy-core
* Requires adoption in CLI

### DIFF
--- a/pkg/types/broker.go
+++ b/pkg/types/broker.go
@@ -28,7 +28,7 @@ import (
 
 // Brokers struct
 type Brokers struct {
-	Brokers []*Broker `json:"brokers"`
+	Brokers []*Broker `json:"service_brokers"`
 }
 
 // Broker broker struct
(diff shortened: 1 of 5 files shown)


dotchev:
Should we increase the api version to v2?

KirilKabakchiev:
Probably no need for v2 - we are in in the minor versions so backward incompatible changes are ok

NickyMateev:
Specification PR: https://github.com/Peripli/specification/pull/40

NickyMateev:
Service Broker Proxy core PR: https://github.com/Peripli/service-broker-proxy/pull/37

NickyMateev:
Service Manager CLI PR: https://github.com/Peripli/service-manager-cli/pull/35

PR #232, opened 2019-02-17
Omit platform credentials for Platform PATCH API response

NickyMateev:
# Pull Request Template

## Motivation

Currently when updating a platform the response body would contain credentials where the password would be messed up (because it's encrypted).

In reality, we have specified in the Service Manager specification that platform credentials will be returned only upon platform credentials. So we should not return then in any form for other APIs.


dotchev:
Looks ok. Don't we need a test to check it?

NickyMateev:
@dotchev Tests check that credentials are not returned on PATCH now.

PR #237, opened 2019-03-19
Add sending request with headers

KirilKabakchiev:
Currently send request util cannot be used for post/patch apis that require content-type header. A new util method is added to allow providing headers (this is also backward compatible as the go API of the old SendRequest method is not changed)

### DIFF
--- a/pkg/util/client.go
+++ b/pkg/util/client.go
@@ -32,6 +32,11 @@ type DoRequestFunc func(request *http.Request) (*http.Response, error)
 
 // SendRequest sends a request to the specified client and the provided URL with the specified parameters and body.
 func SendRequest(ctx context.Context, doRequest DoRequestFunc, method, url string, params map[string]string, body interface{}) (*http.Response, error) {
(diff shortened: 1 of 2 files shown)


pkg/util/client.go
```diff
@@ -32,6 +32,11 @@ type DoRequestFunc func(request *http.Request) (*http.Response, error)
 
 // SendRequest sends a request to the specified client and the provided URL with the specified parameters and body.
 func SendRequest(ctx context.Context, doRequest DoRequestFunc, method, url string, params map[string]string, body interface{}) (*http.Response, error) {
+	return SendRequestWithHeaders(ctx, doRequest, method, url, params, body, map[string]string{})
+}
+
+// SendRequestWithHeaders sends a request to the specified client and the provided URL with the specified parameters, body and headers.
+func SendRequestWithHeaders(ctx context.Context, doRequest DoRequestFunc, method, url string, params map[string]string, body interface{}, headers map[string]string) (*http.Response, error) {
```
dotchev: too many arguments, a struct would be better and also allow for optional arguments

PR #239, opened 2019-03-28
Race condition with new logrus version

dpanayotov:
Caused by:
https://github.com/onrik/logrus/pull/10 line 42
https://github.com/sirupsen/logrus/blob/master/entry.go#L245

The logger's mutex must be shared between all logrus.Entry instances, so that the hooks firing is synced. Fix is to pass the same logger to every entry.

### DIFF
--- a/pkg/log/context_test.go
+++ b/pkg/log/context_test.go
@@ -29,7 +29,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-//TestServiceManager tests servermanager package
+// TestServiceManager tests servermanager package
 func TestLog(t *testing.T) {
 	RegisterFailHandler(Fail)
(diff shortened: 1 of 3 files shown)


pkg/log/context_test.go
```diff
@@ -29,12 +29,28 @@ import (
...
+}
+
+// TestMultipleGoroutinesContextLog validates that no race conditions occur when two go routines log using the context log
+func TestMultipleGoroutinesContextLog(t *testing.T) {
+	ctx := Configure(context.TODO(), DefaultSettings())
+	log := func() { C(ctx).Debug("message") }
+	go log()
+	go log()
+}
+
```
dotchev: do we need a test to mix all of them, i.e. `C` & `D`? 
SM may call any of these in parallel

PR #242, opened 2019-04-09
Fix cf/env_test when STORAGE.URI is set

georgifarashev:
When STORAGE_URI env variable is set the test cf/env_test.go fails.
Solution is to unset the variable before the tests begin.


georgifarashev:
Fixes https://github.com/Peripli/service-manager/issues/243

cf/env_test.go
```diff
@@ -68,6 +68,7 @@ var _ = Describe("CF Env", func() {
 		Expect(os.Setenv("VCAP_APPLICATION", "{}")).ShouldNot(HaveOccurred())
 		Expect(os.Setenv("VCAP_SERVICES", VCAP_SERVICES_VALUE)).ShouldNot(HaveOccurred())
 		Expect(os.Setenv("STORAGE_NAME", "smdb")).ShouldNot(HaveOccurred())
+		Expect(os.Unsetenv("STORAGE_URI")).ShouldNot(HaveOccurred())
```
KirilKabakchiev: should be on line 96

PR #244, opened 2019-04-10
Notifications db model

georgifarashev:
Introduce notifications db model and a dummy controller


storage/postgres/notification.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Notification entity
+//go:generate smgen storage notification github.com/Peripli/service-manager/pkg/types:Notification
+type Notification struct {
+	BaseEntity
+	Resource     string             `db:"resource"`
+	Type         string             `db:"type"`
+	PlatformID   string             `db:"platform_id"`
+	Revision     int64              `db:"revision,auto-increment"`
+	New          sqlxtypes.JSONText `db:"new"`
```
KirilKabakchiev: why not top-level object that has old and new ?

storage/postgres/notification.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+// Notification entity
+//go:generate smgen storage notification github.com/Peripli/service-manager/pkg/types:Notification
+type Notification struct {
+	BaseEntity
+	Resource     string             `db:"resource"`
+	Type         string             `db:"type"`
+	PlatformID   string             `db:"platform_id"`
+	Revision     int64              `db:"revision,auto-increment"`
```
KirilKabakchiev: maybe auto_increment?

storage/postgres/migrations/009_notifications.up.sql
```diff
@@ -0,0 +1,51 @@
...
+CREATE TABLE notifications
+(
+  id            varchar(100) PRIMARY KEY,
+  resource      varchar(100) NOT NULL,
+  type          notification_type NOT NULL,
+  platform_id   varchar(100) REFERENCES platforms (id) ON DELETE CASCADE, -- value is platform_id from platforms table or null
+  revision      bigserial    NOT NULL,
+  created_at    timestamp    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at    timestamp    NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  new           json, -- json with the new value of the object and additional properties
```
KirilKabakchiev: wouldnt it be better to have object json column here and we can put new, old and label changes in it ? that way if we need to change the notification format in the future we won't have to change the db table

storage/postgres/abstract.go
```diff
@@ -206,7 +206,7 @@ func getTags(fields []*structs.Field, set *[]string) {
 			*set = append(*set, embedded...)
 		} else {
 			dbTag := field.Tag("db")
-			if dbTag == "-" {
+			if dbTag == "-" || strings.Contains(dbTag, "auto-increment") {
```
KirilKabakchiev: add a // comment with explanation what happens maybe as it is a "magic" string

PR #246, opened 2019-04-11
Release connection when working with rows

dpanayotov:
Currently, rows are not closed in multiple places.  
This results in connection exhaustion when decrypting credentials for more than 50 items at a time.

### Tests
Tested on a local dockerized Postgres with max connections = 100. 
110 platforms in the database. On `List` the password is decrypted:

## Before this fix
No call to list platforms passes, because for each platform, a new connection is opened to the database, but is not closed / released. So with any more than 50 platforms in the database, this call fails.

## With this fix

* 50 parallel requests. 
We get a spike of idle & total connections.
016FB74C-D0B5-4337-A8B0-ADD80AECB6BE

* 97 parallel requests. 
Sometimes  `sorry, too many clients already` which is expected, as the maximum number of connections is 100
978247BC-92AD-483E-8491-E3433F8CA960

Tried to perform the tests by starting a transaction to fetch the platforms & encryption key inside it. No difference, new connections are opened, but they are active. No new idle connections remain.
566E7CFD-C0E0-4B28-A0DE-DAB7B26838DF

We might decide not to fetch the encryption key on every encrypt/decrypt operation, but fetch it on startup, and keep the secondary encryption key in memory.

### DIFF
--- a/storage/postgres/security.go
+++ b/storage/postgres/security.go
@@ -92,6 +92,7 @@ type keyFetcher struct {
 func (s *keyFetcher) GetEncryptionKey(ctx context.Context) ([]byte, error) {
 	safe := &Safe{}
 	rows, err := listByFieldCriteria(ctx, s.db, "safe", []query.Criterion{})
+	defer closeRows(ctx, rows)
 	if err != nil {
 		return nil, err
 	}
@@ -115,6 +116,7 @@ type keySetter struct {
 // Sets the encryption key by encrypting it beforehand with the encryption key in the environment
 func (k *keySetter) SetEncryptionKey(ctx context.Context, key []byte) error {
 	rows, err := listByFieldCriteria(ctx, k.db, "safe", []query.Criterion{})
+	defer closeRows(ctx, rows)
 	if err != nil {
 		return err
 	}
--- a/storage/postgres/storage.go
+++ b/storage/postgres/storage.go
@@ -214,14 +214,7 @@ func (ps *PostgresStorage) Delete(ctx context.Context, objType types.ObjectType,
 		return nil, err
 	}
 	rows, err := deleteAllByFieldCriteria(ctx, ps.pgDB, entity.TableName(), entity, criteria)
-	defer func() {
-		if rows == nil {
-			return
-		}
-		if err := rows.Close(); err != nil {
-			log.C(ctx).Errorf("Could not release connection when checking database. Error: %s", err)
-		}
-	}()
+	defer closeRows(ctx, rows)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


storage/postgres/security.go
```diff
@@ -92,6 +92,7 @@ type keyFetcher struct {
 func (s *keyFetcher) GetEncryptionKey(ctx context.Context) ([]byte, error) {
```
pankrator: Can we pass here the pgDB object, so we reuse the already opened transaction connection? This way we won't even open another connection?

dpanayotov: The last screenshot shows just that. Instead of changing the interface, just for the tests I did:
```go
        var objectList types.ObjectList
	err := itr.InTransaction(ctx, func(ctx context.Context, txStorage Repository) error {
		var err error
		objectList, err = txStorage.List(ctx, objectType, criteria...)
		if err != nil {
			return err
		}
		encrypter := &security.TwoLayerEncrypter{Fetcher: txStorage.Security().Fetcher()}
			obj := objectList.ItemAt(i)
			if err = transformCredentials(ctx, obj, encrypter.Decrypt); err != nil {
				return err
			}
		}
		return nil
	})
	if err != nil {
		return nil, err
	}

	return objectList, nil
```
Because of the underlying implementation of storage.Security(), the pgDB object is actually `sqlx.Tx` object. https://github.com/Peripli/service-manager/blob/master/storage/postgres/storage.go#L293 and https://github.com/Peripli/service-manager/blob/master/storage/postgres/storage.go#L60

pankrator: Yes, that is want I meant by pgDB object, the `sqlx.Tx`

KirilKabakchiev: How many parallel requests could you do after the resource leak fix?

storage/postgres/storage.go
```diff
@@ -290,9 +284,10 @@ func (ps *PostgresStorage) InTransaction(ctx context.Context, f func(ctx context
 	}()
 
 	transactionalStorage := &PostgresStorage{
-		pgDB:   tx,
-		db:     ps.db,
-		scheme: ps.scheme,
+		pgDB:          tx,
+		db:            ps.db,
+		scheme:        ps.scheme,
+		encryptionKey: ps.encryptionKey,
```
NickyMateev: If we're not going to use this `encryptionKey` we should remove it

NickyMateev: If we're not going to use this `encryptionKey` we should remove it

dpanayotov: Opening a transaction, and calling `txStorage.Security().Fetcher()` inside a transaction currently fails, because the encryption key is not set.

PR #247, opened 2019-04-11
Install fixed version of counterfeiter

pankrator:
# In this PR:

* Makefile install fixed version of counterfeiter which will make life easier

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,8 @@ GO_INT_TEST 	= $(GO) test -p 1 -race -coverpkg $(shell go list ./... | egrep -v 
 GO_UNIT_TEST 	= $(GO) test -p 1 -race -coverpkg $(shell go list ./... | egrep -v "fakes|test|cmd" | paste -sd "," -) \
 				$(shell go list ./... | egrep -v "test") -coverprofile=$(UNIT_TEST_PROFILE)
 
+COUNTERFEITER   ?= "v6.0.2"
+
 #-----------------------------------------------------------------------------
 # Prepare environment to be able to run other make targets
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -74,8 +76,10 @@ ifeq ($(shell which goveralls),)
 	@go get github.com/mattn/goveralls
 endif
 ifeq ($(shell which counterfeiter),)
-	@echo "Installing counterfeither..."
+	@echo "Installing counterfeither $(COUNTERFEITER)..."
 	@go get github.com/maxbrunsfeld/counterfeiter
+	@cd ${GOPATH}/src/github.com/maxbrunsfeld/counterfeiter && git checkout tags/$(COUNTERFEITER) && go install
```
KirilKabakchiev: I think it would be better to put lines 81 and 82 in the generate target so that we make sure this version is used on each generation

Makefile
```diff
@@ -50,11 +50,21 @@ GO_INT_TEST 	= $(GO) test -p 1 -race -coverpkg $(shell go list ./... | egrep -v
...
 #-----------------------------------------------------------------------------
 # Prepare environment to be able to run other make targets
 #-----------------------------------------------------------------------------
 
-prepare: build-gen-binary ## Installs some tools (dep, gometalinter, cover, goveralls)
+prepare-counterfeiter:
+ifeq ($(shell which counterfeiter),)
+	@echo "Installing counterfeiter $(COUNTERFEITER)..."
+	@go get github.com/maxbrunsfeld/counterfeiter
+	@cd ${GOPATH}/src/github.com/maxbrunsfeld/counterfeiter && git checkout tags/$(COUNTERFEITER) && go install
```
KirilKabakchiev: checkout back to the branch it was before ?

PR #248, opened 2019-04-11
SM Notifications

georgifarashev:
Part of what Service Manager stores/owns is information about service brokers, platforms and visibilities (which broker plan should be visible in which platform). Service Manager takes care of propagating information about changes in the state of brokers and visibilities to the Service Manager proxies and the SM proxies apply that in the platforms. The communication is always from the proxy to Service Manager and currently the proxy fetches the brokers and visibilities from Service Manager it cares about regularly (every 30 seconds) and applies that in the platform. The problem is that this is not happening in real time - however, it is expected that once a client creates a visibility in Service Manager, the plan for the visibility should be visible in the specified platform in real time. (almost instantly)

For the purpose of real-time updates we plan to use websocket communication between SM and the proxies in order to allow dispatching of real-time notifications to the proxies that can then be applied on the platform. The notifications themselves are created when certain events occur (example create/update/delete broker, create/delete visibility). The notifications are stored in Service Manager storage and via pg NOTIFY are propagated to all SM nodes that have proxies connected to them. Then via the open websocket connection, the notifications are propagated to the proxies that care about the information. The notification should contain enough information so that when the proxy receives it, it does not need to callback Service Manager for additional details.


dpanayotov:
TODO:
Restart of SM causes proxy not to reconnect - adopt local dev for WS
CFProxy fails to start initially due to wrong CF credentials
Make proxy log connected on info
Check log level everywhere
Fix order of visibility notification created after broker is registered
add SM environment to local-dev

PR #252, opened 2019-04-19
Add notification queue

georgifarashev:
Notification queue which will be used by websocket controllers to receive notifications.

### DIFF
--- a/notifications/interfaces.go
+++ b/notifications/interfaces.go
@@ -36,10 +36,9 @@ type NotificationQueue interface {
 	// It should not block or execute heavy operations.
 	Enqueue(notification *types.Notification) error
 
-	// Next returns the next notification which has to be processed.
-	// If there are no new notifications the call will block.
+	// Channel returns the go channel with received notifications which has to be processed.
 	// If error is returned this means that the NotificationQueue is no longer valid.
-	Next() (*types.Notification, error)
+	Channel() (<-chan *types.Notification, error)
 
 	// Close closes the queue.
 	// Any subsequent calls to Next or Enqueue will return ErrQueueClosed.
--- a/notifications/interfaces.go
+++ b/notifications/interfaces.go
@@ -19,6 +19,7 @@ package notifications
 import (
 	"context"
 	"errors"
+	"sync"
 
 	"github.com/Peripli/service-manager/pkg/types"
 	"github.com/Peripli/service-manager/pkg/web"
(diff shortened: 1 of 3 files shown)


notifications/notification_queue.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/gofrs/uuid"
+
+	"github.com/Peripli/service-manager/pkg/types"
+)
+
+var _ NotificationQueue = &notificationQueue{}
+
+// NewNotificationQueue returns new NotificationQueue with specific size
+func NewNotificationQueue(size int) (NotificationQueue, error) {
```
KirilKabakchiev: Its better (according to the go guidelines) to return concrete type and not interface

notifications/interfaces.go
```diff
@@ -0,0 +1,64 @@
...
+
+// ErrQueueClosed error stating that the queue is closed
+var ErrQueueClosed = errors.New("queue closed")
+
+// ErrQueueFull error stating that the queue is full
+var ErrQueueFull = errors.New("queue is full")
+
+// NotificationQueue is used for receiving notifications
+type NotificationQueue interface {
+	// Enqueue adds a new notification for processing. If queue is full ErrQueueFull is returned.
```
dpanayotov: interface documentation should not be implementation-specific

notifications/interfaces.go
```diff
@@ -0,0 +1,64 @@
...
+	// It should not block or execute heavy operations.
+	Enqueue(notification *types.Notification) error
+
+	// Channel returns the go channel with received notifications which has to be processed.
+	// If error is returned this means that the NotificationQueue is no longer valid.
```
dpanayotov: Why is this error needed? `notification, ok <-queue.Channel()` will return `false` if the channel is closed

notifications/interfaces.go
```diff
@@ -0,0 +1,64 @@
...
+	Channel() (<-chan *types.Notification, error)
+
+	// Close closes the queue.
+	// Any subsequent calls to Next or Enqueue will return ErrQueueClosed.
+	// Any subsequent calls to Close does nothing.
+	Close()
```
dpanayotov: What does it mean to close a queue assuming that a queue might not be a channel?

notifications/notification_queue.go
```diff
@@ -0,0 +1,75 @@
...
+
+type notificationQueue struct {
+	isClosed             bool
+	size                 int
+	notificationsChannel chan *types.Notification
+	id                   string
+}
+
+func (nq *notificationQueue) Enqueue(notification *types.Notification) error {
+	if nq.isClosed {
```
dpanayotov: can this be called by multiple goroutines? mutex may be needed

notifications/interfaces.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// ID returns unique queue identifier
+	ID() string
+}
+
+// Notificator is used for receiving notifications for SM events
```
dpanayotov: not used in the PR

PR #256, opened 2019-05-03
Notifications controller

pankrator:
# In this PR

* Add Websocket server
* Add notifications controller
* Add component tests for notifications sending


api/notifications/notifications.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/Peripli/service-manager/pkg/ws"
+)
+
+const (
+	lastKnownRevisionHeader     = "last_known_revision"
+	lastKnownRevisionQueryParam = "last_known_revision"
+)
+
+func (c *Controller) handleWS(req *web.Request) (*web.Response, error) {
+	user, _ := web.UserFromContext(req.Context())
```
georgifarashev: It is best if we do not dismiss the bool here even though it is probably not possible to be false.

api/notifications/notifications.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not register notification consumer: %v", err)
+	}
+
+	revisionKnownToProxy := 0
+	revisionKnownToProxyStr := req.URL.Query().Get(lastKnownRevisionQueryParam)
+	if revisionKnownToProxyStr != "" {
+		revisionKnownToProxy, err = strconv.Atoi(revisionKnownToProxyStr)
+		if err != nil {
+			c.unregisterConsumer(notificationQueue)
```
georgifarashev: Is it possible to add these in defer somehow? It might be forgotten if this code is refactored in the future.

api/notifications/notifications.go
```diff
@@ -0,0 +1,202 @@
...
+			return nil, &util.HTTPError{
+				StatusCode:  http.StatusBadRequest,
+				Description: fmt.Sprintf("invalid %s query parameter", lastKnownRevisionQueryParam),
+				ErrorType:   "BadRequest",
+			}
+		}
+	}
+
+	if revisionKnownToProxy > 0 {
+		resp, err := c.isRevisionFound(req.Context(), revisionKnownToProxyStr)
```
georgifarashev: Notification cleaner might be executed between this call and the list operation.
Maybe get all notifications between `revisionKnownToProxy` and `lastKnownRevision` inclusive. Then sort them by revision (this will be workaround until there is storage orderby functionality) and check if the first notification has revision `revisionKnownToProxy`. If not - 410. It is also a good idea to check for revisionKnownToProxy == 0. In that case return directly 410.

api/notifications/notifications.go
```diff
@@ -0,0 +1,202 @@
...
+func (c *Controller) sendWsClose(conn *ws.Conn) {
+	// TODO: Timeout?
+	if err := conn.WriteControl(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseGoingAway, ""), time.Time{}); err != nil {
+		log.D().Errorf("Could not send close message: %v", err)
+	}
+}
+
+func (c *Controller) unregisterConsumer(q storage.NotificationQueue) {
+	if unregErr := c.notificator.UnregisterConsumer(q); unregErr != nil {
+		log.D().Errorf("Could not unregister notification consumer: %v", unregErr)
```
georgifarashev: While the context is done it will be ok to log with it. This way there will be trace-ability for the whole request even after it has finished.

api/notifications/notifications.go
pankrator: Not sure if this is possible, a test is better to catch this kind of behavior. I will try to add a test for this case

api/http_handler.go
```diff
@@ -61,12 +61,12 @@ func (h *HTTPHandler) serve(res http.ResponseWriter, req *http.Request) error {
 	}
 
 	response, err := h.Handler.Handle(request)
-	if err != nil {
-		return err
-	}
 	if request.IsResponseWriterHijacked() {
```
dotchev: `err` is lost, at least log it

storage/postgres/storage.go
```diff
@@ -148,6 +148,12 @@ func (ps *PostgresStorage) Ping() error {
 }
 
 func (ps *PostgresStorage) Create(ctx context.Context, obj types.Object) (string, error) {
+	ps.mutex.RLock()
```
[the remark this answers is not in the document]
pankrator: I think this can be fixed if we remove RLock from InTransaction, then we won't have double RLock in one goroutine

api/api.go
```diff
@@ -62,8 +65,8 @@ func (s *Settings) Validate() error {
 	return nil
 }
 
-// NewInterceptableTransactionalRepository returns the minimum set of REST APIs needed for the Service Manager
-func New(ctx context.Context, repository storage.Repository, settings *Settings, encrypter security.Encrypter) (*web.API, error) {
+// New returns the minimum set of REST APIs needed for the Service Manager
+func New(ctx context.Context, repository storage.Repository, settings *Settings, encrypter security.Encrypter, wsServer *ws.Server, notificator storage.Notificator) (*web.API, error) {
```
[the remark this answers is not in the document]
pankrator: If we introduce API builder?

storage/postgres/storage.go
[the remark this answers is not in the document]
georgifarashev: Maybe just don't set ps.db to null when closing.

api/notifications/notifications.go
```diff
@@ -0,0 +1,219 @@
...
+		return nil, -1, nil, fmt.Errorf("could not register notification consumer: %v", err)
+	}
+	defer func() {
+		if err != nil {
+			c.unregisterConsumer(ctx, notificationQueue)
+		}
+	}()
+
+	if lastKnownRevision < revisionKnownToProxy {
+		log.C(ctx).Infof("Notification revision known to proxy %d is greater than revision known to SM %d", revisionKnownToProxy, lastKnownRevision)
```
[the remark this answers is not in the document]
pankrator: It is expected behavior, so this is info

api/notifications/notifications.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (c *Controller) unregisterConsumer(ctx context.Context, q storage.NotificationQueue) {
+	if unregErr := c.notificator.UnregisterConsumer(q); unregErr != nil {
+		log.C(ctx).Errorf("Could not unregister notification consumer: %v", unregErr)
+	}
+}
+
+func (c *Controller) getNotificationList(ctx context.Context, user *web.UserContext, revisionKnownToProxy, lastKnownRevision int64) (*types.Notifications, error) {
+	// TODO: is this +1/-1 ok or we should add less than or equal operator
```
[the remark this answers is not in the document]
pankrator: My suggestion is to add this in another PR. In this case +1/-1 is ok, because revisions are sequential numbers

api/notifications/notifications.go
```diff
@@ -0,0 +1,219 @@
...
+
+	if notificationsList.Len() > 0 && revisionKnownToProxy > 0 {
+		// TODO: we expect that notificationsList is ordered by revision
+		notification := notificationsList.ItemAt(0).(*types.Notification)
+		if notification.Revision != revisionKnownToProxy {
+			log.C(ctx).Infof("Notification with revision %d known to proxy not found", revisionKnownToProxy)
+			return nil, -1, nil, errRevisionNotFound
+		}
+		notificationsList.Notifications = notificationsList.Notifications[1:]
+	}
```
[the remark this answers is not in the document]
pankrator: Can you clarify the comment

api/notifications/notification_controller.go
```diff
@@ -14,39 +14,43 @@
...
-					Description: "Not Implemented",
-					ErrorType:   "Not Implemented",
-				}
-			},
+			Handler: c.handleWS,
 		},
 	}
 }
 
 // TODO: create the actual websocket handling and disable CRUD and List operations
```
KirilKabakchiev: probably remove this todo?

PR #257, opened 2019-05-06
Notifications cleaner

georgifarashev:
This PR introduces notification cleaner which deletes old notifications from SM database

### DIFF
--- a/pkg/sm/sm.go
+++ b/pkg/sm/sm.go
@@ -146,7 +146,7 @@ func New(ctx context.Context, cancel context.CancelFunc, env env.Environment) *S
 
 	notificationCleaner := &storage.NotificationCleaner{
 		Storage:  interceptableRepository,
-		Settings: cfg.Storage,
+		Settings: *cfg.Storage,
 	}
 
 	smb := &ServiceManagerBuilder{
@@ -196,21 +196,15 @@ func (sm *ServiceManager) Run() {
 	log.C(sm.ctx).Info("Running Service Manager...")
 	wg := &sync.WaitGroup{}
 
-	start(sm.Notificator, "notificator", sm.ctx, wg)
-	start(sm.NotificationCleaner, "notification cleaner", sm.ctx, wg)
+	if err := sm.Notificator.Start(sm.ctx, wg); err != nil {
+		log.C(sm.ctx).WithError(err).Panicf("could not start Service Manager notificator")
+	}
+	if err := sm.NotificationCleaner.Start(sm.ctx, wg); err != nil {
+		log.C(sm.ctx).WithError(err).Panicf("could not start Service Manager notification cleaner")
+	}
 
 	sm.Server.Run(sm.ctx)
 	wg.Wait()
-}
-
-type startable interface {
-	Start(ctx context.Context, group *sync.WaitGroup) error
-}
-
-func start(component startable, componentName string, ctx context.Context, group *sync.WaitGroup) {
-	if err := component.Start(ctx, group); err != nil {
-		log.C(ctx).WithError(err).Panicf("could not start Service Manager %s", componentName)
-	}
 }
 
 func initializeSecureStorage(ctx context.Context, secureStorage storage.Security) error {
--- a/pkg/sm/sm.go
+++ b/pkg/sm/sm.go
@@ -51,17 +51,19 @@ import (
 type ServiceManagerBuilder struct {
 	*web.API
 
-	Storage     *storage.InterceptableTransactionalRepository
-	Notificator storage.Notificator
-	ctx         context.Context
-	cfg         *server.Settings
+	Storage             *storage.InterceptableTransactionalRepository
+	Notificator         storage.Notificator
+	NotificationCleaner *storage.NotificationCleaner
+	ctx                 context.Context
+	cfg                 *server.Settings
 }
 
 // ServiceManager  struct
 type ServiceManager struct {
-	ctx         context.Context
-	Server      *server.Server
-	Notificator storage.Notificator
+	ctx                 context.Context
+	Server              *server.Server
+	Notificator         storage.Notificator
+	NotificationCleaner *storage.NotificationCleaner
 }
 
 // DefaultEnv creates a default environment that can be used to boot up a Service Manager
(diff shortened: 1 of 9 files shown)


storage/interfaces.go
```diff
@@ -90,6 +90,8 @@ type NotificationSettings struct {
 	QueuesSize           int           `mapstructure:"queues_size"`
 	MinReconnectInterval time.Duration `mapstructure:"min_reconnect_interval"`
 	MaxReconnectInterval time.Duration `mapstructure:"max_reconnect_interval"`
+	CleanInterval        time.Duration `mapstructure:"clean_interval"`
```
dpanayotov: maybe these 2 need to be validated, because setting a negative `older_than will increase the timestamp in the clean functions and will clean the whole table?

storage/notification_cleaner.go
```diff
@@ -0,0 +1,75 @@
...
+			case <-time.After(cleanInterval):
+				nc.clean(ctx)
+			}
+		}
+	}()
+	return nil
+}
+
+func (nc *NotificationCleaner) clean(ctx context.Context) {
+	cleanTimestamp := time.Now().Add(-nc.Settings.Notification.OlderThan).Format(time.RFC3339)
```
dpanayotov: `nc.Settings.Notification.OlderThan` may be changed externally (unlikely, but still I may initialise it with a valid `olderThan` and later change it to a negative value, so this will truncate the whole table). Maybe introduce a constructor that needs only the two properties, or keep a copy of the settings and not a reference.

storage/notification_cleaner_test.go
```diff
@@ -0,0 +1,116 @@
...
+					objType = objectType
+					criteria = criterion
+					cancel() // stop notification cleaner
+					return nil, nil
+				}
+				err := nc.Start(ctx, wg)
+				Expect(err).ToNot(HaveOccurred())
+				wg.Wait()
+				Expect(objType).To(Equal(types.NotificationType))
+				Expect(criteria).To(HaveLen(1))
```
dpanayotov: maybe check that the left_op is `created_at` and the right op is less than `time.Now()` (this is to catch the negative clean period)

storage/notification_cleaner_test.go
```diff
@@ -0,0 +1,116 @@
...
+			Storage:  fakeStorage,
+			Settings: storage.DefaultSettings(),
+		}
+	})
+
+	AfterEach(func() {
+		if ctx.Err() == nil {
+			cancel()
+		}
+		wg.Wait()
```
dpanayotov: Not needed

storage/notification_cleaner_test.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+var _ = Describe("Notification cleaner", func() {
+	var (
+		ctx           context.Context
+		cancel        context.CancelFunc
+		wg            *sync.WaitGroup
+		fakeStorage   *storagefakes.FakeStorage
+		nc            *storage.NotificationCleaner
+		expectedError = errors.New("*Expected*")
```
dpanayotov: used only once. Do you need to validate something?

storage/notification_cleaner.go
```diff
@@ -0,0 +1,75 @@
...
+	log.C(ctx).Infof("Deleting notifications created before %s", cleanTimestamp)
+
+	q := query.ByField(query.LessThanOperator, "created_at", cleanTimestamp)
+	_, err := nc.Storage.Delete(ctx, types.NotificationType, q)
+	if err != nil && err != util.ErrNotFoundInStorage {
```
dpanayotov: if not notifications are found, the error will never be `ErrNotFoundInStorage` because you're using a repository that handles this error (https://github.com/Peripli/service-manager/blob/master/storage/interceptable_repository.go#L328) Maybe write a test for that one.

test/notification_cleaner_test/notification_cleaner_test.go
```diff
@@ -0,0 +1,139 @@
...
+			oldNotification := randomNotification()
+			oldNotification.CreatedAt = time.Now().Add(-defaultOlderThan - time.Hour)
+			idOld, err := repository.Create(ctx, oldNotification)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(idOld).ToNot(BeNil())
+
+			getNotification(idOld)
+			getNotification(idNew)
+
+			continueDelete <- struct{}{}
```
dpanayotov: Can't this be simplified by setting the `older_than` configuration to be for example `30m` and leave the `created_at` of the old notification to be 1 hour ago?

storage/notification_cleaner_test.go
georgifarashev: Not needed to be as variable.

pkg/sm/sm.go
```diff
@@ -142,12 +144,18 @@ func New(ctx context.Context, cancel context.CancelFunc, env env.Environment) *S
...
-		ctx:         ctx,
-		cfg:         cfg.Server,
-		API:         API,
-		Storage:     interceptableRepository,
-		Notificator: pgNotificator,
+		ctx:                 ctx,
+		cfg:                 cfg.Server,
+		API:                 API,
+		Storage:             interceptableRepository,
+		Notificator:         pgNotificator,
```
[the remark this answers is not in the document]
georgifarashev: Will be fixed with another PR

storage/notification_cleaner.go
```diff
@@ -0,0 +1,75 @@
...
+type NotificationCleaner struct {
+	started bool
+
+	Storage  Repository
+	Settings *Settings
+}
+
+// Start schedules the cleaner. It cannot be used concurrently.
+func (nc *NotificationCleaner) Start(ctx context.Context, group *sync.WaitGroup) error {
+	if nc.started {
```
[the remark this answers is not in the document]
georgifarashev: Start interface could return error so we can use the func declaration in a util that manages go routines with waitgroup.

storage/notification_cleaner_test.go
georgifarashev: This is for safety. Some tests in the future might require this.

storage/notification_cleaner.go
```diff
@@ -64,12 +65,25 @@ func (nc *NotificationCleaner) Start(ctx context.Context, group *sync.WaitGroup)
...
+	if isNotFoundError(err) {
+		log.C(ctx).Debug("no old notifications to delete")
+	} else if err != nil {
 		log.C(ctx).WithError(err).Error("could not delete old notifications")
+	} else {
+		log.C(ctx).Infof("successfully deleted %d old notifications", deletedNotifications.Len())
+	}
+}
+
+func isNotFoundError(err error) bool {
```
dpanayotov: PR #261 addresses the issue that handleStorageError is called from multiple places. I'd suggest merging master into the `notifications` branch and then reverting this here

georgifarashev: Will be fixed when merging master into notifications

PR #261, opened 2019-05-07
fix/refactor storage errors

KirilKabakchiev:
Storage error handling was happening all over the place (in interceptors, in controllers, etc).

Now the only place handling storage errors is happening is in the base controller.

### DIFF
--- a/pkg/util/errors.go
+++ b/pkg/util/errors.go
@@ -103,7 +103,7 @@ var (
 type ErrBadRequestStorage error
 
 // HandleStorageError handles storage errors by converting them to relevant HTTPErrors
-func HandleStorageError(err error, entityNames ...string) error {
+func HandleStorageError(err error, entityName string) error {
 	if err == nil {
 		return nil
 	}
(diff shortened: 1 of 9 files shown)


pkg/util/errors.go
```diff
@@ -92,20 +99,25 @@ var (
 	ErrAlreadyExistsInStorage = errors.New("unique constraint violation")
 )
 
+// ErrBadRequestStorage represents a storage error that should be translated to http.StatusBadRequest
 type ErrBadRequestStorage error
 
 // HandleStorageError handles storage errors by converting them to relevant HTTPErrors
-func HandleStorageError(err error, entityName string) error {
+func HandleStorageError(err error, entityNames ...string) error {
```
dpanayotov: always index 0 is used. `entityName` might be better and godoc should be adjusted to say that if entity name is provided the first one will be used

KirilKabakchiev: i Think its always used with arg now so i will remove the varargs

KirilKabakchiev:
double handling might happen  i suppose - we might get this situation with the resync_helpers that are used both in the entitlements controller and interceptors - but the util seems to work with double handling properly - other than that it seems ok

PR #271, opened 2019-05-12
apply label updates to labels before triggering interceptors

KirilKabakchiev:
Applying the label changes to the labels before handing them to the interceptor chains allows the interceptors to use the new objects labels and creating entitlements and visibilities. This fixes an issue in the internal update interceptors (sub-accounts, entitlements)

### DIFF
--- a/api/base_controller.go
+++ b/api/base_controller.go
@@ -218,13 +218,19 @@ func (c *BaseController) PatchObject(r *web.Request) (*web.Response, error) {
 	if err := util.BytesToObject(r.Body, objFromDB); err != nil {
 		return nil, err
 	}
+
 	objFromDB.SetID(objectID)
 	objFromDB.SetCreatedAt(createdAt)
 	objFromDB.SetUpdatedAt(updatedAt)
+
+	labels, _, _ := query.ApplyLabelChangesToLabels(labelChanges, objFromDB.GetLabels())
+	objFromDB.SetLabels(labels)
+
 	object, err := c.repository.Update(ctx, objFromDB, labelChanges...)
 	if err != nil {
 		return nil, util.HandleStorageError(err, string(c.objectType))
 	}
+
 	stripCredentials(ctx, object)
 
 	return util.NewJSONResponse(http.StatusOK, object)
--- a/pkg/query/update.go
+++ b/pkg/query/update.go
@@ -100,8 +100,6 @@ func ApplyLabelChangesToLabels(changes LabelChanges, labels types.Labels) (types
 	}
 
 	for _, change := range changes {
-		change := change
-
 		switch change.Operation {
 		case AddLabelOperation:
 			fallthrough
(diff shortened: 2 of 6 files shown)


dotchev:
Describe the issue that this change solves

api/base_controller.go
```diff
@@ -218,13 +218,19 @@ func (c *BaseController) PatchObject(r *web.Request) (*web.Response, error) {
 	if err := util.BytesToObject(r.Body, objFromDB); err != nil {
 		return nil, err
 	}
+
 	objFromDB.SetID(objectID)
 	objFromDB.SetCreatedAt(createdAt)
 	objFromDB.SetUpdatedAt(updatedAt)
+
+	labels, _, _ := query.ApplyLabelChangesToLabels(labelChanges, objFromDB.GetLabels())
+	objFromDB.SetLabels(labels)
```
dotchev: it seems this is done twice - here and in `Update`, why?

KirilKabakchiev: To handle the case when intrrceptors add label changes

storage/interceptable_repository.go
```diff
@@ -190,6 +190,9 @@ func (ir *interceptableRepository) Update(ctx context.Context, obj types.Object,
 			return nil, err
 		}
 
+		labels, _, _ := query.ApplyLabelChangesToLabels(labelChanges, newObj.GetLabels())
```
dpanayotov: this is strange. You update the object and receive an updated object as a result, but you still have to update the labels of the object (even though the labels in the db is updated)

KirilKabakchiev: coz interceptors can add label changes while they are executing

pkg/query/update.go
```diff
@@ -89,3 +91,52 @@ func LabelChangesFromJSON(jsonBytes []byte) ([]*LabelChange, error) {
 	}
 	return labelChanges, nil
 }
+
+// ApplyLabelChangesToLabels applies the specified label changes to the specified labels
+func ApplyLabelChangesToLabels(changes LabelChanges, labels types.Labels) (types.Labels, types.Labels, types.Labels) {
```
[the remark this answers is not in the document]
dotchev: actually only the first return value is used, the function can be simplified significantly

dotchev:
How an interceptor can get the labels before the update?

pkg/query/update.go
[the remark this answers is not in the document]
KirilKabakchiev: it will be used in the proxy as well

pkg/query/update.go
```diff
@@ -89,3 +91,52 @@ func LabelChangesFromJSON(jsonBytes []byte) ([]*LabelChange, error) {
...
+		mergedLabels[k] = v
+	}
+
+	for _, change := range changes {
+		change := change
+
+		switch change.Operation {
+		case AddLabelOperation:
+			fallthrough
+		case AddLabelValuesOperation:
```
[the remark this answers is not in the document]
KirilKabakchiev: fallthrough usage  is to be consistent with the way such handling is done in other places in the project

PR #276, opened 2019-05-15
fix race condition in storage

KirilKabakchiev:


storage/postgres/storage.go
```diff
@@ -112,9 +112,7 @@ func (ps *PostgresStorage) Close() error {
 	ps.checkOpen()
 	ps.mutex.Lock()
 	defer ps.mutex.Unlock()
-	defer func() {
-		ps.db = nil
-	}()
+
```
pankrator: With this change can you reopen the storage?

KirilKabakchiev: different SMs dont share the same storage object anymore so different test suites can "reopen" a storage (it would esencially be a new storage).

But within the same SM process you cant reopen a storage and you shouldnt be able to, imo - opening should be done only one time

storage/registry.go
```diff
@@ -21,59 +21,27 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/Peripli/service-manager/pkg/log"
-)
+	"github.com/Peripli/service-manager/pkg/util"
```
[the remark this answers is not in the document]
KirilKabakchiev: Where would you like to move it ?

storage/registry.go
```diff
@@ -21,59 +21,27 @@ import (
...
 )
 
-// Register adds a storage with the given name
-func Register(name string, storage Storage) {
-	mux.RLock()
-	defer mux.RUnlock()
-	logger := log.D()
-	if storage == nil {
-		logger.Panicln("storage: Register storage is nil")
+func OpenWithSafeTermination(ctx context.Context, s Storage, options *Settings, wg *sync.WaitGroup) error {
```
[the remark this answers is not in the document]
KirilKabakchiev: Like you said - it was not possible before either. I just removed unnecessary functions (Register, etc) - we can add issue to the fixit backlog

pkg/util/wg.go
```diff
@@ -0,0 +1,22 @@
...
+func StartInWaitGroup(f func(), group *sync.WaitGroup) {
+	group.Add(1)
+	go func() {
+		defer group.Done()
+		f()
+	}()
+}
+
+func StartInWaitGroupWithContext(ctx context.Context, f func(ctx context.Context), group *sync.WaitGroup) {
+	group.Add(1)
```
dpanayotov: Can be shortened to `StartInWaitGroup(func() { f(ctx) }, group)`

storage/postgres/notificator.go
```diff
@@ -94,7 +95,7 @@ func (n *Notificator) Start(ctx context.Context, group *sync.WaitGroup) error {
 	if err := n.openConnection(); err != nil {
 		return fmt.Errorf("could not open connection to database %v", err)
 	}
-	startInWaitGroup(n.awaitTermination, group)
+	util.StartInWaitGroup(n.awaitTermination, group)
```
dpanayotov: await termination can be inlined here

pkg/util/wg.go
KirilKabakchiev: In general we need to pass the context so the function f can later be invoked with it or do we expect to be able to use the context from closure ?

Issue #272, opened 2019-05-13
Do not return last_notification_revision 0 but skip this header


dotchev:
When a proxy connect to SM and there are no notifications in the db, SM returns header _last_notification_revision_ 0. This is an artificial notification revision as revisions start from 1.
If the websocket connection breaks before the first notification arrives, the proxy will try to reconnect with query parameter _last_notification_revision_ 0. Then SM will return immediately status 410 Gone. This is 
necessary because in the meantime some notifications might have passed and have been deleted from the db. Then the proxy will reconnect without _last_notification_revision_ and execute a full resync.

To avoid confusion, instead of _last_notification_revision_ 0, SM could skip this header in the response. Then if the proxy needs to reconnect before the first notification, it can connect without   _last_notification_revision_ query parameter.

dpanayotov:
Fixed with #289

PR #278, opened 2019-05-17
Notification filtering

georgifarashev:
# Notification filtering

## Motivation

Some notifications might not be necessary for some platforms.
For example if a broker must not be registered in a specific platform then a filter can be introduced on /v1/service_brokers which filters the broker. But the creation of the broker will create a notification for all platforms. A way of filtering this notification must be introduced.

## Approach

Introduce notification filters - functions which filter the recipients of a notification.

## Pull Request status

* [x] Implementation
* [x] Tests


api/notifications/notifications.go
```diff
@@ -24,22 +22,14 @@ const (
 	LastKnownRevisionHeader     = "last_notification_revision"
 	LastKnownRevisionQueryParam = "last_notification_revision"
 
-	unknownRevision int64 = -1
-	noRevision      int64 = 0
+	noRevision int64 = 0
```
dzahariev: Look like this constant is not needed. 
If the revision is Unknown the flow is the same. 

In the producer: https://github.com/Peripli/service-broker-proxy/blob/master/pkg/sbproxy/notifications/producer.go#L43
The value for InvalidRevision is returned when it is unknown.

storage/postgres/notificator.go
```diff
@@ -103,73 +109,133 @@ func (n *Notificator) Start(ctx context.Context, group *sync.WaitGroup) error {
...
+		return nil, err
+	}
+
+	missedNotifications, err := n.storage.ListNotifications(n.ctx, platform.ID, lastKnownRevision, lastKnownRevisionToSM)
 	if err != nil {
-		return nil, invalidRevisionNumber, err
+		return nil, err
+	}
+	filteredMissedNotification := make([]*types.Notification, 0, len(missedNotifications))
+	var recipients []*types.Platform
```
dpanayotov: remove this row

storage/postgres/notificator.go
```diff
@@ -103,73 +109,133 @@ func (n *Notificator) Start(ctx context.Context, group *sync.WaitGroup) error {
...
+	for _, notification := range missedNotifications {
+		recipients = n.filterRecipients([]*types.Platform{platform}, notification)
```
dpanayotov: this is very weird... it says "filter notifications" but filters platforms and after that this filtering doesn't use the returned result. The problem is that you're using the same function to do 2 different things.

storage/postgres/notificator.go
```diff
@@ -103,73 +109,133 @@ func (n *Notificator) Start(ctx context.Context, group *sync.WaitGroup) error {
...
+		return nil, types.INVALIDREVISION, err
 	}
-	queue, err := storage.NewNotificationQueue(n.queueSize)
+	return queueWithMissedNotifications, lastKnownRevisionToSM, nil
+}
+
+func (n *Notificator) filterRecipients(recipients []*types.Platform, notification *types.Notification) []*types.Platform {
+	for _, filter := range n.notificationFilters {
+		recipients = filter(recipients, notification)
+		if len(recipients) == 0 {
```
dpanayotov: filter returns an empty slice if any filter decides that no platform should receive it and the same list if all filters decide that at least one platform should receive it?

storage/postgres/notification_storage.go
```diff
@@ -69,3 +79,38 @@ func (ns *notificationStorageImpl) GetNotification(ctx context.Context, id strin
 	}
 	return notificationObj.(*types.Notification), nil
 }
+
+func (ns *notificationStorageImpl) ListNotifications(ctx context.Context, platformID string, from, to int64) ([]*types.Notification, error) {
+	// TODO: replace with less than or equal operator when available
```
dzahariev: Avoid TODOs

storage/postgres/notification_storage.go
```diff
@@ -69,3 +79,38 @@ func (ns *notificationStorageImpl) GetNotification(ctx context.Context, id strin
...
+	listQuery1 := query.ByField(query.GreaterThanOperator, "revision", strconv.FormatInt(from, 10))
+	listQuery2 := query.ByField(query.LessThanOperator, "revision", strconv.FormatInt(to+1, 10))
+
+	filterByPlatform := query.ByField(query.EqualsOrNilOperator, "platform_id", platformID)
+	objectList, err := ns.storage.List(ctx, types.NotificationType, listQuery1, listQuery2, filterByPlatform)
+	if err != nil {
+		return nil, err
+	}
+	notificationsList := objectList.(*types.Notifications)
+	// TODO: Should be done in the database with order by
```
dzahariev: The same - create a ticket/issues for the desired/missing functionalities and clean the TODOs.

storage/interfaces.go
```diff
@@ -240,9 +238,17 @@ type Notificator interface {
 	Start(ctx context.Context, group *sync.WaitGroup) error
 
 	// RegisterConsumer returns notification queue, last_known_revision and error if any.
+	// Notifications after lastKnownRevision will be added to the queue.
+	// If lastKnownRevision is -1 no previous notifications will be sent.
 	// When consumer wants to stop listening for notifications it must unregister the notification queue.
-	RegisterConsumer(userContext *web.UserContext) (NotificationQueue, int64, error)
+	RegisterConsumer(platform *types.Platform, lastKnownRevision int64) (NotificationQueue, int64, error)
```
dpanayotov: minor - platform should be consumer

storage/postgres/notification_storage.go
```diff
@@ -30,6 +36,10 @@ import (
 type notificationStorage interface {
```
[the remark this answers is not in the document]
georgifarashev: I agree. This interface will be removed when the storage supports sorting by a specific property and limit of returned results (in sql - order by and limit).

storage/postgres/notificator.go
```diff
@@ -103,73 +109,133 @@ func (n *Notificator) Start(ctx context.Context, group *sync.WaitGroup) error {
...
+	var recipients []*types.Platform
+	for _, notification := range missedNotifications {
+		recipients = n.filterRecipients([]*types.Platform{platform}, notification)
+		if len(recipients) != 0 {
+			filteredMissedNotification = append(filteredMissedNotification, notification)
+		}
+	}
+
+	if n.queueSize < len(filteredMissedNotification) {
+		log.C(n.ctx).Debugf("too many missed notifications %d", len(filteredMissedNotification))
```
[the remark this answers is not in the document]
georgifarashev: Yes, but if it can't queue them (queues have limited size) the proxy will do a full resync.

storage/postgres/notification_storage.go
georgifarashev: They were from a previous PR. I will remove them.

api/notifications/notifications.go
georgifarashev: SM sends 0 as last notification when there are no notifications in the database. This 0 is received by the proxy and if its connection drops it should resend it as last known revision. In this case SM sends 410 (see line 47 and 79). Idea to improve this is not to send the 0 at all (both by the SM and proxy). But this requires changes in the proxy as well. It should be done in a different PR.

api/notifications/notifications.go
```diff
@@ -58,9 +48,18 @@ func (c *Controller) handleWS(req *web.Request) (*web.Response, error) {
 		return util.NewJSONResponse(http.StatusGone, nil)
 	}
 
-	notificationQueue, lastKnownRevision, notificationsList, err := c.registerConsumer(ctx, revisionKnownToProxy, user)
+	user, ok := web.UserFromContext(req.Context())
+	if !ok {
```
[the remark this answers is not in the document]
georgifarashev: It is not used above. I have moved it to the place where it is needed for a first time.

storage/postgres/notificator.go
georgifarashev: Hopefully with the rename of `NotificationFilterFunc` to `ReceiversFilterFunc` it might be more clear. The idea is to filter given recipients for a specific notification by returning the filtered recipients. This way the one who is writing the filter can process the notification once for all recipients. `filterRecipients` function iterates over all registered filter functions and applies them to the initial recipients and given notification.

storage/postgres/notificator.go
georgifarashev: This is an optimisation - if a filter in the chain of filters returns 0 recipients then it is not necessary to call the other filters as there are no platforms to filter. See also comment above.

api/notifications/notifications.go
georgifarashev: https://github.com/Peripli/service-manager/issues/272

pkg/util/errors.go
```diff
@@ -100,6 +100,9 @@ var (
 
 	// ErrConcurrentResourceModification error returned when concurrent resource updates are happening
 	ErrConcurrentResourceModification = errors.New("another resource update happened concurrently. Please reattempt the update")
+
+	// ErrInvalidNotificationRevision provided notification revision is not valid, must return http status GONE
+	ErrInvalidNotificationRevision = errors.New("notification revision is not valid")
```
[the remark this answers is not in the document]
georgifarashev: I have created an issue for that - https://github.com/Peripli/service-manager/issues/281 as it could be done in a different PR.

pkg/types/notification.go
```diff
@@ -35,6 +35,9 @@ const (
 
 	// DELETED represents a notification type for deleting a resource
 	DELETED OperationType = "DELETED"
+
+	// INVALIDREVISION revision with invalid value
+	INVALIDREVISION int64 = -1
```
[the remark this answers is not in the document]
georgifarashev: It is used in the storage/postgres and api/notifications. It will be strange for any of them to import the other. It cannot be in storage as well as postgres must not import storage. This seemed the most appropriate as InvalidRevision is some value of a property of notification similar to the operation types (CREATED, DELETED ...).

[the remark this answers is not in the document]
dpanayotov: Can it be `InvalidNotificationRevision` then, because `types.InvalidRevision` doesn't mean much

PR #280, opened 2019-05-22
Expose notification creation

dotchev:
Also add support for websocket connections in the test context.
We need these features in a wrapper of SM.


test/common/test_context.go
```diff
@@ -416,4 +425,45 @@ func (ctx *TestContext) CleanupAdditionalResources() {
...
+
+	wsEndpoint := smEndpoint.String()
+	conn, resp, err := websocket.DefaultDialer.Dial(wsEndpoint, headers)
+	if conn != nil {
+		ctx.wsConnections = append(ctx.wsConnections, conn)
+	}
+	return conn, resp, err
+}
+
+func (ctx *TestContext) CloseWebSocket(conn *websocket.Conn) {
```
dpanayotov: this is not used

dotchev: It will be used in tests in sm-sap

PR #282, opened 2019-05-23
[Fixit] Storage and Encryption

KirilKabakchiev:
# Pull Request Template

## PR Size
PR contains 900 lines of fakes and 500 lines of tests so effective PR size is about +750/-XX

## Motivation

Certain improvements in the storage layer and in the encryption/decryption flows:
* Encapsulate encryption/decryption in storage layer as separate "sub-layer" - separate "repository decorator" called encrypting repository wraps the actual repository in sm.go and takes care of encryption/decryption
* Storage takes care of encrypting and decrypting, storage always returns decrypted credentials if obj is Secured so that interceptors can later on (if required) use these credentials
* Remove KeyFetcher, KeySetter
* Second Layer Encryption key is fetched just once and reused afterwards instead of being refetched on each op
* remove Security() methods from storage; embedd Security as Secured to be storage native
* Create now returns the whole created object instead of just id
* Remove Registry interface from health and make it a struct

## Pull Request status

* [x] Initial implementation
* [x] Refactoring
* [x] Unit tests
* [x] Integration tests


storage/interfaces.go
```diff
@@ -195,20 +193,25 @@ type Credentials interface {
...
+	GetEncryptionKey(ctx context.Context, transformationFunc func(context.Context, []byte, []byte) ([]byte, error)) ([]byte, error)
 
-	// Setter provides means to change the encryption  key
-	Setter() security.KeySetter
+	SetEncryptionKey(ctx context.Context, key []byte, transformationFunc func(context.Context, []byte, []byte) ([]byte, error)) error
+}
+
+// SecuredTransactionalRepository is a transactional repository that allows securing data by encrypting it with an encryption key
+//go:generate counterfeiter . securedTransactionalRepository
+type securedTransactionalRepository interface {
```
dpanayotov: this should be exported probably

pkg/sm/sm.go
```diff
@@ -110,23 +110,31 @@ func New(ctx context.Context, cancel context.CancelFunc, env env.Environment) *S
...
-	encrypter := &security.TwoLayerEncrypter{
-		Fetcher: securityStorage.Fetcher(),
+	// Decorate the storage with credentials ecryption/decryption
+	encryptingRepository, err := storage.NewEncryptingRepository(ctx, smStorage, encrypter)
+	if err != nil {
+		panic(fmt.Sprintf("error setting up encrypting repository: %s", err))
 	}
 
+	// Decorate the storage with transactional interceptors
+	interceptableRepository := storage.NewInterceptableTransactionalRepository(encryptingRepository)
```
dpanayotov: I'd prefer all of this decoration to be hidden behind a configuration

pkg/sm/sm.go
```diff
@@ -110,23 +110,31 @@ func New(ctx context.Context, cancel context.CancelFunc, env env.Environment) *S
...
-		panic(fmt.Sprintf("error initialzing secure storage: %v", err))
+	// Initialize the storage with graceful termination
+	if err := storage.InitializeWithSafeTermination(ctx, smStorage, cfg.Storage, waitGroup); err != nil {
+		panic(fmt.Sprintf("error opening storage: %s", err))
 	}
 
-	encrypter := &security.TwoLayerEncrypter{
-		Fetcher: securityStorage.Fetcher(),
+	// Decorate the storage with credentials ecryption/decryption
+	encryptingRepository, err := storage.NewEncryptingRepository(ctx, smStorage, encrypter)
```
dpanayotov: What if I create a `NewEcnryptingRepository` with uninitialized storage?

KirilKabakchiev: it will panic during setting up of the encryption key

pkg/sm/sm.go
KirilKabakchiev: wouldnt it be too hidden then ? Is your idea to pass a config to SM builder that contains decorators or do you mean something else?

storage/encrypting_repository.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/Peripli/service-manager/pkg/log"
+
+	"github.com/Peripli/service-manager/pkg/security"
+
+	"github.com/Peripli/service-manager/pkg/query"
+	"github.com/Peripli/service-manager/pkg/types"
+)
+
+func NewEncryptingRepository(ctx context.Context, repository securedTransactionalRepository, encrypter security.Encrypter) (*EncryptingRepository, error) {
```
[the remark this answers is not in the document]
KirilKabakchiev: This interface is needed so that we can pass a repository that is secured and to not cast to Secured. We can export, or we can remove this interface and pass two arguments (Secured and TransactionalRepository) but we will pass the same struct smStroage for both). Or we can pass just TransactionaRepository and cast.

storage/interceptable_repository.go
```diff
@@ -99,33 +88,27 @@ func (ir *interceptableRepository) Create(ctx context.Context, obj types.Object)
...
-		err = createObjectFunc(ctx, ir.repositoryInTransaction, obj)
+		createdObj, err = createObjectFunc(ctx, ir.repositoryInTransaction, obj)
 	}
 
 	if err != nil {
-		return "", err
-	}
-
-	if securedObj, isSecured := obj.(types.Secured); isSecured {
-		securedObj.SetCredentials(nil)
```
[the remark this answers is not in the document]
KirilKabakchiev: yeah on create I would prefer to always return credentials - seems more consistent

pkg/sm/sm.go
dpanayotov: That's my concern. You need to know to do something hidden before constructing an object. That's what my proposal for hiding all of this behind a configuration. Like with flags what to enable - "I want a repository with interceptors and with encryption" would return a ready to use `storage.InterceptableTransactionalRepository`

pkg/sm/sm.go
dpanayotov: That would be an option `smb.EnableStorageInterceptors().EnableStorageEncryption()`. Or just the storage configuration can be extended with options to allow interceptors and encryption. 
```
storage:
    uri: ....
    interceptors: true
    encryption: true
```

pkg/sm/sm.go
KirilKabakchiev: I agree that requiring the hidden Opening before creating the object should be adjusted but I am a bit concerned about allowing switching encryption on and off - imagine SM is started in ecrypt: off, some data is stored, then its restarted with encrypt: on  and some data is store again ... and stuff will start failing  as now we have both encrypted and unecrypted data

PR #285, opened 2019-05-23
[Fixit] Authentication and Credentials

KirilKabakchiev:
# Pull Request Template

## Motivation

* Improve authentication filter code / packages structure
* Remove Credentials interface and Data field from types.Credentials
* Basic auth filter uses repository native query (list with username field query) instead of specialized one that was coming from Credentials interface
## Pull Request status

* [x] Initial implementation
* [x] Refactoring
* [x] Unit tests

# TODO

Change base to master before merging


api/filters/basic_authentication.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Authenticate authenticates by using the provided Basic credentials
+func (a *basicAuthenticator) Authenticate(request *http.Request) (*web.UserContext, httpsec.Decision, error) {
+	username, password, ok := request.BasicAuth()
+	if !ok {
+		return nil, httpsec.Abstain, nil
+	}
+
+	ctx := request.Context()
+	byUsername := query.ByField(query.EqualsOperator, "username", username)
```
dpanayotov: the sbproxy has in-memory implementation of credentials storage that would require a dummy implementation of the whole repository

KirilKabakchiev: not necessarily - i refactored the authenticatiom so that the pkg basic auth filter can be used with a custom inmemory authenticator

KirilKabakchiev: proxy will call this method with custom inmemory authenticator  
```
func NewAuthenticationFilter(authenticator http.Authenticator, name string, matchers []web.FilterMatcher) *AuthenticationFilter {
	return &AuthenticationFilter{
		Authentication: &middlewares.Authentication{
			Authenticator: authenticator,
		},
		matchers: matchers,
		name:     name,
	}
}
```

KirilKabakchiev: https://github.com/Peripli/service-broker-proxy/pull/77

PR #286, opened 2019-05-27
[Fixit] Allow patching labels of services and plans

DimitarPetrov:
## Motivation

Currently there is no way to add/remove labels on service offerings and service plans.

## Approach

Introduces a patch API to service offerings and service plans apis that allows patching only the labels.

### DIFF
--- a/api/base_controller.go
+++ b/api/base_controller.go
@@ -204,25 +204,27 @@ func (c *BaseController) PatchObject(r *web.Request) (*web.Response, error) {
 	if err != nil {
 		return nil, err
 	}
-	if r.Body, err = sjson.DeleteBytes(r.Body, "labels"); err != nil {
-		return nil, err
-	}
 
 	objFromDB, err := c.repository.Get(ctx, c.objectType, objectID)
 	if err != nil {
 		return nil, util.HandleStorageError(err, string(c.objectType))
 	}
 
-	createdAt := objFromDB.GetCreatedAt()
-	updatedAt := objFromDB.GetUpdatedAt()
+	if c.objectType != types.ServiceOfferingType && c.objectType != types.ServicePlanType {
+		if r.Body, err = sjson.DeleteBytes(r.Body, "labels"); err != nil {
+			return nil, err
+		}
+		createdAt := objFromDB.GetCreatedAt()
+		updatedAt := objFromDB.GetUpdatedAt()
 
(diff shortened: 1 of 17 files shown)


test/service_offering_test/service_offering_test.go
```diff
@@ -36,7 +36,7 @@ func TestServiceOfferings(t *testing.T) {
 
 var _ = test.DescribeTestsFor(test.TestCase{
 	API:            "/v1/service_offerings",
-	SupportsLabels: false,
+	SupportsLabels: true,
```
dpanayotov: As everything now supports labels, this field and the checks can be removed completely

api/base_controller.go
```diff
@@ -236,6 +236,33 @@ func (c *BaseController) PatchObject(r *web.Request) (*web.Response, error) {
 	return util.NewJSONResponse(http.StatusOK, object)
 }
 
+// PatchObject handles the update of labels of the object with the id specified in the request
+func (c *BaseController) PatchLabelsOnly(r *web.Request) (*web.Response, error) {
```
dpanayotov: this is a lot of code duplication. Can you see how to extract this part?

api/base_controller.go
```diff
@@ -236,6 +236,33 @@ func (c *BaseController) PatchObject(r *web.Request) (*web.Response, error) {
...
+		return nil, util.HandleStorageError(err, string(c.objectType))
+	}
+
+	labels, _, _ := query.ApplyLabelChangesToLabels(labelChanges, objFromDB.GetLabels())
+	objFromDB.SetLabels(labels)
+
+	object, err := c.repository.Update(ctx, objFromDB, labelChanges...)
+	if err != nil {
+		return nil, util.HandleStorageError(err, string(c.objectType))
+	}
```
dpanayotov: credentials stripping is missing

api/filters/patch_labels_only.go
```diff
@@ -0,0 +1,68 @@
...
+		{
+			Matchers: []web.Matcher{
+				web.Path(web.ServicePlansURL + "/**"),
+				web.Methods(http.MethodPatch),
+			},
+		},
+	}
+}
+
+func onlyLabelsRequest(body []byte) bool {
```
KirilKabakchiev: you could inline this function - its short enough and is not used anywhere else

PR #287, opened 2019-05-27
[Fixit] Change parser to ANTLR4

dpanayotov:
# What is this PR about
This PR changes the parsing logic from our own implementation to one provided by ANTLR4.
This way, introducing a new operator is pretty straightforward.
See https://github.com/Peripli/specification/blob/sm-ops/api.md#query-examples for examples 
# Breaking changes:
1. EqualsOperator - previously `=` is now `eq`
2. NotEqualsOperator - previously `!=` is now `ne`
3. Every string literal in RHS value of a query MUST be enclosed in single quotes
4. The query separator is changed from `|` to `and`
5. The literal separator in an array RHS value is changed from `||` to `,`
6. Multiple values in RHS - previously in square braces `[..]`, now in parentheses `(..)`
7. Changes the format of all dates from `RFC3339` to `RFC3339Nano` 
8. All string literals must be enclosed in single quotes `'`

So if you previously had a query like:
`/v1/visibilities?fieldQuery=platform_id+=+tcb-platform-test`
You now have to change operator from `=` to `eq` and quote the parameter like so:
`/v1/visibilities?fieldQuery=platform_id+eq+'tcb-platform-test'`

# New 
* You can now pass numbers with floating point and a sign as a RHS. If you want to get entities with `some_field` greater than `-1.5` you can do:
`/v1/visibilities?fieldQuery=some_field+gt+-1.5`
* You can now query for boolean fields **without** quoting the value:
`/v1/service_plans?fieldQuery=free+eq+true`
* You can now query by date time:
`/v1/service_brokers?fieldQuery=created_at+eq+2019-09-20T12:59:36.034829Z` This is particularly useful if you want to know which entities were created/updated before/after a certain time

# Size
pkg/query/parser directory is automatically generated and it has 1452 lines, so the size of this PR would be around 500 rows of new code


KirilKabakchiev:
What is the performance difference compared to the old parser that was implemented im the SM code ?

dpanayotov:
@KirilKabakchiev I would say there's no real difference
Time on my machine for 50 runs of https://github.com/Peripli/service-manager/blob/master/test/service_plan_test/service_plan_test.go#L53 
Master branch: Average 2.211569ms
This branch: Average 2.139315ms

pkg/query/listener.go
```diff
@@ -113,6 +117,23 @@ func getCriterionFields(key, op, right antlr.TerminalNode) (leftOp string, opera
 	return
 }
 
+func criterionDateToUTC(c *Criterion) error {
```
KirilKabakchiev: why is this needed?

dpanayotov: For this test - https://github.com/Peripli/service-manager/pull/287/commits/5f850b569d699e5e07660c1f3e29eaca775dff47#diff-a88508bea7d1a156ea3b6cd0ca291d40R401 
Where the query has offset from UTC but in the db we store in utc

pkg/query/Query.g4
```diff
@@ -0,0 +1,24 @@
...
+DATETIME: FOUR_DIGITS '-' TWO_DIGITS '-' TWO_DIGITS 'T' TWO_DIGITS ':' TWO_DIGITS ':' TWO_DIGITS '.' FIVE_DIGITS ('Z'|(SIGN TWO_DIGITS ':' TWO_DIGITS)) ;//2016-06-08T17:41:22Z
+DIGIT: INTEGER+ ;
+INTEGER: [0-9] ;
+TWO_DIGITS: DIGIT DIGIT ;
+FOUR_DIGITS: TWO_DIGITS TWO_DIGITS ;
+FIVE_DIGITS: FOUR_DIGITS DIGIT ;
+Key: [-_/a-zA-Z0-9\\]+ ;
+ValueSeparator: ',' | ', ' ;
+Whitespace: ' ' ;
+WS : [ \t\r\n]+ -> skip ; // skip spaces, tabs, newlines
```
KirilKabakchiev: I adjusted most of the comments in the gammar. Only thing left is the date - there are two options - either something more readable such as https://github.com/antlr/grammars-v4/blame/76873507fe61c9246831d7e903cb64d19075ece4/toml/toml.g4#L95 or the regex in the comment above

```
grammar Query ;

expression: criterions EOF ;
criterions: criterion (Concat criterions)? ;
criterion: multivariate | univariate  ;
multivariate: Key  MultiOp  multiValues ;
univariate: Key UniOp Value ;
multiValues: OpenBracket manyValues? CloseBracket ;
manyValues: Value (ValueSeparator manyValues)? ;

MultiOp: Whitespace 'in' Whitespace | Whitespace 'notin' Whitespace;
UniOp: Whitespace 'eq' Whitespace| Whitespace 'ne' Whitespace | Whitespace 'gt' Whitespace | Whitespace 'lt' Whitespace | Whitespace 'ge' Whitespace | Whitespace 'le' Whitespace | Whitespace 'en' Whitespace;
Concat: Whitespace 'and' Whitespace ;
Value: STRING | NUMBER | BOOLEAN | DATETIME ;
ValueSeparator: ',' | ', ' ;
Key: [-_/a-zA-Z0-9\\]+ ;
OpenBracket: '(' | '[' ;
CloseBracket: ')' | ']' ;

fragment BOOLEAN: 'true' | 'false' ;
fragment STRING: '\'' ('\\'. | '\'\'' | ~('\'' | '\\'))* '\'' ;
fragment DATETIME: FOUR_DIGITS '-' TWO_DIGITS '-' TWO_DIGITS 'T' TWO_DIGITS ':' TWO_DIGITS ':' TWO_DIGITS '.' FIVE_DIGITS ('Z'|(SIGN TWO_DIGITS ':' TWO_DIGITS)) ;//2016-06-08T17:41:22Z
fragment FIVE_DIGITS: FOUR_DIGITS DIGIT ;
fragment FOUR_DIGITS: TWO_DIGITS TWO_DIGITS ;
fragment TWO_DIGITS: DIGIT DIGIT ;
fragment NUMBER : SIGN? DIGIT ('.' DIGIT)?;
fragment SIGN: '-' | '+' ;
fragment DIGIT: INTEGER+ ;
fragment INTEGER: [0-9] ;
fragment Whitespace: ' ' ;

WS: [ \t\r\n]+ -> skip ; // skip spaces, tabs, newlines
```

pkg/query/listener.go
```diff
@@ -0,0 +1,144 @@
...
+	if c.LeftOp != "created_at" && c.LeftOp != "updated_at" {
+		return nil
+	}
+	rightOp := make([]string, 0, len(c.RightOp))
+	for _, r := range c.RightOp {
+		parsedTime, err := time.Parse(time.RFC3339Nano, r)
+		if err != nil {
+			return err
+		}
+		formatted := parsedTime.UTC().Format(time.RFC3339Nano)
```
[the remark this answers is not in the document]
dpanayotov: This is for testing currently. I haven't decided how to address the format change, because it's also a breaking change in the code - maybe introduce a `FormatRFCNano` method and deprecate the old `FormatRFC`

pkg/query/Query.g4
dpanayotov: Seems strange for the operator to include the whitespace in its definition. Would require processing of whitespaces in the parser

api/filters/selection_criteria.go
```diff
@@ -40,11 +44,24 @@ func (*SelectionCriteria) Name() string {
 // Run represents the selection criteria middleware function that processes the request and configures the request-scoped selection criteria.
 func (l *SelectionCriteria) Run(req *web.Request, next web.Handler) (*web.Response, error) {
 	ctx := req.Context()
-	criteria, err := query.BuildCriteriaFromRequest(req.Request)
-	if err != nil {
-		return nil, err
+	var criteria []query.Criterion
+	for _, queryType := range query.CriteriaTypes {
+		queryValue := req.URL.Query().Get(string(queryType))
+		unescaped, err := url.QueryUnescape(queryValue)
```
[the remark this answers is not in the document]
dpanayotov: not covered - https://github.com/Peripli/service-manager/blob/antlr_parser/test/list.go#L442

test/list.go
```diff
@@ -394,6 +397,61 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase) bool {
 			})
 		})
 
+		Context("by date", func() {
```
DimitarPetrov: Paging adaptation would be needed on GET calls

test/public_plans_interceptor_test/public_plans_test.go
```diff
@@ -294,6 +296,16 @@ var _ = Describe("Service Manager Public Plans Interceptor", func() {
 		})
 	})
 
+	Context("when fetching with query parameter", func() {
```
DimitarPetrov: Also must be adapted with paging

pkg/query/operators.go
```diff
@@ -0,0 +1,220 @@
...
+	// LessThanOperator takes two operands and tests if the left is lesser than the right
+	LessThanOperator ltOperator = "lt"
+	// LessThanOrEqualOperator takes two operands and tests if the left is lesser than or equal the right
+	LessThanOrEqualOperator leOperator = "le"
+	// InOperator takes two operands and tests if the left is contained in the right
+	InOperator inOperator = "in"
+	// NotInOperator takes two operands and tests if the left is not contained in the right
+	NotInOperator notInOperator = "notin"
+	// EqualsOrNilOperator takes two operands and tests if the left is equal to the right, or if the left is nil
+	EqualsOrNilOperator enOperator = "en"
```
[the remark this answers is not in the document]
dpanayotov: we don't have a use-case for this currently. we may leave it for later

pkg/query/selection.go
```diff
@@ -185,13 +169,8 @@ func (c Criterion) Validate() error {
...
 	}
-
-	if strings.ContainsRune(c.LeftOp, Separator) {
-		parts := strings.FieldsFunc(c.LeftOp, func(r rune) bool {
-			return r == Separator
-		})
-		possibleKey := parts[len(parts)-1]
-		return &util.UnsupportedQueryError{Message: fmt.Sprintf("separator %c is not allowed in %s with left operand \"%s\". Maybe you meant \"%s\"? Make sure if the separator is present in any right operand, that it is escaped with a backslash (\\)", Separator, c.Type, c.LeftOp, possibleKey)}
+	if strings.Contains(c.LeftOp, Separator) {
+		return &util.UnsupportedQueryError{Message: fmt.Sprintf("separator %s is not allowed in %s with left operand \"%s\".", Separator, c.Type, c.LeftOp)}
```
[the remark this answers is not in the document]
dpanayotov: this could be useful if you're providing multiple field and label queries at the same time - it would be easier to find in which query the problem is

storage/interceptable_repository.go
```diff
@@ -187,13 +187,17 @@ func (ir *interceptableRepository) Update(ctx context.Context, obj types.Object,
...
+	updateInProgress, _ := ctx.Value("op_in_progress").(bool)
+	if !oldObj.GetUpdatedAt().UTC().Equal(obj.GetUpdatedAt().UTC()) && !updateInProgress {
 		return nil, util.ErrConcurrentResourceModification
 	}
 
 	if updateInterceptorChain, found := ir.updateInterceptor[objectType]; found {
 		delete(ir.updateInterceptor, objectType)
 
+		// specify that an update is in progress so that multiple updates to the entity are not taken as false positive concurrent update
+		ctx = context.WithValue(ctx, "op_in_progress", true)
```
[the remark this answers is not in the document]
dpanayotov: as they are specifically meant to be used here I'd rather not expose them

test/list.go
```diff
@@ -391,6 +395,58 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase) bool {
...
+			})
+
+			It("and is not encoded, returns 400", func() {
+				ctx.SMWithOAuth.GET(t.API).WithQuery("labelQuery", fmt.Sprintf("%s eq '%s'", labelKey, labelValue)).
+					Expect().Status(http.StatusBadRequest).Body().Contains("not URL encoded")
+			})
+
+			It("and is encoded, returns 200", func() {
+				escapedLabelValue := url.QueryEscape(labelValue)
+				ctx.SMWithOAuth.GET(t.API).WithQuery("labelQuery", fmt.Sprintf("%s eq '%s'", labelKey, escapedLabelValue)).
```
DimitarPetrov: Maybe paging adaptation leftover. (ListWithQuery)

PR #291, opened 2019-05-30
Split child node strings in environment set keys with underscores

NickyMateev:
# Split child nodes in environment set with underscore

## Motivation

Currently when a new property is added to a configuration struct and in the cases it doesn't have `mapstructure` tag, the name used for generation of the set of environment variables will use the lowercased field name of the property.

This is okay when the new property is a single word, for example let's look at the following config:

```go
type Settings struct {
	Version    string 
	ClientID          string `mapstructure:"client_id" description:"id of the client from which the token must be issued"`
	SkipSSLValidation bool   `mapstructure:"skip_ssl_validation" description:"whether to skip ssl verification when making calls to external services"`
	TokenBasicAuth    bool   `mapstructure:"token_basic_auth" description:"specifies if client credentials to the authorization server should be sent in the header as basic auth (true) or in the body (false)"`
}

```

This would translate to the following environment set keys:
```
api.version
api.client_id
api.skip_ssl_validation
api.token_basic_auth
```

When the property name consists of multiple words things look a little different:

```go
type Settings struct {
	TokenIssuerURL    string 
	ClientID          string `mapstructure:"client_id" description:"id of the client from which the token must be issued"`
	SkipSSLValidation bool   `mapstructure:"skip_ssl_validation" description:"whether to skip ssl verification when making calls to external services"`
	TokenBasicAuth    bool   `mapstructure:"token_basic_auth" description:"specifies if client credentials to the authorization server should be sent in the header as basic auth (true) or in the body (false)"`
}
```

Now the result would be the following environment set keys:
```
api.tokenissuerurl
api.client_id
api.skip_ssl_validation
api.token_basic_auth
```

For the sake of consistency and with the intention to not create unexpected bugs this PR aims to fix this and introduce a change which will ensure that all child properties which are not tagged with `mapstructure` will have their names not only lowercased, but also split by underscores.

So now the above scenario would look like so:
```
api.token_issuer_url
api.client_id
api.skip_ssl_validation
api.token_basic_auth
```

## Approach

Before lowercasing the env key, the last child element is fetched and using a utility function the child element string has underscores inserted into it.

## Pull Request status

* [x] Initial implementation

### DIFF
--- a/pkg/env/helpers.go
+++ b/pkg/env/helpers.go
@@ -19,6 +19,7 @@ package env
 import (
 	"reflect"
 	"strings"
+	"unicode"
 
 	"github.com/fatih/structs"
 )
@@ -94,7 +95,11 @@ func buildDescriptionTreeWithParameters(value interface{}, tree *descriptionTree
 		if index == -1 {
 			index = 0
 		}
-		key := strings.ToLower(buffer[0:index])
+		split := strings.Split(buffer[0:index], ".")
+		lastChild := split[len(split)-1]
+		split[len(split)-1] = underscoreSplit(lastChild)
+
+		key := strings.ToLower(strings.Join(split, "."))
 		*result = append(*result, configurationParameter{Name: key, DefaultValue: value})
 		tree.Children = nil
 		return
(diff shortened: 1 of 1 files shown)


NickyMateev:
Not necessary, right now - closing this.

PR #292, opened 2019-05-30
Construct SMBuilder with config instead of environment

NickyMateev:
# Construct SMBuilder with configuration instead of environment

## Motivation

Service Manager custom implementations most likely would require information regarding different base SM configuration properties. In this way, custom SM implementations can wrap the default SM configuration easier, only provide the default one to the SM Builder and not have the default configuration "encapsulated" inside the constructor of the SM Builder.

## Approach

The `sm.New()` function now takes as final parameter a config instead of environment.

This PR also removes panicing inside some functions and delegating the error down to the `main()` functions.

## Pull Request status

* [x] Initial implementation

### DIFF
--- a/pkg/sm/sm.go
+++ b/pkg/sm/sm.go
@@ -43,10 +43,7 @@ import (
 	"github.com/Peripli/service-manager/storage/postgres"
 
 	"github.com/Peripli/service-manager/api/filters"
-	"github.com/Peripli/service-manager/cf"
-	"github.com/Peripli/service-manager/pkg/env"
 	"github.com/Peripli/service-manager/pkg/web"
-	"github.com/spf13/pflag"
 )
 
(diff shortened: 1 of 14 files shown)


pkg/sm/sm.go
```diff
@@ -79,23 +79,18 @@ func DefaultEnv(additionalPFlags ...func(set *pflag.FlagSet)) env.Environment {
...
 	}
 	if err := cf.SetCFOverrides(environment); err != nil {
-		panic(fmt.Errorf("error setting CF environment values: %s", err))
+		return nil, fmt.Errorf("error setting CF environment values: %s", err)
 	}
-	return environment
+	return environment, nil
 }
 
 // New returns service-manager Server with default setup. The function panics on bad configuration
```
DimitarPetrov: Maybe change the comment since function does not panic anymore

PR #294, opened 2019-05-31
Store and return raw catalog when SM acts as a Broker

KirilKabakchiev:
* remove pmorie/osb-client and implement our own logic to calls brokers' get catalog endpoint - the reason is that we'd like to support the upcoming OSB generic catalog extensions
* improve storage wiring in sm.go so that creating an ecrypting repository does not secretly depend on logic that happens when storage is initialized (Storage.Open)
* add more tests (broker_tests.go and osb_tests.go)
* sm now returns 502 instead of 400 when registering or updating a broker that cannot be reached during catalog fetching


pkg/sm/sm.go
```diff
@@ -98,40 +99,38 @@ func New(ctx context.Context, cancel context.CancelFunc, env env.Environment) *S
...
-	waitGroup := &sync.WaitGroup{}
-
 	smStorage := &postgres.Storage{
 		ConnectFunc: func(driver string, url string) (*sql.DB, error) {
 			return sql.Open(driver, url)
 		},
 	}
 
+	// Decorate the storage with credentials encryption/decryption
+	encryptingDecorator := storage.EncryptingDecorator(ctx, &security.AESEncrypter{}, smStorage)
```
KirilKabakchiev: it no longer matters if this is before or after storage initialization (it does not depend on storage.Open)

storage/storagefakes/fake_key_store.go
```diff
@@ -0,0 +1,346 @@
+// Code generated by counterfeiter. DO NOT EDIT.
```
NickyMateev: This isn't used anywhere... do we need it?

storage/postgres/keystore.go
```diff
@@ -33,6 +33,9 @@ type Safe struct {
 	UpdatedAt time.Time `db:"updated_at"`
 }
 
+type KeyStore struct {
```
NickyMateev: Leftover?

storage/catalog/fetcher_test.go
```diff
@@ -0,0 +1,172 @@
...
+		return catalog.Fetcher(common.DoHTTP(t.reaction, t.expectations), version)
+	}
+
+	basicAuth := func(username, password string) string {
+		auth := username + ":" + password
+		return base64.StdEncoding.EncodeToString([]byte(auth))
+	}
+
+	BeforeEach(func() {
+		testBroker = &types.ServiceBroker{
```
NickyMateev: There's no need to construct the broker in a `BeforeEach` every single time

storage/catalog/fetcher_test.go
```diff
@@ -0,0 +1,172 @@
...
+				},
+			},
+		}
+	})
+
+	entries := []TableEntry{
+		Entry("successfully fetches the catalog bytes", testCase{
+			expectations: &common.HTTPExpectations{
+				URL: url,
+				Headers: map[string]string{
```
NickyMateev: Extract the headers as they are the same for every test case

test/broker_test/broker_test.go
```diff
@@ -1214,14 +1339,28 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
 					Context("when an existing service plan is modified", func() {
 						Context("when catalog service plan id is modified but the catalog name is not", func() {
+							var expectedCatalog string
+
 							BeforeEach(func() {
-								sbCatalog := brokerServer.Catalog
+								expectedCatalog = string(brokerServer.Catalog)
 
-								catalog, err := sjson.Set(string(sbCatalog), "services.0.plans.0.id", "new-id")
+								catalog, err := sjson.Set(string(brokerServer.Catalog), "services.0.plans.0.id", "new-id")
```
NickyMateev: Why not use `expectedCatalog` here directly? (in this instance and in some places above)

test/broker_test/broker_test.go
```diff
@@ -1087,44 +1189,55 @@ var _ = test.DescribeTestsFor(test.TestCase{
 									Status(http.StatusConflict).JSON().Object().Keys().Contains("error", "description")
 
 								assertInvocationCount(brokerServer.CatalogEndpointRequests, 1)
+							})
+
+							It("is returned from the repository in the brokers catalog field", func() {
+								ctx.SMWithOAuth.PATCH("/v1/service_brokers/" + brokerID).
```
NickyMateev: This code is pretty similar and could probably be extracted into something like this:


```go
func verifyRepositoryReturnsBrokerCatalogField(brokerID, expectedCatalog string, expectedStatus int) {
        ctx.SMWithOAuth.PATCH("/v1/service_brokers/" + brokerID).
	    WithJSON(common.Object{}).
	    Expect().
	    Status(expectedStatus)

 	brokerFromDB, err := repository.Get(context.TODO(), types.ServiceBrokerType, brokerID)
	Expect(err).ToNot(HaveOccurred())

 	Expect(string(brokerFromDB.(*types.ServiceBroker).Catalog)).To(MatchJSON(expectedCatalog))
}
```

storage/interfaces.go
```diff
@@ -172,6 +172,9 @@ type TransactionalRepository interface {
 	InTransaction(ctx context.Context, f func(ctx context.Context, storage Repository) error) error
 }
 
+// RepositoryDecorator allows decorating a TransactionalRepository
+type RepositoryDecorator func(TransactionalRepository) (TransactionalRepository, error)
```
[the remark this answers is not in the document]
KirilKabakchiev: then it depends on storage.Open to have been called secretly (you would only be able to do NewEncryptingRepository after storage has been initialized (initializesStorageWithSafeTErmination)

storage/registry.go
```diff
@@ -40,8 +40,18 @@ func InitializeWithSafeTermination(ctx context.Context, s Storage, options *Sett
 	}, wg)
 
 	if err := s.Open(options); err != nil {
-		return fmt.Errorf("error opening storage: %s", err)
+		return nil, fmt.Errorf("error opening storage: %s", err)
 	}
 
-	return nil
+	var decoratedRepository TransactionalRepository
```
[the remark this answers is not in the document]
KirilKabakchiev: idea of the decorators was to allow hiding the wrapping of strorages/ repositories so that creating the NewEncryptingRepository does not depend on the storage being open

pkg/sm/sm.go
```diff
@@ -98,40 +99,38 @@ func New(ctx context.Context, cancel context.CancelFunc, env env.Environment) *S
 		panic(fmt.Sprintf("error validating configuration: %s", err))
 	}
 
+	// Setup the default http client
 	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: cfg.API.SkipSSLValidation}
+	http.DefaultClient.Transport = http.DefaultTransport
```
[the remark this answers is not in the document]
KirilKabakchiev: just following the initial idea of using default transport and adding it to the default http client and then using the default http client when creating the catalog fetcher (it accepts a http function)

storage/catalog/fetcher_test.go
KirilKabakchiev: There is also no way in go to declare it as a constant - this is the only way to guarantee it will be in the "expected" state before each test runs - otherwise someone might add more tests later on that modify it and this would lead to tests not testing what should be tested

pkg/types/service_offering.go
```diff
@@ -52,16 +52,34 @@ func (e *ServiceOffering) Validate() error {
...
-		return errors.New("service offering catalog name missing")
+		return fmt.Errorf("service offering catalog name missing")
 	}
 	if e.BrokerID == "" {
-		return errors.New("service offering broker id missing")
+		return fmt.Errorf("service offering broker id missing")
+	}
+	var array []interface{}
+	if len(e.Tags) != 0 {
+		if err := json.Unmarshal(e.Tags, &array); err != nil {
```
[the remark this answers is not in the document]
KirilKabakchiev: This wont work - the test below fails which points out that the validation for metadata with value "{invalid" passed successfully and it was successfully inserted in the database when using .Valid

```
v1/service_brokers
/Users/i321655/go/src/github.com/Peripli/service-manager/test/test.go:48
  additional non-generic tests
  /Users/i321655/go/src/github.com/Peripli/service-manager/test/broker_test/broker_test.go:57
    PATCH
    /Users/i321655/go/src/github.com/Peripli/service-manager/test/broker_test/broker_test.go:449
      when the broker catalog is modified
      /Users/i321655/go/src/github.com/Peripli/service-manager/test/broker_test/broker_test.go:825
        when an existing service plan is modified
        /Users/i321655/go/src/github.com/Peripli/service-manager/test/broker_test/broker_test.go:1356
          when metadata is invalid json
          /Users/i321655/go/src/github.com/Peripli/service-manager/test/broker_test/broker_test.go:1414
            the catalog is correctly returned by the repository [It]
            /Users/i321655/go/src/github.com/Peripli/service-manager/test/broker_test/broker_test.go:1035
```

storage/catalog/fetcher_test.go
NickyMateev: Clever, I understand

PR #295, opened 2019-06-03
[Fixit] Introduce SQL query builder

pankrator:
## Motivation

Currently building queries is hard to read and reason about. Introducing query builder will cleanup the sql building.
This change adds order by and limit to the storage layer.


storage/postgres/query_builder.go
```diff
@@ -0,0 +1,330 @@
...
+
+func (qb *QueryBuilder) List(ctx context.Context, entity PostgresEntity) (*sqlx.Rows, error) {
+	if qb.err != nil {
+		return nil, qb.err
+	}
+
+	baseTableName := entity.TableName()
+	var baseQuery string
+	label := entity.LabelEntity()
+	if label == nil {
```
dpanayotov: can this be moved to `constructBaseQueryForLabelable` where `if label == nil { return the code from constructBaseQueryForEntity}` and delete `constructBaseQueryForEntity`

storage/postgres/query_builder.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func NewQueryBuilder(db pgDB) *QueryBuilder {
+	return &QueryBuilder{
+		db: db,
+	}
+}
+
+func (qb *QueryBuilder) List(ctx context.Context, entity PostgresEntity) (*sqlx.Rows, error) {
+	if qb.err != nil {
```
dpanayotov: write some godoc somewhere that these methods are not thread-safe and the QueryBuilder cannot be reused.
I'd rather have queryBuilder in postgresql storage and reuse it instead of create new one every time.
The chaining can be extracted in another structure created once-per-build and `queryBuilder.NewQuery().WithCriteria()....` instead of `queryBuilder.WithCriteria()...`

storage/postgres/query_builder.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func (qb *QueryBuilder) Return(fields ...string) *QueryBuilder {
+	qb.returningFields = append(qb.returningFields, fields...)
+
+	return qb
+}
+
+func (qb *QueryBuilder) Limit(limit int) *QueryBuilder {
+	if limit <= 0 {
```
dpanayotov: this check should be in `Criterion.Validate`

storage/postgres/query_builder.go
```diff
@@ -0,0 +1,330 @@
...
+	return qb.db.QueryxContext(ctx, qb.sql, qb.queryParams...)
+}
+
+func (qb *QueryBuilder) Return(fields ...string) *QueryBuilder {
+	qb.returningFields = append(qb.returningFields, fields...)
+
+	return qb
+}
+
+func (qb *QueryBuilder) Limit(limit int) *QueryBuilder {
```
dpanayotov: These 2 should be private. You should configure these only via `WithCriteria`

storage/postgres/query_builder.go
```diff
@@ -0,0 +1,330 @@
...
+
+type orderRule struct {
+	field     string
+	orderType query.OrderType
+}
+
+type QueryBuilder struct {
+	db pgDB
+
+	sql         string
```
dpanayotov: maybe make it `strings.Builder` for performance

storage/postgres/query_builder.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	return validateFields(columns, fields...)
+}
+
+func validateReturningFields(columns map[string]bool, returningFields ...string) error {
+	if len(returningFields) > 0 {
+		if returningFields[0] == "*" {
+			return nil
+		}
+		return validateFields(columns, returningFields...)
```
dpanayotov: the error will not be meaningful - enrich it with context about the 'returning` part

storage/postgres/query_builder.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+	return ""
+}
+
+func validateOrderFields(columns map[string]bool, orderRules ...orderRule) error {
+	fields := make([]string, 0, len(orderRules))
+	for _, or := range orderRules {
+		fields = append(fields, or.field)
+	}
+	return validateFields(columns, fields...)
```
dpanayotov: enrich error with context that this is about the `order by` part

storage/postgres/query_builder.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func (qb *QueryBuilder) WithLock() *QueryBuilder {
+	if _, ok := qb.db.(*sqlx.Tx); ok {
+		qb.hasLock = true
+	}
+	return qb
+}
+
+func (qb *QueryBuilder) finalizeSQL(entity PostgresEntity) error {
```
dpanayotov: `buildQueryForEntity` which would return `(string, error)` if you make qb reusable

pkg/query/selection.go
```diff
@@ -49,6 +49,9 @@ const (
 	NotInOperator Operator = "notin"
 	// EqualsOrNilOperator takes two operands and tests if the left is equal to the right, or if the left is nil
 	EqualsOrNilOperator Operator = "eqornil"
+
+	// NoOperator signifies that this is not an operator
+	NoOperator Operator = "noop"
```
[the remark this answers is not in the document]
dpanayotov: should this operator be in the global operators list?

pkg/query/selection.go
```diff
@@ -88,6 +91,14 @@ const (
 	FieldQuery CriterionType = "fieldQuery"
 	// LabelQuery denotes that the query should be executed on the entity's labels
 	LabelQuery CriterionType = "labelQuery"
+	// ResultQuery is used to further process result
+	ResultQuery CriterionType = "resultQuery"
```
dpanayotov: should this be in the `supportedQueryTypes`?

pkg/query/selection.go
```diff
@@ -88,6 +91,14 @@ const (
...
+)
+
+const (
+	OrderBy string = "orderBy"
+
+	Limit string = "limit"
```
dpanayotov: add appropriate validation for these and the resultQuery in the validate method

storage/postgres/storage.go
```diff
@@ -229,7 +235,9 @@ func (ps *Storage) Delete(ctx context.Context, objType types.ObjectType, criteri
 	if err != nil {
 		return nil, err
 	}
-	rows, err := deleteAllByFieldCriteria(ctx, ps.pgDB, entity.TableName(), entity, criteria)
+
+	qBuilder := NewQueryBuilder(ps.pgDB)
+	rows, err := qBuilder.WithCriteria(criteria...).Return("*").Delete(ctx, entity)
```
dpanayotov: This order is strange. It seems more natural to say `Delete(ctx, entity).WithCriteria().Return()`

pkg/query/selection.go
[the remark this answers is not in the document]
pankrator: Currently no, because it will not be parsed from the URL's query

storage/postgres/query_builder.go
pankrator: It is nice to have these, as they allow for easier query building

pkg/query/selection.go
pankrator: Currently no, because it will not be parsed from the URL's query

storage/postgres/storage.go
pankrator: If it is implemented the way you propose, it will need to have one more method call to signify execution of query. For example: `qBuilder.Delete().WithCriteria().Return().Result()`

pkg/query/selection.go
```diff
@@ -114,6 +125,21 @@ func ByLabel(operator Operator, leftOp string, rightOp ...string) Criterion {
...
+const (
+	AscOrder  OrderType = "ASC"
+	DescOrder OrderType = "DESC"
+)
+
+func WithOrder(field string, orderType OrderType) Criterion {
+	return newCriterion(OrderBy, NoOperator, []string{field, string(orderType)}, ResultQuery)
+}
+
+func WithLimit(limit string) Criterion {
```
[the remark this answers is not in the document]
pankrator: I agree here, but then we have to do it in several places also.
What do you think about changing rightOp to interface{}?

pkg/query/selection.go
```diff
@@ -114,6 +125,21 @@ func ByLabel(operator Operator, leftOp string, rightOp ...string) Criterion {
...
 }
 
+type OrderType string
+
+const (
+	AscOrder  OrderType = "ASC"
+	DescOrder OrderType = "DESC"
+)
+
+func WithOrder(field string, orderType OrderType) Criterion {
```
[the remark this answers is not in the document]
pankrator: It becomes too long in my opinion if you have Limit and Order by you have to say twice `ByResult()`.
What  do you think about `query.OrderResultBy('field', AscOrder')` and `query.LimitResultBy(2)`?

storage/postgres/query_builder.go
pankrator: Removed them

PR #296, opened 2019-06-05
Allow pflag creation for slices

NickyMateev:
# Allow pflag creation for slices

## Motivation

There could be situations where slices are part of SM's configuration and corresponding pflags should be able to be created based on that.

## Approach

Remove the restriction for slices.

## Pull Request status

For example:

* [x] Initial implementation

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -77,7 +77,12 @@ func CreatePFlags(set *pflag.FlagSet, value interface{}) {
 
 	for i, parameter := range parameters {
 		if set.Lookup(parameter.Name) == nil {
-			set.Var(&flag{value: parameter.DefaultValue}, parameter.Name, descriptions[i])
+			switch val := parameter.DefaultValue.(type) {
+			case []string:
+				set.StringSlice(parameter.Name, val, descriptions[i])
+			default:
+				set.Var(&flag{value: val}, parameter.Name, descriptions[i])
+			}
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


dpanayotov:
Can you add some tests in env_test that use slices

pkg/env/env.go
```diff
@@ -77,7 +77,12 @@ func CreatePFlags(set *pflag.FlagSet, value interface{}) {
 
 	for i, parameter := range parameters {
 		if set.Lookup(parameter.Name) == nil {
-			set.Var(&flag{value: parameter.DefaultValue}, parameter.Name, descriptions[i])
+			switch val := parameter.DefaultValue.(type) {
+			case []string:
```
dpanayotov: remove this and add the following to the `flag.Type` method:
```golang
if kind == reflect.Slice {
	return fmt.Sprintf("%sSlice", typeOfValue.Elem().Name())
}
```

pkg/env/helpers.go
```diff
@@ -126,5 +126,5 @@ func buildDescriptionTreeWithParameters(value interface{}, tree *descriptionTree
```
dpanayotov: maybe add 
```golang
if name == "-" || name == ",squash" {
	continue
}
```
on line 112

pkg/env/env.go
NickyMateev: This opens the possibility of calling .Set() on the flag, which won't reflect the way a stringSlice flag would behave, could cause issues in the future. What do you think?

PR #297, opened 2019-06-10
Extend supported content types for APIs

NickyMateev:
# Extend supported content types for APIs

## Pull Request status

* [x] Initial implementation
* [x] Unit tests

### DIFF
--- a/pkg/util/api.go
+++ b/pkg/util/api.go
@@ -31,6 +31,7 @@ var (
 	reservedSymbolsRFC3986 = strings.Join([]string{
 		":", "/", "?", "#", "[", "]", "@", "!", "$", "&", "'", "(", ")", "*", "+", ",", ";", "=",
 	}, "")
+	supportedContentTypes = []string{"application/json", "application/x-www-form-urlencoded"}
 )
 
 // InputValidator should be implemented by types that need input validation check. For a reference refer to pkg/types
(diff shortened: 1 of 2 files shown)


pkg/util/api.go
```diff
@@ -31,6 +31,7 @@ var (
 	reservedSymbolsRFC3986 = strings.Join([]string{
 		":", "/", "?", "#", "[", "]", "@", "!", "$", "&", "'", "(", ")", "*", "+", ",", ";", "=",
 	}, "")
+	supportedContentTypes = []string{"application/json", "application/x-www-form-urlencoded"}
```
pankrator: Maybe map of supported types strings and boolean? I think it will be easier to use it. Just a thought

pkg/util/api.go
```diff
@@ -49,10 +50,18 @@ func ToRFCFormat(timestamp time.Time) string {
...
+	for _, supportedType := range supportedContentTypes {
+		if strings.Contains(contentType, supportedType) {
+			contentTypeSupported = true
+			break
+		}
+	}
+
+	if !contentTypeSupported {
 		return nil, &HTTPError{
 			ErrorType:   "InvalidMediaType",
```
pankrator: UnsupportedMediaType?

pkg/util/api.go
NickyMateev: That's not a bad idea

NickyMateev: Actually, in this context, this won't be very pragmatic, as the actual content type header will contain additional information in the string as well, ex:

`application/json; charset=utf-8`

So we can't define a map beforehand in which we'll later search with specific request content-type header values. That's why the current implementation uses `strings.Contains` and has to traverse the whole array.

pkg/util/api.go
```diff
@@ -49,13 +50,21 @@ func ToRFCFormat(timestamp time.Time) string {
 }
 
 // RequestBodyToBytes reads the request body and returns []byte with its content or an error if
-// the media type is incorrect or if the body is not a valid JSON
+// the media type is unsupported or if the body is not a valid JSON
 func RequestBodyToBytes(request *http.Request) ([]byte, error) {
 	contentType := request.Header.Get("Content-Type")
-	if !strings.Contains(contentType, "application/json") {
+	contentTypeSupported := false
```
dpanayotov: use `slice.StringsContaining(supportedContentTypes, contentType)`

NickyMateev: `slice.StringsContaining(supportedContentTypes, contentType)` actually works in reverse to what we need here. It will check that the provided request `contentType` is contained in any of the `supportedContentTypes`, but like I explained in the above comment, this will not work as the provided `contentType` will contain much more information than what we've defined in our `supportedContentTypes`. So `strings.Contains` will never match.

PR #298, opened 2019-06-11
Filter for protected labels on resources

pankrator:
# Pull Request Template

## Motivation

Protected labels are labels which are not allowed to change from external (API) calls. They should only be modified by Service Manager itself during certain circumstances.

## Approach

HTTP request filter is implemented which accepts configuration which labels are protected.


pkg/label/filter.go
```diff
@@ -0,0 +1,96 @@
...
+			}
+		}
+	}
+	return next.Handle(req)
+}
+
+func (flo *FilterOperations) FilterMatchers() []web.FilterMatcher {
+	return []web.FilterMatcher{
+		{
+			Matchers: []web.Matcher{
```
NickyMateev: Specify `web.Methods(http.MethodPost, http.MethodPatch)` for both Matchers as the filters should not be applied to GET and DELETE operations.

pkg/label/filter.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+func (flo *FilterOperations) Name() string {
+	return "FilterOperations"
+}
+
+func (flo *FilterOperations) Run(req *web.Request, next web.Handler) (*web.Response, error) {
+	if req.Method == http.MethodPost {
+		bodyMap := gjson.ParseBytes(req.Body).Map()
```
NickyMateev: No need to include the variable type into it's name - you can just use `body` and `labels` instead of `bodyMap` and `labesBytes` - the type of the variable is clear from the declaration.

pkg/label/filter.go
```diff
@@ -0,0 +1,96 @@
+package label
```
NickyMateev: Maybe we should move this to something like `pkg/filters/labels` or `pkg/filters/operations`.

Then we can move the security filters from `pkg/security/filters` to `pkg/filters/security` as well.

pkg/label/filter.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/Peripli/service-manager/pkg/types"
+
+	"github.com/Peripli/service-manager/pkg/query"
+	"github.com/Peripli/service-manager/pkg/util"
+	"github.com/Peripli/service-manager/pkg/web"
+	"github.com/tidwall/gjson"
+)
+
+// DenyLabelOperations describe denied label key operations
+type DenyLabelOperations map[string][]query.LabelOperation
```
NickyMateev: ForbiddenLabelOperations?

pkg/label/filter.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func NewFilterOperations(deniedOperations DenyLabelOperations) *FilterOperations {
+	return &FilterOperations{
+		deniedOperations: deniedOperations,
+	}
+}
+
+func (flo *FilterOperations) Name() string {
+	return "FilterOperations"
```
NickyMateev: Not sure if the filter name is descriptive enough

pankrator: ForbiddenLabelOperationsFilter ?

NickyMateev: Sounds good

pkg/filters/labels/filter.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}
+	return nil
+}
+
+type ForibiddenLabelOperationsFilter struct {
+	forbiddenOperations ForbiddenLabelOperations
+}
+
+func NewForbiddenLabelOperationsFilter(forbiddenOperations ForbiddenLabelOperations) *ForibiddenLabelOperationsFilter {
```
KirilKabakchiev: this filter should probably be registered as part of the default api in oss sm and its configuration can be provided by flags,env vars and app yml under api.

pkg/filters/labels/filter.go
```diff
@@ -0,0 +1,134 @@
...
+					}
+				}
+			}
+		}
+	}
+	return next.Handle(req)
+}
+
+func (flo *ForibiddenLabelOperationsFilter) FilterMatchers() []web.FilterMatcher {
+	return []web.FilterMatcher{
```
KirilKabakchiev: probably should match all routes ?

pkg/filters/labels/filter_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/Peripli/service-manager/pkg/util"
+	"github.com/Peripli/service-manager/pkg/web"
+	"github.com/Peripli/service-manager/pkg/web/webfakes"
+
+	"github.com/Peripli/service-manager/pkg/query"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestLabelsFilter(t *testing.T) {
```
KirilKabakchiev: coverage on coveralls says -0.5% on project coverage after this pr which seems pretty high reduction

pkg/filters/labels/filter.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (flo *ForibiddenLabelOperationsFilter) Name() string {
+	return "ForbiddenLabelOperationsFilter"
+}
+
+func (flo *ForibiddenLabelOperationsFilter) Run(req *web.Request, next web.Handler) (*web.Response, error) {
+	if req.Method == http.MethodPost {
+		body := gjson.ParseBytes(req.Body).Map()
+		labelsBytes := []byte(body["labels"].String())
```
KirilKabakchiev: what if the key labels key is not present in the map (nil panic) ?

pkg/filters/labels/filter.go
```diff
@@ -0,0 +1,134 @@
...
+		forbiddenOperations: forbiddenOperations,
+	}
+}
+
+func (flo *ForibiddenLabelOperationsFilter) Name() string {
+	return "ForbiddenLabelOperationsFilter"
+}
+
+func (flo *ForibiddenLabelOperationsFilter) Run(req *web.Request, next web.Handler) (*web.Response, error) {
+	if req.Method == http.MethodPost {
```
KirilKabakchiev: what about delete ?

pkg/filters/labels/filter.go
```diff
@@ -0,0 +1,134 @@
...
+			for _, lo := range labelOperations {
+				if lo == query.AddLabelOperation || lo == query.AddLabelValuesOperation {
+					return nil, &util.HTTPError{
+						ErrorType:   "BadRequest",
+						Description: fmt.Sprintf("Set/Add values for label %s is not allowed", lKey),
+						StatusCode:  http.StatusBadRequest,
+					}
+				}
+			}
+		}
```
KirilKabakchiev: would it be clearer to have separate filters for post, patch and delete rather than having one with big switch and register it with all web.Methods ? I mean the if/switch case in a filter is a bit odd

pkg/filters/labels/filter.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/Peripli/service-manager/pkg/types"
+
+	"github.com/Peripli/service-manager/pkg/query"
+	"github.com/Peripli/service-manager/pkg/util"
+	"github.com/Peripli/service-manager/pkg/web"
+	"github.com/tidwall/gjson"
+)
+
+// ForbiddenLabelOperations describe denied label key operations
+type ForbiddenLabelOperations map[string][]query.LabelOperation
```
KirilKabakchiev: what is the use case behind  allowing to create particular protected label but to disallow deleting a particular protected label  and vise versa ? It seems a bit odd to have a label that you can now have a label that can be deleted but cannot be added

pkg/filters/labels/filter.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (flo ForbiddenLabelOperations) Validate() error {
```
KirilKabakchiev: minor - file ordering (method should be after the struct) and in most places people put the constructor before the type as well although i havent found super strict guidelines for that

pkg/filters/labels/filter.go
pankrator: len on nil map returns 0

pkg/filters/labels/filter.go
pankrator: It is a good idea to separate them. I am just wondering about what names to choose for the different filters. For example:
`ForbiddenLabelCreate` and `ForbiddenLabelPatch`?

pkg/filters/labels/filter.go
pankrator: No particular use case, but in case it is needed in the future

api/api.go
```diff
@@ -119,6 +123,7 @@ func New(ctx context.Context, options *Options) (*web.API, error) {
 			&filters.SelectionCriteria{},
 			&filters.PlatformAwareVisibilityFilter{},
 			&filters.PatchOnlyLabelsFilter{},
+			labels.NewForbiddenLabelOperationsFilter(options.APISettings.ProctedLabels),
```
NickyMateev: Just a quick comment - it might be worth registering this filter before the SelectionCriteria filter as there's no need for some filters to execute if the end result would be that the filter would cut the request eventually

pkg/filters/labels/filter.go
KirilKabakchiev: sounds ok

pkg/filters/labels/filter.go
KirilKabakchiev: wouldnt body["labels"].String()  result in npe ?

PR #301, opened 2019-06-17
Async Healthcheck

DimitarPetrov:
## Motivation

Currently SM healthcheck works synchronously (calls all dependant components synchronously every time a call to /v1/health comes). This is not optimal and introduces unnecessary load both on SM and the dependencies.

 Using https://github.com/InVisionApp/go-health healthchecks can happen asynchronously from the API call. We could remember the results of the last few executions so that we can return 500 on /v1/health only if the pg health indicator has failed 3-4 times in a row. This way we can avoid micro outages (e.g. if we couldnt reach the db for 5min straight, then we report status 500) otherwise we report 200 OK with body that mentions that db is down. This way hopefully we avoid microoutages reports caused by other components. We can leverage logic from the dependency in 1. that would help implement this

## Approach
Each indicator could be configured separately in application.yml. The properties which could be configured are if the indicator is fatal or not, which says if the health of this indicator will affect the overall health. Failures treshold which is how much times in a row an indicator must report down until the component is considered down. And last is the interval time between each check for the component.

Example structure is:
```
...
health:
  indicators:
    storage:
      fatal: true
      failures_treshold: 3
      interval: 30
...
``` 

### Positive response sample
```
{
  "details": {
    "ping": {
      "check_time": "2019-06-25T09:44:23.270103+03:00",
      "fatal": true,
      "first_failure_at": "0001-01-01T00:00:00Z",
      "name": "ping",
      "num_failures": 0,
      "status": "UP"
    },
    "storage": {
      "check_time": "2019-06-25T09:44:23.270114+03:00",
      "fatal": true,
      "first_failure_at": "0001-01-01T00:00:00Z",
      "name": "storage",
      "num_failures": 0,
      "status": "UP"
    }
  },
  "status": "UP"
}
```

### Negative response sample (Treshold not exceeded and overall status is UP)
```
{
  "details": {
    "ping": {
      "check_time": "2019-06-25T09:46:23.272538+03:00",
      "fatal": true,
      "first_failure_at": "0001-01-01T00:00:00Z",
      "name": "ping",
      "num_failures": 0,
      "status": "UP"
    },
    "storage": {
      "check_time": "2019-06-25T09:46:23.274558+03:00",
      "error": "dial tcp [::1]:5432: connect: connection refused",
      "fatal": true,
      "first_failure_at": "2019-06-25T09:46:23.274626+03:00",
      "name": "storage",
      "num_failures": 1,
      "status": "DOWN"
    }
  },
  "status": "UP"
}
```
I made a PR in the library to make some of the fields in response optional since some of them are most of the time zero values like `first_failure_at` and `num_failures` when the component is UP, but still no response.
https://github.com/InVisionApp/go-health/pull/64


KirilKabakchiev:
If we use more than the 100 lines of async collection of healths its worth using it - otherwise not. If we decide to use it we should make sure the pkg package does not expose any imports to this library (e.g. its internal implementation detail of the sm framework)

My idea was about from the async collection of healths to also configure the library logging to use our logger so that we get health logging. 

Also the library stores health failure timestamps and consecutive number of failures and other details which we could look at and decide if we want to enrich our health response with these extra details.
and some predefined healthcheckers such as Reachable, HTTP and SQLDB - but those are not a big deal.

pkg/sm/sm.go
```diff
@@ -182,10 +196,34 @@ func (smb *ServiceManagerBuilder) Build() *ServiceManager {
...
+			Interval: time.Duration(smb.health.Interval) * time.Second,
+		})
+
+		if err != nil {
+			return err
+		}
+	}
+	smb.RegisterControllers(healthcheck.NewController(healthz, smb.HealthAggregationPolicy, smb.health.FailuresTreshold))
+
+	err := healthz.Start()
```
dpanayotov: somewhere you should listen for `ctx.Done()` and call `healthz.Stop()`

pkg/sm/sm.go
```diff
@@ -182,10 +196,34 @@ func (smb *ServiceManagerBuilder) Build() *ServiceManager {
...
+	if err != nil {
```
dpanayotov: inline

storage/interfaces.go
```diff
@@ -182,7 +175,6 @@ type TransactionalRepositoryDecorator func(TransactionalRepository) (Transaction
 //go:generate counterfeiter . Storage
 type Storage interface {
 	OpenCloser
-	Pinger
```
dpanayotov: This works just because the storage implementation is hardcoded in `sm.New` to use `postgres.Storage`. Assuming it is not, this should be reverted.

pkg/health/types.go
```diff
@@ -16,6 +16,36 @@
...
 
+import (
+	"fmt"
+	"github.com/InVisionApp/go-health"
+)
+
+// Settings type to be loaded from the environment
+type Settings struct {
+	FailuresTreshold int64 `mapstructure:"failures_treshold" description:"maximum failures in a row until component is considered down"`
+	Interval         int64 `description:"seconds between health checks of components"`
```
dpanayotov: change this to `time.Duration` and add `mapstructure`. if `time.Duration` adjust the description

pkg/health/types.go
```diff
@@ -16,6 +16,36 @@
...
+		Interval:         60,
+	}
+}
+
+// Validate validates the health settings
+func (s *Settings) Validate() error {
+	if s.FailuresTreshold < 0 {
+		return fmt.Errorf("validate Settings: FailuresTreshold must be >= 0")
+	}
+	if s.Interval < 0 {
```
dpanayotov: I suppose this should be something larger. Like 30s/60s at least?

pkg/health/aggregation_policy.go
```diff
@@ -16,26 +16,40 @@
...
 	details := make(map[string]interface{})
 	for k, v := range healths {
-		details[k] = v
+		details[k] = ConvertStatus(v.Status)
 	}
 	return New().WithStatus(overallStatus).WithDetails(details)
 }
+
+// ConvertStatus converts go-health status to Status
+func ConvertStatus(status string) Status {
```
dpanayotov: I can't decide if it would be better to change our statuses from UP and Down to OK and Failed to avoid this.

pkg/health/aggregation_policy.go
```diff
@@ -16,26 +16,40 @@
...
 // Apply aggregates the given healths
-func (*DefaultAggregationPolicy) Apply(healths map[string]*Health) *Health {
+func (*DefaultAggregationPolicy) Apply(healths map[string]health.State, failureTreshold int64) *Health {
 	if len(healths) == 0 {
 		return New().WithDetail("error", "no health indicators registered").Unknown()
 	}
 	overallStatus := StatusUp
 	for _, health := range healths {
-		if health.Status == StatusDown {
+		if health.Status == "failed" && health.ContiguousFailures > failureTreshold {
```
dpanayotov: `health.ContiguousFailures > failureTreshold` introduces behavior, which I'm not sure we want for all health indicators.
If the storage is down when healthcheck call comes, you might get a response with body that says `failed`, but you'll get status `200 OK`. This will happen if the time between the detection of storage down and the call is less than `failureThreshold * interval`. 
Assuming we configure `interval=60s` and `/v1/monitor/health` is called every 5 minutes - on the 4th minute the storage fails and now on the health call we get `200 OK` but with failure response. Assuming there is a storage outage for 4 minutes (1 minute before the next health call) then the next health call will report UP again. However, the service has not been operational for 4 minutes and all storage operations were failing.

Does it make sense to have the health configuration be a `map[string]Settings` where each health indicator can be separately configured with `failureThreshold` and `interval`? Then we might configure the storage healthcheck to report down even after 1 failure.

api/healthcheck/healthcheck_controller.go
```diff
@@ -17,32 +17,38 @@
...
 }
 
 // healthCheck handler for GET /v1/monitor/health
 func (c *controller) healthCheck(r *web.Request) (*web.Response, error) {
 	ctx := r.Context()
 	logger := log.C(ctx)
-	logger.Debugf("Performing health check with %s...", c.indicator.Name())
-	healthResult := c.indicator.Health()
+	logger.Debugf("Performing health check...")
+	healthState, _, _ := c.health.State()
```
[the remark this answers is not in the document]
DimitarPetrov: Here second argument is overall status of health, which is not needed since we have custom logic for aggregation, and the error is the potential error for overall health failure (error in some of the indicators). This error is not literally ignored because it is presented in the state of the concrete indicator which failed and is processed in aggregation (included into response).

pkg/health/types.go
```diff
@@ -115,6 +177,19 @@ type Registry struct {
...
 	HealthIndicators []Indicator
 
-	// HealthAggregationPolicy is the registered health aggregationPolicy
-	HealthAggregationPolicy AggregationPolicy
+	// Indicator Settings of the registry
+	HealthSettings map[string]*IndicatorSettings
+}
+
+// ConfigureIndicators configures registry's indicators with provided settings
+func (r *Registry) ConfigureIndicators() {
```
dpanayotov: This seems like a hidden extra step that needs to be performed. What do you think about hiding the indicators and add a `func (Registry) AddHealthIndicator(Inidicator)` that configures it?

pkg/sm/sm.go
```diff
@@ -174,9 +182,11 @@ func New(ctx context.Context, cancel context.CancelFunc, cfg *config.Settings) (
 
 // Build builds the Service Manager
 func (smb *ServiceManagerBuilder) Build() *ServiceManager {
-	// setup server and add relevant global middleware
-	smb.installHealth()
+	if err := smb.installHealth(); err != nil {
+		panic(err)
```
dpanayotov: `log.C(smb.ctx).Panic()`

api/healthcheck/healthcheck_controller.go
```diff
@@ -17,32 +17,40 @@
...
 type controller struct {
-	indicator health.Indicator
+	health    h.IHealth
+	tresholds map[string]int64
 }
 
-// NewController returns a new healthcheck controller with the given indicators and aggregation policy
-func NewController(indicators []health.Indicator, aggregator health.AggregationPolicy) web.Controller {
+// NewController returns a new healthcheck controller with the given health and tresholds
+func NewController(health h.IHealth, indicators []health.Indicator) web.Controller {
```
dpanayotov: pass only the minimum that is required. you don't need the whole indicators slice.

pkg/sm/sm.go
```diff
@@ -183,7 +183,7 @@ func New(ctx context.Context, cancel context.CancelFunc, cfg *config.Settings) (
 // Build builds the Service Manager
 func (smb *ServiceManagerBuilder) Build() *ServiceManager {
 	if err := smb.installHealth(); err != nil {
-		panic(err)
+		log.C(smb.ctx).Panic()
```
dpanayotov: Panic(err)

pkg/sm/sm.go
```diff
@@ -132,7 +134,13 @@ func New(ctx context.Context, cancel context.CancelFunc, cfg *config.Settings) (
 		return nil, fmt.Errorf("error creating core api: %s", err)
 	}
 
-	API.HealthIndicators = append(API.HealthIndicators, &storage.HealthIndicator{Pinger: storage.PingFunc(smStorage.Ping)})
+	storageHealthIndicator, err := storage.NewStorageHealthIndicator(storage.PingFunc(smStorage.PingContext))
+	if err != nil {
+		return nil, fmt.Errorf("error creating storage health indicator: %s", err)
+	}
+
+	API.HealthIndicators = append(API.HealthIndicators, storageHealthIndicator)
```
KirilKabakchiev: You could wrap the access to the HealthIndicators in a register method that also adds defaultsettings for this indicator if they are missing and "completes" any incomplete settings for the indicator that is being registered - this will help get rid of the Configurable interface

pkg/health/types.go
```diff
@@ -91,22 +140,35 @@ func (h *Health) WithDetails(details map[string]interface{}) *Health {
...
 }
 
-// AggregationPolicy is an interface to provide aggregated health information
-//go:generate counterfeiter . AggregationPolicy
-type AggregationPolicy interface {
-	// Apply processes the given healths to build a single health
-	Apply(healths map[string]*Health) *Health
+// ConfigurableIndicator is an interface to provide configurable health of a component
+//go:generate counterfeiter . ConfigurableIndicator
+type ConfigurableIndicator interface {
```
KirilKabakchiev: I would suggest tey to get rid of this interface. - thereare some ideas in the other comments

api/healthcheck/healthcheck_controller.go
```diff
@@ -17,32 +17,36 @@
...
 type controller struct {
-	indicator health.Indicator
+	health    h.IHealth
+	tresholds map[string]int64
 }
 
-// NewController returns a new healthcheck controller with the given indicators and aggregation policy
-func NewController(indicators []health.Indicator, aggregator health.AggregationPolicy) web.Controller {
+// NewController returns a new healthcheck controller with the given health and tresholds
+func NewController(health h.IHealth, tresholds map[string]int64) web.Controller {
```
KirilKabakchiev: NewController is a public api that will be used in other projects- so you have two options - go for a stable api (pass settings and know that when new settings are added the controller public api wont change) or pass minimal and risk having to refactor a public api afterwards. Both approaches are acceptable.

pkg/sm/sm.go
```diff
@@ -200,38 +199,31 @@ func (smb *ServiceManagerBuilder) Build() *ServiceManager {
...
+	thresholds := make(map[string]int64)
 
 	for _, indicator := range smb.HealthIndicators {
-		if configurableIndicator, ok := indicator.(health.ConfigurableIndicator); ok {
-			if settings, ok := smb.HealthSettings[configurableIndicator.Name()]; ok {
-				configurableIndicator.Configure(settings)
-			} else {
-				configurableIndicator.Configure(health.DefaultIndicatorSettings())
-			}
+		settings, ok := smb.cfg.Health.Indicators[indicator.Name()]
```
KirilKabakchiev: you should probably also handle partcially configured settings that are present in the map - meaning "complete" them with default values

KirilKabakchiev: Discussed and it seems that now partically configuted indicator will result in confiuration validation error - which is also ok

storage/healthcheck.go
```diff
@@ -16,24 +16,33 @@
...
+import (
+	"github.com/InVisionApp/go-health/checkers"
+	"github.com/Peripli/service-manager/pkg/health"
+)
+
+func NewHealthIndicator(pingFunc PingFunc) (health.Indicator, error) {
+	sqlConfig := &checkers.SQLConfig{
+		Pinger: pingFunc,
+	}
+	sqlChecker, err := checkers.NewSQL(sqlConfig)
```
dpanayotov: this is implementation specific in the storage package. Move this to postgres package or rename this to `SQLHealthIndicator` / `PostgresHealthIndicator`.

PR #303, opened 2019-06-18
Delete by Label Query/Criteria

KirilKabakchiev:
* Allow deletion by label query (delete resources by specifying matching label criteria
* Refactor query builder tests to be usable as documentation for the query by showing the exact queries built

### DIFF
--- a/storage/postgres/query_builder_test.go
+++ b/storage/postgres/query_builder_test.go
@@ -61,12 +61,12 @@ var _ = Describe("Postgres Storage Query builder", func() {
 				_, err := qb.NewQuery().List(ctx, entity)
 				Expect(err).ShouldNot(HaveOccurred())
 				Expect(executedQuery).Should(Equal(trim(`SELECT t.*,
-		visibility_labels.id "visibility_labels.id",
-		visibility_labels.key "visibility_labels.key",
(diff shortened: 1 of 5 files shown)


storage/postgres/query_builder_test.go
```diff
@@ -117,93 +192,252 @@ var _ = Describe("Postgres Storage Query builder", func() {
...
+  (SELECT *
+   FROM visibility_labels
+   WHERE visibility_id IN
+	   (SELECT visibility_id
+		FROM visibility_labels
+		WHERE (visibility_labels.key = ?
+			   AND visibility_labels.val = ?)
+		  OR (visibility_labels.key = ?
+			  AND visibility_labels.val IN (?, ?)))) visibility_labels ON t.id = visibility_labels.visibility_id
+WHERE t.id = visibilities.id;`)))
```
dpanayotov: this is strange. I couldn't find an example on `USING` the way you're using it here. The only documentation shows it is used as an alias in `JOIN` statements.

KirilKabakchiev: yeah but there is no LEFT USING (to replace LEFT JOIN) - so here im self joining first so that i can LEFT JOIN if there are not label queries

PR #305, opened 2019-06-19
Multitenant Resource Access

KirilKabakchiev:
The change introduces two new filters `OIDCTenantCriteriaFilter` and `OIDCTenantLabelingFilter`. Together with `ProtectedLabelsFilter`, the three filters allow for a multitenant resource access in SM.

`OIDCTenantLabelingFilter` allows for automatic labelling of resources that are created (POST) requests via oauth. The resources are labeled with the key which is picked from `TenantCriteriaSettings` (LabelKey) and the value for which (the label value) is picked from the claims of the oauth token (the name of the claim that holds the label value is also specified in TenantCriteriaSettings(LabelTokenClaim). In order for the filter to be triggered, the token must be issued from a special oauth client (multitenant/subuacc oauth client) - this can be changed to a list of oauth clients if necessary. The name of the oauth client that would enable multitenant resource access is specified in TenantCriteriaSettings (ClientID). The claim in the token that contains the client id value for this token is specified in (ClientIDTokenClaim). 

`OIDCTenantCriteriaFilter` allows for automatic filtering of resources returned during LIST,DELETE and PATCH by adding an additional label criteria to the context. It is configured with exactly the same configuration `TenantCriteriaSettings` as `OIDCTenantLabelingFilter`. The filter runs only if the token from the Authorization header was issued by the "multitenant oauth client" described in the previous paragraph.

`ProtectedLabelsFilter` (introduced with an older change) allows for protecting labels to be unmodifiable by HTTP requests. 

The 3 filters complement each other by allowing to introduce a label (let's call it `tenantDelimiter`) that is protected by `ProtectedLabelsFilter`, is added on POST requests in the request body to resources by `OIDCTenantLabelingFilter` (label key is `tenantDelimiter` and label value is picked from the claims of the token and the claim to be used is configrable) and is used as filter label criteria during GET/PATCH/DELETE by `OIDCTenantCriteriaFilter` to only allow access to the tenant specific resources (resources created by this tenant) levergaing the fact that the oauth token contains the tenant delimiter/identifier. So for tokens issued by the "special" mutlitenant/subaccount oauth client(s), service manager would behave in "multitenant" way in respect to all resources that are not transitive (transitive resources are such that don't have a POST API but are created as a result of creating another resource (for example services and plans do not have POST API and are therefore transitive)

The two new filters are always registered but if any of the config properties they require is not present, then these filters would do nothing.

Note: In the current approach `OIDCTenantLabelingFilter` is a filter and not an interceptor and thats why transitive resources cannot be labelled. If we decide we want to label transitive resources we can move it's logic to an interceptor and then transitive services, plans and visibilities will also be labeled  with the tenant identifier.

Note: Currently the filters are registered for brokers and platforms  - if we decide, we could register them for all web.Paths but then `OIDCTenantLabelingFilter` would have to become an interceptor. Currently I've specified explicit web.Paths (brokers and platforms)  for the filters because ottherwise transitive visibilities won't be labeled but visibilities created via /v1/visibilities will be labeled.


KirilKabakchiev:
image

api/api.go
```diff
@@ -41,12 +43,13 @@ const osbVersion = "2.13"
...
-	ProctedLabels     []string `mapstructure:"protected_labels" description:"defines labels which cannot be modified/added by REST API requests"`
-	OSBVersion        string   `mapstructure:"-"`
+	TokenIssuerURL    string                          `mapstructure:"token_issuer_url" description:"url of the token issuer which to use for validating tokens"`
+	ClientID          string                          `mapstructure:"client_id" description:"id of the client from which the token must be issued"`
+	SkipSSLValidation bool                            `mapstructure:"skip_ssl_validation" description:"whether to skip ssl verification when making calls to external services"`
+	TokenBasicAuth    bool                            `mapstructure:"token_basic_auth" description:"specifies if client credentials to the authorization server should be sent in the header as basic auth (true) or in the body (false)"`
+	ProctedLabels     []string                        `mapstructure:"protected_labels" description:"defines labels which cannot be modified/added by REST API requests"`
```
dpanayotov: Should these 2 even be in Peripli?

api/api.go
```diff
@@ -117,9 +122,11 @@ func New(ctx context.Context, options *Options) (*web.API, error) {
 			filters.NewBasicAuthnFilter(options.Repository),
 			bearerAuthnFilter,
 			secfilters.NewRequiredAuthnFilter(),
-			filters.NewForbiddenLabelOperationsFilter(options.APISettings.ProctedLabels),
+			filters.NewProtectedLabelsFilter(options.APISettings.ProctedLabels),
 			&filters.SelectionCriteria{},
 			&filters.PlatformAwareVisibilityFilter{},
+			filters.NewOIDCTenantCriteriaFilter(options.APISettings.TenantCriteria),
```
dpanayotov: if this is to be in Peripli, it should be configurable. I don't think these 3 must be registered by default

dpanayotov: What if I want another tenant criteria and not this one? Do I leave these and add 2 more filters? What if both tenant filters match the requirements for scoping? Should these be filters that can be added and removed or only 1 tenancy rule can be active and it be added to the service-manager itself and not as a filter?

KirilKabakchiev: It would be strange to gave to separate tenant matching filters - if you wanted another tenant criteria you would put it in the Tenant Settings

KirilKabakchiev: Its configrable with the settings - you could specify a different criteria/label to use

api/api.go
KirilKabakchiev: Not necesserily but it makes more sense to have them here in my opinion and just specify our requirements in the settings

api/filters/oidc_tenant_label.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/Peripli/service-manager/pkg/query"
+
+	"github.com/tidwall/gjson"
+
+	"github.com/Peripli/service-manager/pkg/log"
+
+	"github.com/Peripli/service-manager/pkg/web"
+)
+
+// sm-dev adopt and add new config properties in the dev en
```
KirilKabakchiev: This comment should be removed - its a todo for sm-sap and sm local-dev

PR #306, opened 2019-06-20
Fix deletion query building

KirilKabakchiev:
Currently the delete query attempts to return labels. If we delete a visibility that has a one label called `org_id` and three values (org1,org2 and org3) the query will return one row containing the details about the visibility and the first deleted label value (org1). Then the created notification will include only one value (org1) in the payload for the label `org_id`  and the proxy will disable the service plan access for only this org and miss disabling for the other two.

There is no great benefit in trying to complicate the query further and try to return all deleted labels as deleting the resource will delete all labels linked to it anyway (unless we find some neat solution). 
Proposed solution with this PR is to revert to not returning information about labels when deleting a resource (this was the old implementation before introducing delete by label query). Delete by label query will still work.

### DIFF
--- a/storage/postgres/query_builder.go
+++ b/storage/postgres/query_builder.go
@@ -221,21 +221,23 @@ func (pgq *pgQuery) limitSQL() *pgQuery {
 
 func (pgq *pgQuery) returningSQL(entity PostgresEntity) *pgQuery {
 	for i := range pgq.returningFields {
-		pgq.returningFields[i] = fmt.Sprintf("%s.%s", mainTableAlias, pgq.returningFields[i])
-	}
-
-	if len(pgq.returningFields) != 0 {
-		labelEntity := entity.LabelEntity()
-		if labelEntity != nil {
-			labelsTableName := labelEntity.LabelsTableName()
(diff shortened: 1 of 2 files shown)


dpanayotov:
Delete returning all
```sql
WITH visibility_labels AS(
    DELETE FROM visibility_labels WHERE visibility_id = '12345' RETURNING *
), visibilities AS (
    DELETE FROM visibilities WHERE id = '12345' RETURNING *
)   SELECT t.*,
           visibility_labels.id "visibility_labels.id",
           visibility_labels.key "visibility_labels.key",
           visibility_labels.val "visibility_labels.val",
           visibility_labels.created_at "visibility_labels.created_at",
           visibility_labels.updated_at "visibility_labels.updated_at",
           visibility_labels.visibility_id "visibility_labels.visibility_id"
FROM visibilities t
LEFT JOIN visibility_labels ON t.id = visibility_labels.visibility_id;
```

KirilKabakchiev:
We should consider if it would make sense to return the labels also as this query seems like managable effort to implement - for delete visibility in cf it just happens that we dont need them but this might not be true for other platforms

PR #307, opened 2019-06-20
 Make test SM server listener replaceable

NickyMateev:
#  Make test SM server listener replaceable

## Motivation

Some tests might require to provide a custom server listener, ex. in order to control on which port it is deployed and be able to construct accurate expectations.

## Approach

Introduce a new `BuildWithListener` method for building `TestContext` structs with a predefined listener.

## Pull Request status

* [x] Initial implementation

### DIFF
--- a/test/common/test_context.go
+++ b/test/common/test_context.go
@@ -271,15 +271,14 @@ func (tcb *TestContextBuilder) BuildWithListener(listener net.Listener) *TestCon
 func NewSMListener() (net.Listener, error) {
 	minPort := 8100
 	maxPort := 9999
-
-	rand.Seed(time.Now().UnixNano())
-	port := rand.Intn(maxPort-minPort) + minPort
-
 	retries := 10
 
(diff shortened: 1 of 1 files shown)


test/common/test_context.go
```diff
@@ -260,7 +268,29 @@ func (tcb *TestContextBuilder) Build() *TestContext {
...
+	maxPort := 9999
+
+	rand.Seed(time.Now().UnixNano())
+	port := rand.Intn(maxPort-minPort) + minPort
+
+	retries := 10
+
+	var listener net.Listener
+	var err error
+	for ; retries >= 0; retries-- {
```
dpanayotov: why the retries

NickyMateev: In case the random selected random port is already selected, I think it would be best to try a different port, so that the execution does not stop. But I see now that I haven't placed the logic for random port number generation inside the loop, will fix it now..

PR #309, opened 2019-06-30
Fix postgres Listen not called in some cases

georgifarashev:
# Fix postgres Listen not called in some cases

## Scenario

If only one proxy is connected to SM it should receive notifications.
When this proxy disconnects, SM calls Unlisted to postgres so that it does not receive unnecessary notifications. When a proxy reconnects again SM does not call listen again.

## Fix

Always call unlisten/listen when proxy disconnects/connects. 
Also improve logging for these scenarios.

### DIFF
--- a/api/notifications/notifications.go
+++ b/api/notifications/notifications.go
@@ -60,7 +60,7 @@ func (c *Controller) handleWS(req *web.Request) (*web.Response, error) {
 	}
 
 	correlationID := logger.Data[log.FieldCorrelationID].(string)
-	childCtx := newContextWithCorrelationID(c.baseCtx, correlationID)
+	childCtx, childCtxCancel := newContextWithCorrelationID(c.baseCtx, correlationID)
 
 	defer func() {
 		if err := recover(); err != nil {
@@ -82,7 +82,7 @@ func (c *Controller) handleWS(req *web.Request) (*web.Response, error) {
 
 	done := make(chan struct{}, 2)
 
-	go c.closeConn(childCtx, conn, done)
+	go c.closeConn(childCtx, childCtxCancel, conn, done)
 	go c.writeLoop(childCtx, conn, notificationQueue, done)
 	go c.readLoop(childCtx, conn, done)
 
@@ -173,7 +173,8 @@ func extractPlatformFromContext(userContext *web.UserContext) (*types.Platform, 
 	return platform, nil
 }
 
-func newContextWithCorrelationID(baseCtx context.Context, correlationID string) context.Context {
+func newContextWithCorrelationID(baseCtx context.Context, correlationID string) (context.Context, context.CancelFunc) {
 	entry := log.C(baseCtx).WithField(log.FieldCorrelationID, correlationID)
-	return log.ContextWithLogger(baseCtx, entry)
+	newCtx := log.ContextWithLogger(baseCtx, entry)
+	return context.WithCancel(newCtx)
 }
--- a/storage/postgres/notificator.go
+++ b/storage/postgres/notificator.go
@@ -42,7 +42,7 @@ const (
 )
 
 type Notificator struct {
-	isListening bool
+	isListening bool // To be used only under connectionMutex.Lock
 	isConnected int32
 
 	queueSize int
@@ -118,6 +118,8 @@ func (n *Notificator) Start(ctx context.Context, group *sync.WaitGroup) error {
 }
 
 func (n *Notificator) addConsumer(platform *types.Platform, queue storage.NotificationQueue) (int64, error) {
+	// must listen and add consumer under connectionMutex lock as UnregisterConsumer
+	// might stop notification processing if no other consumers are present
 	n.connectionMutex.Lock()
 	defer n.connectionMutex.Unlock()
 	if !n.isListening {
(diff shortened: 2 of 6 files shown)


storage/postgres/notificator.go
```diff
@@ -257,23 +267,53 @@ type notifyEventPayload struct {
...
+			}
+			lastNotificationReceived = time.Now()
+			log.C(n.ctx).Debugf("Received new notification from channel %s", pqNotification.Channel)
+			payload, err := getPayload(pqNotification.Extra)
+			if err != nil {
+				log.C(n.ctx).WithError(err).Error("Could not unmarshal notification payload")
+				return
+			} else {
+				if err = n.processNotificationPayload(payload); err != nil {
+					log.C(n.ctx).WithError(err).Error("Closing consumers")
```
dpanayotov: the error could be `could not process notification payload. Closing consumers...`

storage/postgres/notificator.go
```diff
@@ -257,23 +267,53 @@ type notifyEventPayload struct {
...
+					return
+				}
 			}
+		case <-time.After(n.dbPingInterval):
+			log.C(n.ctx).Debugf("No notifications in %s. Pinging db", time.Since(lastNotificationReceived))
```
dpanayotov: pinging connection

storage/postgres/notificator.go
```diff
@@ -323,40 +365,31 @@ func (n *Notificator) getRecipients(platformID string) []*types.Platform {
...
 	}
 }
 
 func (n *Notificator) stopConnection() {
-	err := n.stopListening()
-	logger := log.C(n.ctx)
-	if err != nil {
-		logger.WithError(err).Info("could not unlisten notification channel")
-	}
+	n.stopProcessing()
```
dpanayotov: write godoc that you expect this to be caught in `processNotifications` which would in turn close all consumers and unlisten

storage/postgres/notificator.go
```diff
@@ -323,40 +365,31 @@ func (n *Notificator) getRecipients(platformID string) []*types.Platform {
...
-		logger.WithError(err).Info("could not close db connection")
-	}
-}
 
-func (n *Notificator) stopListening() error {
-	n.connectionMutex.Lock()
-	defer n.connectionMutex.Unlock()
-	if atomic.LoadInt32(&n.isListening) == aFalse {
-		return nil
+	atomic.StoreInt32(&n.isConnected, aFalse)
```
dpanayotov: can it happen that `connection.Close` fails but `isConnected` is `true`? as this is called only when SM is stopped I suppose this won't be a problem, but still

storage/postgres/notificator.go
```diff
@@ -95,10 +100,11 @@ func (n *Notificator) Start(ctx context.Context, group *sync.WaitGroup) error {
 	n.setConnection(n.connectionCreator.NewConnection(func(isConnected bool, err error) {
 		if isConnected {
 			atomic.StoreInt32(&n.isConnected, aTrue)
+			log.C(n.ctx).Debug("db connection for notifications established")
 		} else {
 			atomic.StoreInt32(&n.isConnected, aFalse)
-			log.C(n.ctx).WithError(err).Info("connection to db closed, closing all consumers")
-			n.closeAllConsumers()
+			log.C(n.ctx).WithError(err).Error("DB connection for notifications closed")
+			n.stopProcessing()
```
dpanayotov: add godoc/log what this does because it's somewhat hidden (applies to every place calling `stopProcessing`)

storage/postgres/notificator.go
```diff
@@ -95,10 +100,11 @@ func (n *Notificator) Start(ctx context.Context, group *sync.WaitGroup) error {
 	n.setConnection(n.connectionCreator.NewConnection(func(isConnected bool, err error) {
 		if isConnected {
 			atomic.StoreInt32(&n.isConnected, aTrue)
+			log.C(n.ctx).Debug("db connection for notifications established")
 		} else {
 			atomic.StoreInt32(&n.isConnected, aFalse)
-			log.C(n.ctx).WithError(err).Info("connection to db closed, closing all consumers")
-			n.closeAllConsumers()
+			log.C(n.ctx).WithError(err).Error("DB connection for notifications closed")
```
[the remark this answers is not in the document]
georgifarashev: As this is caused by connection disruption with the DB and it is a severe problem.

api/notifications/notifications.go
```diff
@@ -82,7 +82,7 @@ func (c *Controller) handleWS(req *web.Request) (*web.Response, error) {
 
 	done := make(chan struct{}, 2)
 
-	go c.closeConn(childCtx, conn, done)
+	go c.closeConn(childCtx, childCtxCancel, conn, done)
```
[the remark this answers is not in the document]
georgifarashev: Unfortunately in the `readLoop` the `conn.ReadMessage()` blocks until the connection is closed. So we have to close the connection in order to stop the go routine.

storage/postgres/notificator.go
```diff
@@ -115,11 +117,34 @@ func (n *Notificator) Start(ctx context.Context, group *sync.WaitGroup) error {
 	return nil
 }
 
-func (n *Notificator) addConsumer(platform *types.Platform, queue storage.NotificationQueue) int64 {
+func (n *Notificator) addConsumer(platform *types.Platform, queue storage.NotificationQueue) (int64, error) {
+	n.connectionMutex.Lock()
```
dpanayotov: add comment that these two must be done in the same lock

storage/postgres/notificator.go
```diff
@@ -42,8 +42,8 @@ const (
 )
 
 type Notificator struct {
+	isListening bool
```
dpanayotov: why remove the atomic assignment?

georgifarashev: It is not necessary as isListening is only used under connectionMutex.Lock

PR #310, opened 2019-07-01
Test context can now be configured with requestor for some operations

pankrator:
This is needed, because we want to configure the type of requestor (authentication) to use when create/update/delete some resources


pankrator:
No need for this

PR #314, opened 2019-07-24
Add additional correlation id header

dpanayotov:


pkg/log/request.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 // CorrelationIDHeaders are the headers whose values will be taken as a correlation id for incoming requests
-var CorrelationIDHeaders = []string{"X-Correlation-ID", "X-CorrelationID", "X-ForRequest-ID", "X-Request-ID"}
+var CorrelationIDHeaders = []string{"X-Correlation-ID", "X-CorrelationID", "X-ForRequest-ID", "X-Request-ID", "X-Vcap-Request-Id"}
```
dotchev: Is this the right order?
It seems this should be configurable for each deployment depending on the infrastructure.
I guess X-Vcap-* is relevant only for CF.

fmui:
And while you are on it, there is a correlation header now in the OSB sepc: https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#request-identity

pkg/log/request.go
dpanayotov: What do you mean by order?
There is a FixIt item to allow this list to be extensible/replaced per deployment.

PR #315, opened 2019-07-24
Add timeout for osb calls

georgifarashev:
# Add timeout for osb calls

## Motivation

Currently osb calls do not have timeout. This could lead to out of memory if multiple calls hang to a broken broker.

## Approach

Introduce timeout configurations for all outbound communication.

## Pull Request status

* [x] Initial implementation
* [x] Tests

### DIFF
--- a/pkg/sm/sm.go
+++ b/pkg/sm/sm.go
@@ -21,6 +21,7 @@ import (
 	"crypto/tls"
 	"database/sql"
 	"fmt"
+	"net"
 	"net/http"
 	"sync"
 
@@ -75,10 +76,16 @@ func New(ctx context.Context, cancel context.CancelFunc, cfg *config.Settings) (
 		return nil, fmt.Errorf("error validating configuration: %s", err)
 	}
 
-	// Setup the default http client
-	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: cfg.API.SkipSSLValidation}
-	http.DefaultClient.Transport = http.DefaultTransport
-	http.DefaultClient.Timeout = cfg.Server.RequestTimeout
+	// Setup the default http client and transport
+	transport := http.DefaultTransport.(*http.Transport)
+
+	transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: cfg.API.SkipSSLValidation}
+	transport.ResponseHeaderTimeout = cfg.HTTPClient.ResponseHeaderTimeout
+	transport.TLSHandshakeTimeout = cfg.HTTPClient.TLSHandshakeTimeout
+	transport.IdleConnTimeout = cfg.HTTPClient.IdleConnTimeout
+	transport.DialContext = (&net.Dialer{Timeout: cfg.HTTPClient.DialTimeout}).DialContext
+
+	http.DefaultClient.Transport = transport
 
 	// Setup logging
 	ctx = log.Configure(ctx, cfg.Log)
--- a/test/broker_test/broker_test.go
+++ b/test/broker_test/broker_test.go
@@ -249,24 +249,24 @@ var _ = test.DescribeTestsFor(test.TestCase{
 					})
 				})
 
-				Context("when the broker call for catalog times out ", func() {
+				Context("when the broker call for catalog times out", func() {
+					const (
+						timeoutDuration             = time.Millisecond * 500
+						additionalDelayAfterTimeout = time.Second
+					)
+
 					var (
-						timeoutTestCtx          *common.TestContext
-						brokerCatalogStopCancel context.CancelFunc
+						timeoutTestCtx *common.TestContext
 					)
 
 					BeforeEach(func() {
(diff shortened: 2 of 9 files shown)


test/common/test_context.go
```diff
@@ -67,8 +67,9 @@ type TestContextBuilder struct {
 
 	shouldSkipBasicAuthClient bool
 
-	Environment func(f ...func(set *pflag.FlagSet)) env.Environment
-	Servers     map[string]FakeServer
+	Environment    func(f ...func(set *pflag.FlagSet)) env.Environment
+	Servers        map[string]FakeServer
+	TestHttpClient *http.Client
```
KirilKabakchiev: Probably HTTPClient should be sufficient as name

test/osb_test/osb_test.go
```diff
@@ -69,7 +73,7 @@ func assertMissingBrokerError(req *httpexpect.Response) {
 		Value("description").String().Contains("could not find such broker")
 }
 
-func assertStoppedBrokerError(req *httpexpect.Response) {
+func assertBadBrokerError(req *httpexpect.Response) {
```
KirilKabakchiev: Unresponsive instead of bad? Bad does not bring much semantics

test/osb_test/osb_test.go
```diff
@@ -148,13 +154,68 @@ var _ = Describe("Service Manager OSB API", func() {
...
+		failingBrokerServer.BindingLastOpHandler = failingHandler
+		failingBrokerServer.BindingAdaptCredentialsHandler = failingHandler
+
+		queryParameterVerificationServer.ServiceInstanceHandler = queryParameterVerificationHandler(headerKey, headerValue)
+		queryParameterVerificationServer.BindingHandler = queryParameterVerificationHandler(headerKey, headerValue)
+		queryParameterVerificationServer.CatalogHandler = queryParameterVerificationHandler(headerKey, headerValue)
+		queryParameterVerificationServer.ServiceInstanceLastOpHandler = queryParameterVerificationHandler(headerKey, headerValue)
+		queryParameterVerificationServer.BindingLastOpHandler = queryParameterVerificationHandler(headerKey, headerValue)
+
+		timeoutBrokerReleaseCtx, timeoutBrokerRelease = context.WithCancel(context.Background())
```
KirilKabakchiev: Does Delaying make more.eense than stopping (it's not really stopping

test/osb_test/osb_test.go
```diff
@@ -148,13 +154,68 @@ var _ = Describe("Service Manager OSB API", func() {
...
+		stoppingHandler := func(rw http.ResponseWriter, req *http.Request) {
```
KirilKabakchiev: Would it make sense to extract as constant with meaningful name instead of the comment? Something like additionalDelayAfterTimeout

test/broker_test/broker_test.go
```diff
@@ -246,6 +249,45 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+						timeoutTestCtx          *common.TestContext
+						brokerCatalogStopCancel context.CancelFunc
+					)
+
+					BeforeEach(func() {
+						timeoutDuration := time.Millisecond * 500
+						timeoutTestCtx = common.NewTestContextBuilder().WithEnvPreExtensions(func(set *pflag.FlagSet) {
+							Expect(set.Set("httpclient.response_header_timeout", timeoutDuration.String())).ToNot(HaveOccurred())
+						}).Build()
+
```
KirilKabakchiev: Do we need a separate context or can we reuse the one coming from the request ?

test/broker_test/broker_test.go
```diff
@@ -246,6 +249,45 @@ var _ = test.DescribeTestsFor(test.TestCase{
 					})
 				})
 
+				Context("when the broker call for catalog times out ", func() {
```
KirilKabakchiev: Unnecessary space at the end

pkg/sm/sm.go
```diff
@@ -75,10 +76,16 @@ func New(ctx context.Context, cancel context.CancelFunc, cfg *config.Settings) (
...
+	// Setup the default http client and transport
+	transport := http.DefaultTransport.(*http.Transport)
+
+	transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: cfg.API.SkipSSLValidation}
+	transport.ResponseHeaderTimeout = cfg.HTTPClient.ResponseHeaderTimeout
+	transport.TLSHandshakeTimeout = cfg.HTTPClient.TLSHandshakeTimeout
+	transport.IdleConnTimeout = cfg.HTTPClient.IdleConnTimeout
+	transport.DialContext = (&net.Dialer{Timeout: cfg.HTTPClient.DialTimeout}).DialContext
+
+	http.DefaultClient.Transport = transport
```
KirilKabakchiev: Add a fixit task to remove change from the default client to a seperatly built one. As a reasoning you may give the case with httpexpect and the test context requiring to manually copy the default transport

PR #316, opened 2019-07-25
Include failing request details in the error

dotchev:
## Motivation

Sometimes Service Manager logs errors like this

> Call to client failed with: StatusCode: 404 Body: ...

It is not clear from the error message which service was called.

## Approach

Include also the request method and url in the error message

Logging is removed from `HandleResponseError` to prevent double logging. The error should be logged by the final error handler.

## Pull Request status

* [x] Initial implementation
* [x] Refactoring
* [x] Unit tests


pkg/util/errors.go
```diff
@@ -76,20 +76,18 @@ func WriteError(ctx context.Context, err error, writer http.ResponseWriter) {
...
-		return fmt.Errorf("error processing response body of resp with status code %d: %s", response.StatusCode, err)
+		body = []byte(fmt.Sprintf("error reading response body: %s", err))
 	}
 
 	err = fmt.Errorf("StatusCode: %d Body: %s", response.StatusCode, body)
 	if response.Request != nil {
-		log.C(response.Request.Context()).Errorf("Call to client failed with: %s", err)
-	} else {
-		log.D().Errorf("Call to client failed with: %s", err)
+		return fmt.Errorf("request %s %s failed: %s", response.Request.Method, response.Request.URL, err)
```
KirilKabakchiev: This will likely propagate information about an internal error (example SM internally calling another system as a client) to the api caller. But it will also make troubleshooting easier. So i am fine either way

KirilKabakchiev: discussed and as error is not HTTPError, no internal details will be propagated.

dotchev: here a plain error is returned, but only `HTTPError` is sent in the http response
so this should result in status 500 unless it is handled earlier

PR #317, opened 2019-08-06
Filter resources based on visibilities (only for platform)

pankrator:
# Pull Request Template

## Motivation

K8S proxies should be able to list only brokers, services and plans that are visible for them.

## Approach

Add filter for each resource based on the requesting platform's visibilities.
1) Get all visibilities for the platform including empty platform visibilities(public).
2) Get all plans filtered by the visibilities from 1)
3) Get all services filtered by the plans from 2)
4) Get all brokers filtered by the services from 3)

## Pull Request status

* [x] Initial implementation
* [x] Component tests


api/filters/brokers_filter_by_visibility.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return nil, err
+	}
+	offerings := objectList.(*types.ServiceOfferings)
+	if offerings.Len() < 1 {
+		return nil, nil
+	}
+	brokerIDs := make([]string, 0, offerings.Len())
+	for _, p := range offerings.ServiceOfferings {
+		brokerIDs = append(brokerIDs, p.BrokerID)
```
georgifarashev: Should we remove the duplicates? 2 offerings might have the same BrokerID. Maybe use a map and then collect the keys.

api/filters/brokers_filter_by_visibility.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return nil, err
+	}
+	plans := objectList.(*types.ServicePlans)
+	if plans.Len() < 1 {
+		return nil, nil
+	}
+	serviceIDs := make([]string, 0, plans.Len())
+	for _, p := range plans.ServicePlans {
+		serviceIDs = append(serviceIDs, p.ServiceOfferingID)
```
georgifarashev: Remove duplicate serviceIDs? See comment above.

api/filters/brokers_filter_by_visibility.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return nil, err
+	}
+	visibilityList := objectList.(*types.Visibilities)
+	if visibilityList.Len() < 1 {
+		return nil, nil
+	}
+	planIDs := make([]string, 0, visibilityList.Len())
+	for _, vis := range visibilityList.Visibilities {
+		planIDs = append(planIDs, vis.ServicePlanID)
```
georgifarashev: Remove duplicate planIDs? See comment above.

api/filters/brokers_filter_by_visibility.go
```diff
@@ -0,0 +1,125 @@
...
+	c := query.ByField(query.InOperator, "id", planIDs...)
+	return &c, nil
+}
+
+func (vf *BrokersFilterByVisibility) FilterMatchers() []web.FilterMatcher {
+	return []web.FilterMatcher{
+		{
+			Matchers: []web.Matcher{
+				web.Path(web.ServiceBrokersURL + "/**"),
+				web.Methods(http.MethodGet, http.MethodPatch, http.MethodPut, http.MethodDelete),
```
georgifarashev: Probably makes sense to be only on GET.

api/filters/services_filter_by_visibility.go
```diff
@@ -0,0 +1,59 @@
...
+	return func(ctx context.Context, platformID string) (*query.Criterion, error) {
+		planQuery, err := plansCriteria(ctx, repository, platformID)
+		if err != nil {
+			return nil, err
+		}
+		if planQuery == nil {
+			return nil, nil
+		}
+
+		servicesQuery, err := servicesCriteria(ctx, repository, planQuery)
```
georgifarashev: just ```return servicesCriteria(ctx, repository, planQuery)``` should be sufficient.

api/filters/visibility_filtering_middleware.go
```diff
@@ -0,0 +1,56 @@
...
+	userCtx, ok := web.UserFromContext(ctx)
+	if !ok {
+		return nil, errors.New("No user found")
+	}
+	if userCtx.AuthenticationType != web.Basic {
+		log.C(ctx).Debugf("Authentication is %s, not basic so proceed without visibility filter criteria", userCtx.AuthenticationType)
+		return next.Handle(req)
+	}
+	platform := &types.Platform{}
+	if err := userCtx.Data(platform); err != nil {
```
georgifarashev: We should check if the platform is CF we should not filter. There we expect all brokers and plans to be available and apply the visibilities directly in the platform. This will result in a nasty ```if``` thought :(

api/filters/visibility_filtering_middleware.go
```diff
@@ -0,0 +1,56 @@
...
+		for _, serviceID := range finalQuery.RightOp {
+			if serviceID == resourceID {
+				foundResource = true
+				break
+			}
+		}
+	}
+
+	if finalQuery == nil || (!foundResource && resourceID != "") {
+		ctx = query.ContextWithCriteria(ctx, []query.Criterion{query.LimitResultBy(0)})
```
georgifarashev: Why not just stop the request with empty result?

api/filters/visibility_filtering_middleware.go
```diff
@@ -0,0 +1,56 @@
...
+	if err := userCtx.Data(platform); err != nil {
+		return nil, err
+	}
+
+	finalQuery, err := m.FilteringFunc(ctx, platform.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	resourceID := req.PathParams["id"]
```
georgifarashev: The FilteringFunc makes db calls which transfer all filtered resources. Could we instead introduce filtering only for one element instead of creating the finalQuery and then check if the id is in this query? This would save us a bit of db bandwidth :)

api/filters/visibility_filtering_middleware.go
pankrator: We don't know the type of the resource being filtered

api/filters/visibility_filtering_middleware.go
pankrator: I will try and see how it looks like

api/filters/catalog_filter_by_visibility.go
```diff
@@ -0,0 +1,217 @@
...
+		return next.Handle(req)
+	}
+
+	res, err := next.Handle(req)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != http.StatusOK {
+		return res, nil
+	}
```
georgifarashev: These two ifs can be combined:
```go
if err != nil || res.StatusCode != http.StatusOK {
	return res, err;
}
```

api/filters/visibility_filtering_middleware.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (m visibilityFilteringMiddleware) Run(req *web.Request, next web.Handler) (*web.Response, error) {
+	ctx := req.Context()
+	userCtx, ok := web.UserFromContext(ctx)
+	if !ok {
+		return nil, errors.New("no user found")
+	}
+	if userCtx.AuthenticationType != web.Basic {
+		log.C(ctx).Debugf("Authentication is %s, not basic so proceed without visibility filter criteria", userCtx.AuthenticationType)
```
KirilKabakchiev: Is there a test case that filtering is not applied for SMWithOauth?

test/service_plan_test/service_plan_test.go
```diff
@@ -96,6 +96,138 @@ var _ = test.DescribeTestsFor(test.TestCase{
 				})
 			})
 
+			Describe("GET", func() {
+				var k8sPlatform *types.Platform
+				var k8sAgent *httpexpect.Expect
+
+				BeforeEach(func() {
+					k8sPlatformJSON := common.MakePlatform("k8s-platform", "k8s-platform", "kubernetes", "test-platform-k8s")
```
KirilKabakchiev: Do we need this platform or can we use smWithBasic and the platform from the test context?

api/filters/visibility_filtering_middleware.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/Peripli/service-manager/pkg/log"
+	"github.com/Peripli/service-manager/pkg/query"
+	"github.com/Peripli/service-manager/pkg/types"
+	"github.com/Peripli/service-manager/pkg/web"
+)
+
+const k8sPlatformType string = "kubernetes"
+
+type visibilityFilteringMiddleware struct {
+	SingleResourceFilterFunc func(ctx context.Context, resourceID, platformID string) (bool, error)
```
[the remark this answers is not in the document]
pankrator: Perfect naming :D :bowing_man:

api/filters/visibility_filtering_middleware.go
```diff
@@ -0,0 +1,136 @@
...
+	if err := userCtx.Data(platform); err != nil {
+		return nil, err
+	}
+	if platform.Type != k8sPlatformType {
+		log.C(ctx).Debugf("Platform type is %s, which is not kubernetes. Skip filtering on visibilities", platform.Type)
+		return next.Handle(req)
+	}
+
+	resourceID := req.PathParams["id"]
+	isSingleResource := (req.PathParams["id"] != "")
```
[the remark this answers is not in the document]
pankrator: I agree about the middleware responsibilities, but I think there will be more filters to register and follow along the request flow.

api/filters/visibility_filtering_middleware.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if finalQuery == nil {
+		return util.NewJSONResponse(http.StatusOK, m.EmptyResourceArrayFunc())
+	}
+
+	criterion := query.CriteriaForContext(ctx)
+	merged := false
+	for i, c := range criterion {
```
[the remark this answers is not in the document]
pankrator: Already discussed

test/osb_test/osb_test.go
```diff
@@ -336,6 +344,78 @@ var _ = Describe("Service Manager OSB API", func() {
...
+			Context("for platform with no visibilities", func() {
+				It("should return empty services catalog", func() {
+					k8sAgent.GET(fmt.Sprintf("%s/%s/v2/catalog", web.OSBURL, brokerID)).
+						Expect().Status(http.StatusOK).JSON().Object().Value("services").Array().Length().Equal(0)
+				})
+			})
+
+			Context("for platform with public visibility for one plan", func() {
+				It("should return only the plan with visibility", func() {
+					ctx.SMWithOAuth.POST(web.VisibilitiesURL).WithJSON(common.Object{
```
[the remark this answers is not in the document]
pankrator: We know that this broker should return only one plan, so in this case the length check is required. Other than that, we check the ID of the returned plan to make sure this is the plan we expect

test/osb_test/osb_test.go
```diff
@@ -336,6 +344,78 @@ var _ = Describe("Service Manager OSB API", func() {
...
+			})
+
+			Context("for platform with no visibilities", func() {
+				It("should return empty services catalog", func() {
+					k8sAgent.GET(fmt.Sprintf("%s/%s/v2/catalog", web.OSBURL, brokerID)).
+						Expect().Status(http.StatusOK).JSON().Object().Value("services").Array().Length().Equal(0)
+				})
+			})
+
+			Context("for platform with public visibility for one plan", func() {
```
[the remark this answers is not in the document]
pankrator: You cannot have empty plans, when there are public visibilities for these plans. Or may be I don't understand the case. Can you give more details

test/osb_test/osb_test.go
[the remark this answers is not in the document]
pankrator: I agree about the part with assert before the operation and after the operation

test/osb_test/osb_test.go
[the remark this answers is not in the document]
pankrator: Nevermind, I didn't read the text carefully

test/service_plan_test/service_plan_test.go
pankrator: The filtering applies only for K8S platforms, so we need to register a new one

api/filters/visibility_filtering_middleware.go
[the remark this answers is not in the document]
KirilKabakchiev: Currently the code in this filter is a bit hard to follow

api/filters/catalog_filter_by_visibility.go
```diff
@@ -0,0 +1,183 @@
...
+	if userCtx.AuthenticationType != web.Basic {
+		log.C(ctx).Debugf("Authentication is %s, not basic so proceed without visibility filter criteria", userCtx.AuthenticationType)
+		return next.Handle(req)
+	}
+	platform := &types.Platform{}
+	if err := userCtx.Data(platform); err != nil {
+		return nil, err
+	}
+	if platform.Type != k8sPlatformType {
+		log.C(ctx).Debugf("Platform type is %s, which is not kubernetes. Skip filtering on visibilities", platform.Type)
```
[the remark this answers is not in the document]
KirilKabakchiev: this is a an if that is probably going to be refactored pretty soon due to changing scenarios - i just want to make sure that we have concrete test cases that verify that after refactoring the behaviour is the same (in this case it would be (when fetching catalog for cloudfoundry platform and there are no visibilities for service x plan y, it is still part of the returned catalog)

test/service_plan_test/service_plan_test.go
```diff
@@ -96,6 +96,138 @@ var _ = test.DescribeTestsFor(test.TestCase{
 				})
 			})
 
+			Describe("GET", func() {
```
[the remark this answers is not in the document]
pankrator: There are already such tests

test/common/common.go
```diff
@@ -211,6 +213,29 @@ func RegisterBrokerInSM(brokerJSON Object, SM *httpexpect.Expect, headers map[st
...
 }
 
+func RegisterVisibilityForPlanAndPlatform(SM *httpexpect.Expect, planID, platformID string) {
+	SM.POST(web.VisibilitiesURL).WithJSON(Object{
+		"service_plan_id": planID,
+		"platform_id":     platformID,
+	}).Expect().Status(http.StatusCreated)
+}
+
+func RegisterVisibilityForBrokerID(SM *httpexpect.Expect, brokerID string) {
```
NickyMateev: What this function really does is essentially "publicize" all plans for a broker with the given ID by creating visibilities for each one. Why don't you rename the function to something more descriptive like "_PublicizeAllBrokerPlans_" or something similar?

test/delete_list.go
```diff
@@ -266,15 +266,6 @@ func DescribeDeleteListFor(ctx *common.TestContext, t TestCase) bool {
 				expectedStatusCode: http.StatusBadRequest,
 			},
 		),
-		Entry("returns 400 when query is duplicated",
```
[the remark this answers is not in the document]
pankrator: I agree with you, but component tests are already too complex to refactor and field query and label query use the same Test entry template. Also the label query does not support for duplicate keys and this test won't work...

test/common/common.go
pankrator: It is more like: `CreateVisibilitiesForAllBrokerPlans`. What do you think?

NickyMateev: Yeah, sounds good

KirilKabakchiev:
image

PR #319, opened 2019-08-15
Dynamic Log Switching

KirilKabakchiev:
# Pull Request Template

## Motivation

Allows dynamically switching log levels of SM's loggers without requiring app restart.

## Approach

Two ways to dynamically switch the application log level are introduced:
1. A new PUT API /v1/config/logging that accepts  a logging configuration (log format and log level). The input from the API is used to reconfigure the loggers
2. File watchers on the application.yml (meaning if one changes the application.yml (or whatever config file is used), the application loggers will be reconfigured

## Pull Request status

For example:

* [x] Initial implementation
* [x] Refactoring
* [x] Unit tests
* [x] Integration tests


api/configuration/logging_controller.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"net/http"
+
+	"github.com/Peripli/service-manager/pkg/log"
+	"github.com/Peripli/service-manager/pkg/util"
+	"github.com/Peripli/service-manager/pkg/web"
+)
+
+// Controller logging configuration controller
```
dpanayotov: If this is only for logging, rename to `configuration.LoggingController`. Otherwise change the godoc

pkg/env/env.go
```diff
@@ -143,18 +146,46 @@ func (v *ViperEnv) setupConfigFile() error {
 		}
 		return fmt.Errorf("could not read configuration cfg: %s", err)
 	}
+
+	v.Viper.WatchConfig()
+
+	dynamicLogHandler := func(env Environment) func(event fsnotify.Event) {
```
dpanayotov: this would be better to be in the log package, as this package is not be interested in log config changes

pkg/log/context.go
```diff
@@ -23,46 +23,66 @@ import (
...
 	}
-	regMutex     = sync.Mutex{}
-	once         = sync.Once{}
-	defaultEntry = logrus.NewEntry(logrus.StandardLogger())
+
+	supportedOutputs = map[string]io.Writer{
+		os.Stdout.Name(): os.Stdout,
+		os.Stdin.Name():  os.Stdin,
+		os.Stderr.Name(): os.Stderr,
+		"ginkgowriter":   ginkgo.GinkgoWriter,
```
dpanayotov: I'd rather not have test dependencies in productive code and leave the configuration as is.

pkg/log/context.go
```diff
@@ -23,46 +23,66 @@ import (
...
+
 	supportedFormatters = map[string]logrus.Formatter{
 		"json": &logrus.JSONFormatter{},
 		"text": &logrus.TextFormatter{},
 	}
-	regMutex     = sync.Mutex{}
-	once         = sync.Once{}
-	defaultEntry = logrus.NewEntry(logrus.StandardLogger())
+
+	supportedOutputs = map[string]io.Writer{
```
dpanayotov: No way to add other outputs. What if we want to log to a file?

pkg/log/context.go
```diff
@@ -23,46 +23,66 @@ import (
...
+		os.Stdout.Name(): os.Stdout,
+		os.Stdin.Name():  os.Stdin,
```
dpanayotov: won't writing to stdin throw an error?

pkg/log/context.go
```diff
@@ -74,40 +94,70 @@ func (s *Settings) Validate() error {
...
-			Level:     level,
-			Out:       settings.Output,
-			Hooks:     make(logrus.LevelHooks),
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	currentSettings = settings
+	level, err := logrus.ParseLevel(settings.Level)
+	if err != nil {
+		panic(fmt.Sprintf("Could not parse log level configuration: %s", err))
```
dpanayotov: As this is now called at runtime, I think panicking should be replaced with returning an error. Otherwise you'll get 500 when providing `level: dummy` instead of 400.

pkg/log/context.go
```diff
@@ -74,40 +94,70 @@ func (s *Settings) Validate() error {
...
-		hook := filename.NewHook()
-		hook.Field = FieldComponentName
-		logger.AddHook(hook)
-		defaultEntry = logrus.NewEntry(logger)
-		defaultEntry = defaultEntry.WithField(FieldCorrelationID, "-")
-	})
-	return ContextWithLogger(ctx, defaultEntry)
+	}
+	e := entry.(*logrus.Entry)
+	e.Logger = &logrus.Logger{
```
dpanayotov: I remember that this caused a race condition in the past, because you need the same mutex from the entry's logger, but it's private

storage/notification_cleaner.go
```diff
@@ -50,7 +50,7 @@ func (nc *NotificationCleaner) Start(ctx context.Context, group *sync.WaitGroup)
 			group.Done()
 		}()
 		cleanInterval := nc.Settings.Notification.CleanInterval
-		log.C(ctx).Infof("Scheduling notification cleaning every %s", cleanInterval.String())
+		log.D().Infof("Scheduling notification cleaning every %s", cleanInterval.String())
```
dpanayotov: Why change to default?

test/logging_test/logging_test.go
```diff
@@ -0,0 +1,74 @@
...
+			ctx.SMWithOAuth.GET(web.LoggingConfigURL).
+				Expect().
+				Status(http.StatusOK).JSON().Object().ContainsMap(map[string]string{
+				"level":  "error",
+				"format": "json",
+			})
+		})
+	})
+
+	Context("PUT", func() {
```
dpanayotov: Add negative tests with invalid log format and level

pkg/log/context.go
```diff
@@ -23,46 +23,66 @@ import (
...
+}
 
 // Settings type to be loaded from the environment
 type Settings struct {
-	Level  string    `description:"minimum level for log messages"`
-	Format string    `description:"format of log messages. Allowed values - text, json"`
-	Output io.Writer `mapstructure:"-"`
+	Level  string `description:"minimum level for log messages" json:"level,omitempty"`
+	Format string `description:"format of log messages. Allowed values - text, json" json:"format,omitempty"`
+	Output string `description:"output for the logs. Allowed values - /dev/stdout, /dev/stdin, /dev/stderr, ginkgowriter" json:"output,omitempty"`
```
dpanayotov: This way you can change the output at runtime which may cause more bad than good IMO.

pkg/log/context_test.go
```diff
@@ -64,6 +63,7 @@ var _ = Describe("log", func() {
 				expectPanic(&Settings{
 					Level:  "invalid",
 					Format: "text",
+					Output: os.Stderr.Name(),
```
dzahariev: When log level is not recognised, we better fallback to working (safe) log level (like "error") instead crash.

api/configuration/logging_controller.go
```diff
@@ -0,0 +1,56 @@
...
+	ctx := r.Context()
+	logCfg := log.Configuration()
+	log.C(ctx).Debugf("Obtaining log configuration with level: %s and format: %s", logCfg.Format, logCfg.Level)
+
+	return util.NewJSONResponse(http.StatusOK, logCfg)
+}
+
+func (c *Controller) setLoggingConfiguration(r *web.Request) (*web.Response, error) {
+	ctx := r.Context()
+	loggingConfig := log.Configuration()
```
dpanayotov: This config should be validated before passing it to the `Configure` func

test/logging_test/logging_test.go
dpanayotov: Add test that shows configuring the output is forbidden (at least it should be IMO).

pkg/log/context.go
KirilKabakchiev: we need to change the output for thr ginkgo tests so that it does not spam logs for successfull tests - the old way doesnt work anymore, so we needed a way to switch the writer. I am open to suggestions

pkg/log/context.go
KirilKabakchiev: not changing the logger causes a race condition currently

api/configuration/logging_controller.go
KirilKabakchiev: BytesToObject does that - welcome to the hidden validation "lair"

pkg/log/context.go
KirilKabakchiev: check the previous comment

pkg/log/context.go
KirilKabakchiev: seems to work without copying the logger after I changed one of the tests that failed with race, so we can keep the same logger for all entries

pkg/log/context_test.go
KirilKabakchiev: we wont crash - it will return 500 - but returning 400 will be better

pkg/env/env.go
KirilKabakchiev: normally i would put it there but then the proxies will have to manually add this handler also and I'd prefer for it to be transparent (always part of the env)

pkg/log/context.go
KirilKabakchiev: no idea - i will remove it anyway

pkg/log/context.go
KirilKabakchiev: I could add logic for this  but whats the point of we cant change the log output from the API (your other comments)

pkg/log/context.go
```diff
@@ -23,92 +23,145 @@ import (
...
+
+	entry := ctx.Value(logKey{})
+	if entry == nil {
+		entry = defaultEntry
+	} else {
+		defaultEntry.Logger.SetOutput(output)
+		defaultEntry.Logger.SetFormatter(formatter)
+		defaultEntry.Logger.SetLevel(level)
+	}
+	e := copyEntry(entry.(*logrus.Entry))
```
dpanayotov: This will change the log level of the configuration request. Do we want that? e.g. you log on error before the controller and log on debug after it.

pkg/env/env.go
```diff
@@ -143,18 +146,50 @@ func (v *ViperEnv) setupConfigFile() error {
...
+			if strings.Contains(event.String(), "WRITE") || strings.Contains(event.String(), "CREATE") {
+				logLevel := env.Get("log.level").(string)
+				logFormat := env.Get("log.format").(string)
+
+				log.C(ctx).Warnf("Reconfiguring logrus logging using level %s and format %s", logLevel, logFormat)
+				var err error
+				ctx, err = log.Configure(ctx, &log.Settings{
+					Level:  logLevel,
+					Format: logFormat,
+					Output: os.Stdout.Name(),
```
dpanayotov: if output change is forbidden at runtime this may override the default one

pkg/log/context.go
```diff
@@ -23,92 +23,145 @@ import (
...
-		logger.AddHook(hook)
-		defaultEntry = logrus.NewEntry(logger)
-		defaultEntry = defaultEntry.WithField(FieldCorrelationID, "-")
-	})
-	return ContextWithLogger(ctx, defaultEntry)
+func Configure(ctx context.Context, settings *Settings) (context.Context, error) {
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	currentSettings = settings
```
dpanayotov: if the provided settings are invalid, then until they're reconfigured you'll be receiving invalid settings on getConfig. This is only valid when modifying the file, but still.

pkg/env/env.go
```diff
@@ -143,18 +146,50 @@ func (v *ViperEnv) setupConfigFile() error {
...
+				})
+				if err != nil {
+					log.C(ctx).Errorf("Could not set log level to %s and log format to %s after config file modification event of type %s: %s", logLevel, logFormat, event.String(), err)
```
dpanayotov: `log.WithError`

pkg/log/context.go
KirilKabakchiev: we could add another validation of the settings 
`	if err := settings.Validate(); err != nil {
		return nil, err
	}
`

at the start of log.Configure

pkg/env/env.go
KirilKabakchiev: true - a more general question would be:

Does it make sense to allow log output switching via the apploication.yml file ? If you have access to the file system you could switch the output to a file (if we add support for files as output) which might or might not be helpful?

Or we can just forbid changing the output from the application.yml as well

dpanayotov: Unless we add the possibility to log to a file I don't think we need to support changing the output

pkg/log/context.go
KirilKabakchiev: I am not sure - probably it would be easier to understand if the log level for the current request is kept the same - we could do `	_, err = log.Configure(ctx, &loggingConfig)
` and skip assigning the new ctx in the log controller

pkg/log/context.go
```diff
@@ -119,7 +119,10 @@ func Configure(ctx context.Context, settings *Settings) (context.Context, error)
 	mutex.Lock()
 	defer mutex.Unlock()
 
-	currentSettings = settings
+	if err := settings.Validate(); err != nil {
+		return nil, err
+	}
+
 	level, err := logrus.ParseLevel(settings.Level)
```
dpanayotov: these 3 checks are performed in `Validate`

PR #325, opened 2019-08-27
Platforms connection in health-check

DimitarPetrov:
## Motivation

Connection with platforms should be included in the health-check so it is easy to figure out if any state is propagated to a platform.

## Approach
In the configuration under health you could specify platforms of which type to be included in the health-check. In the indicators section you could specify different properties for a given platform type.
It looks like this:
```
...
health:
  platform_types:
    - cloudfoundry
    - kubernetes
  indicators:
    storage:
      fatal: true
      failures_threshold: 3
      interval: 30s
    cloudfoundry_platforms:
      fatal: true
      failures_threshold: 3
      interval: 30s
    kubernetes_platforms:
      fatal: false
      interval: 30s
...
```
And health-check response for such a configuration looks like this:
```
{
  "details": {
    "cloudfoundry_platforms": {
      "check_time": "2019-08-27T13:25:29.391051+03:00",
      "details": {},
      "fatal": true,
      "first_failure_at": "0001-01-01T00:00:00Z",
      "name": "cloudfoundry_platforms",
      "num_failures": 0,
      "status": "UP"
    },
    "kubernetes_platforms": {
      "check_time": "2019-08-27T13:25:29.391177+03:00",
      "details": {
        "k8s": {
          "details": {
            "since": "0001-01-01T00:00:00Z"
          },
          "status": "DOWN"
        }
      },
      "error": "there is inactive kubernetes platforms",
      "first_failure_at": "2019-08-27T13:24:59.388415+03:00",
      "name": "kubernetes_platforms",
      "num_failures": 2,
      "status": "DOWN"
    },
    "storage": {
      "check_time": "2019-08-27T13:25:29.388445+03:00",
      "fatal": true,
      "first_failure_at": "0001-01-01T00:00:00Z",
      "name": "storage",
      "num_failures": 0,
      "status": "UP"
    }
  },
  "status": "UP"
}
```


pkg/health/types.go
```diff
@@ -26,16 +26,39 @@ import (
...
+
+// indicatorNames is a list of names of indicators which will be registered with default settings
+// as part of default health settings, this will allow binding them as part of environment.
+// If an indicator is registered but not specified in this list, it will be configured with
+// default settings again, but this defaults could be overridden only via application.yml,
+// env variables and pflags won't have any effect. If an indicator is specified in this list
+// but later not registered nothing will happen.
+var indicatorNames = [...]string{
+	StorageIndicatorName,
+	"kubernetes" + PlatformIndicatorSuffix,
```
dpanayotov: SM must not care about different platform types in the code. These should be a configuration

KirilKabakchiev: If we don't mention the indicator names here somwhow then they cannot be configured via the configuration

dotchev:
Describe in the _Approach_ section the meaning of `failures_threshold` and `interval` properties

dotchev:
Any documentation for this feature?

DimitarPetrov:
> Describe in the _Approach_ section the meaning of `failures_threshold` and `interval` properties

This properties are part of different change which introduces async healthcheck.
This is the PR https://github.com/Peripli/service-manager/pull/301 which introduces them.

pkg/health/types.go
```diff
@@ -26,16 +26,39 @@ import (
...
+// PlatformIndicatorSuffix is the suffix for a platform type indicator
+const PlatformIndicatorSuffix = "_platforms"
+
+// indicatorNames is a list of names of indicators which will be registered with default settings
+// as part of default health settings, this will allow binding them as part of environment.
+// If an indicator is registered but not specified in this list, it will be configured with
+// default settings again, but this defaults could be overridden only via application.yml,
+// env variables and pflags won't have any effect. If an indicator is specified in this list
+// but later not registered nothing will happen.
+var indicatorNames = [...]string{
```
dotchev: what about indicators outside Peripli? who will add them here?

storage/postgres/migrations/20190829101500_add_platforms_status.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE platforms ADD COLUMN active boolean NOT NULL DEFAULT '0';
+ALTER TABLE platforms ADD COLUMN last_active TIMESTAMP NOT NULL DEFAULT '1970-01-01 00:00:00+00';
```
dotchev: that's strange, better use NULL is the platform was never active

storage/postgres/notificator.go
```diff
@@ -456,17 +465,30 @@ func (c *consumers) Delete(queue storage.NotificationQueue) {
 		for index, platform := range c.platforms {
 			if platform.ID == platformIDToDelete {
 				c.platforms = append(c.platforms[:index], c.platforms[index+1:]...)
+				if err := c.updatePlatform(platform.ID, func(p *types.Platform) {
+					p.Active = false
+					p.LastActive = time.Now()
+				}); err != nil {
```
dotchev: this `if` is hard to read, break into multiple statements

api/healthcheck/healthcheck_controller.go
```diff
@@ -70,6 +72,9 @@ func (c *controller) aggregate(overallState map[string]h.State) *health.Health {
 	details := make(map[string]interface{})
 	for name, state := range overallState {
 		state.Status = convertStatus(state.Status)
```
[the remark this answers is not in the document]
DimitarPetrov: If we decide to adopt library’s status “ok” and “failed” this change should be propagated everywhere since all the components check UP to determine if it is started correctly.

pkg/health/types.go
DimitarPetrov: They will maintain separate list of indicator names and extend default settings with them while configuring their settings. Other approach is to export this array lets discuss which is better.

storage/postgres/migrations/20190829101500_add_platforms_status.up.sql
DimitarPetrov: If we use NULL then in Platform struct we need to use sql.NullString instead of time.Time so rows scanning can be done properly for null values and deal with casting, which for me seems like unnecessary complication but i am okay with both solutions.

PR #328, opened 2019-09-13
Unify response for list operations

DimitarPetrov:
## Motivation
All List operations must return the same response scheme

## Approach
```
{
    "num_items": <the number of items returned>,
    "items": [
         { ... }
    ]
}
```

### DIFF
--- a/api/base_controller.go
+++ b/api/base_controller.go
@@ -193,12 +193,18 @@ func (c *BaseController) ListObjects(r *web.Request) (*web.Response, error) {
 		return nil, util.HandleStorageError(err, string(c.objectType))
 	}
 
-	for i := 0; i < objectList.Len(); i++ {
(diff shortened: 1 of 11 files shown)


api/base_controller.go
```diff
@@ -193,12 +193,18 @@ func (c *BaseController) ListObjects(r *web.Request) (*web.Response, error) {
 		return nil, util.HandleStorageError(err, string(c.objectType))
 	}
 
+	page := types.ObjectPage{
+		Items: make([]types.Object, 0, objectList.Len()),
+	}
+
 	for i := 0; i < objectList.Len(); i++ {
+		page.ItemsCount++
```
KirilKabakchiev: another approach could be to just change the json tag on the objectlist generator to `json:"items"` and add an Items() method to the ObjectList interface and the generator

PR #329, opened 2019-09-13
Support max_items query parameter

DimitarPetrov:
## Motivation
SM list API must handle the max_items query parameter to determine the maximum number of items to return.
This max_items parameter must not be greater than an internally configured value.

## Approach
Parameter max_items introduces which means maximum number of items to return in the response.If the client sets max_items to 0, it is assumed that the client is only interested in the total number of items.

### DIFF
--- a/api/filters/paging.go
+++ b/api/filters/paging.go
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2018 The Service Manager Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package filters
+
+import (
+	"context"
+	"fmt"
+	"github.com/Peripli/service-manager/pkg/query"
+	"github.com/Peripli/service-manager/pkg/util"
+	"github.com/Peripli/service-manager/pkg/web"
+	"net/http"
+	"strconv"
+)
+
+// PagingFilterName is the name of the paging filter
+const PagingFilterName = "PagingFilter"
+
+// pagingFilter is filter that adds paging criteria to request
+type pagingFilter struct {
+	DefaultPageSize int
+	MaxPageSize     int
+}
+
+// NewPagingFilter returns new paging filter for given default and max page sizes
+func NewPagingFilter(defaultPageSize, maxPageSize int) web.Filter {
+	return &pagingFilter{
+		DefaultPageSize: defaultPageSize,
+		MaxPageSize:     maxPageSize,
+	}
+}
+
+// Name returns the name of the paging filter
+func (*pagingFilter) Name() string {
+	return PagingFilterName
+}
+
+// Run represents the paging middleware function that processes the request and configures needed criteria.
+func (pf *pagingFilter) Run(req *web.Request, next web.Handler) (*web.Response, error) {
+	ctx := req.Context()
+	maxItems := req.URL.Query().Get("max_items")
+	limit, err := pf.validateMaxItemsQuery(maxItems)
+	if err != nil {
+		return nil, err
+	}
+	ctx = context.WithValue(ctx, "limit", limit)
+	if limit > 0 {
+		ctx, err = query.AddCriteria(ctx, query.LimitResultBy(limit))
+		if err != nil {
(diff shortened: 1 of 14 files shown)


api/filters/paging.go
```diff
@@ -0,0 +1,130 @@
...
+func (pf *pagingFilter) Run(req *web.Request, next web.Handler) (*web.Response, error) {
+	ctx := req.Context()
+	maxItems := req.URL.Query().Get("max_items")
+	limit, err := pf.validateMaxItemsQuery(maxItems)
+	if err != nil {
+		return nil, err
+	}
+	ctx = context.WithValue(ctx, "limit", limit)
+	if limit > 0 {
+		ctx, err = query.AddCriteria(ctx, query.LimitResultBy(limit))
```
KirilKabakchiev: Currently there are two layers in the implementation that i think miss proper validation and can kind of mess up the limitting

1. If another filter adds a limit criteria after this one, this one will be silently overriden 
2. in the pgQuery builder (postgres implementation) there is a method of adding limit that currently override the limit everytime limit is set by it.
3. Order by created_at is currently hardcoded in an ugly place in the code, maybe now as it will be required for paging we can refactor it slightly

storage/postgres/query_builder.go
```diff
@@ -103,6 +103,40 @@ func (pgq *pgQuery) List(ctx context.Context, entity PostgresEntity) (*sqlx.Rows
...
+	if pgq.err != nil {
+		return 0, pgq.err
+	}
+
+	tableName := entity.TableName()
+	labelsEntity := entity.LabelEntity()
+
+	baseQuery := fmt.Sprintf("SELECT COUNT(*) FROM %s %s", tableName, mainTableAlias)
+
+	if labelsEntity != nil {
```
KirilKabakchiev: Do we care about the labels when counting? Probably we do for visibilities as its often the case for a vis to have 10k labels - which brings another point - a default page size / max page size of 50 is suitable for most entities but probably not for lets say visibilities - so we might need a more sophisticated logic per entity for these sizes

storage/postgres/query_builder.go
```diff
@@ -103,6 +103,40 @@ func (pgq *pgQuery) List(ctx context.Context, entity PostgresEntity) (*sqlx.Rows
...
+	pgq.sql.WriteString(baseQuery)
+
+	pgq.orderByFields = nil
+
+	if err := pgq.finalizeSQL(ctx, entity, false); err != nil {
+		return 0, err
+	}
+
+	count := make([]int, 0, 1)
+	err := pgq.db.SelectContext(ctx, &count, pgq.sql.String(), pgq.queryParams...)
```
KirilKabakchiev: You could probably use GetContext and pass a single int to scan in instead of an array

test/list.go
```diff
@@ -423,6 +420,36 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase) bool {
...
+					It("returns 400", func() {
+						ctx.SMWithOAuth.GET(t.API).WithQuery("max_items", "invalid").Expect().Status(http.StatusBadRequest)
+					})
+				})
+				Context("with zero max items query", func() {
+					It("returns count of the items only", func() {
+						resp := ctx.SMWithOAuth.GET(t.API).WithQuery("max_items", 0).Expect().Status(http.StatusOK).JSON()
+
+						resp.Object().NotContainsKey("items")
+						resp.Path("$.num_items").Number().Gt(0)
```
KirilKabakchiev: Verify that the actual items are not in the reaponse and also that num items are the expected nimber of items

test/notification_test/notification_test.go
```diff
@@ -175,11 +175,11 @@ var _ = Describe("Notifications Suite", func() {
 				object := ctx.SMWithOAuth.GET("/v1/service_offerings").WithQuery("fieldQuery", "broker_id = "+id).
 					Expect()
 
-				so := object.Status(http.StatusOK).JSON().Object().Value("service_offerings").Array().First()
+				so := object.Status(http.StatusOK).JSON().Object().Value("items").Array().First()
 
 				servicePlanID := ctx.SMWithOAuth.GET("/v1/service_plans").WithQuery("fieldQuery", fmt.Sprintf("service_offering_id = %s", so.Object().Value("id").String().Raw())).
 					Expect().
-					Status(http.StatusOK).JSON().Object().Value("service_plans").Array().First().Object().Value("id").String().Raw()
+					Status(http.StatusOK).JSON().Object().Value("items").Array().First().Object().Value("id").String().Raw()
```
KirilKabakchiev: This is a more genera comment - everywhere (in every test we do list) it will currently return a single page. However normally in those places we want all items. Right now all tests do work because the items that are created are leas than the default page size which is not reliable in general - so i think we will need a common utility used across all tests when talking about listing stuff that aggregates all pages

DimitarPetrov: This will be part of the token change since now we cannot iterate over the pages because the functionality is missing. With the PR that introduces the token and the real paging logic the tests must aggregate all the pages on each list call.

PR #330, opened 2019-09-16
Paging for List operations

DimitarPetrov:
## Motivation

Paging of all list call responses should be implemented. This is a topic branch for all the related tasks.

## Approach

Implementation of this part of the specification: https://github.com/Peripli/specification/blob/sm-ops/api.md#paging-parameters


docs/development/paging.md
```diff
@@ -0,0 +1,32 @@
...
+    {
+      "id": "a62b83e8-1604-427d-b079-200ae9247b60",
+      ...
+    },
+    ...
+  ]
+}
+```
+
+The `num_items` value MAY NOT be accurate the next time the client retrieves the result set or the next page in the result set.
```
NickyMateev:
```suggestion
The `num_items` value **MAY NOT** be accurate the next time the client retrieves the result set or the next page in the result set.
```

pkg/types/interfaces.go
```diff
@@ -52,3 +53,10 @@ type ObjectList interface {
 	ItemAt(index int) Object
 	Len() int
 }
+
+// ObjectPage is the DTO for a given page of resources when listing
```
NickyMateev: I think it would be useful to add to the doc that "_Token represents the base64 encoded paging_sequence of the last entity in items list_", as I think this is only mentioned in the documentation **.md** page and nowhere in the code.

pkg/util/list.go
```diff
@@ -0,0 +1,108 @@
...
+
+	li.next = responseBody.Token
+	li.done = li.next == ""
+	return !li.done, responseBody.NumItems, nil
+}
+
+// ListAll retrieves all the objects from the given url by loading all the pages
+// items should be a pointer to a slice, that will be populated with all the items
+// doRequest function executes the HTTP request, it is responsible for authentication
+func ListAll(ctx context.Context, doRequest DoRequestFunc, url string, items interface{}) error {
```
NickyMateev: Throughout this paging utility, the `items` parameter is a raw interface value, and later on there's some moderate reflection use. Can't this be typed and use `*[]types.Object` throughout?

PR #331, opened 2019-09-16
Add config api endpoint

KirilKabakchiev:
The API allows calling `/v1/config` and seeing the whole SM configuration as it currently is. It also takes into account any dynamic config changes that may have occurred and any overrides the application may have done to the configuration after bootstrap


pkg/security/filters/required_authn.go
```diff
@@ -58,14 +63,10 @@ func (raf *requiredAuthnFilter) FilterMatchers() []web.FilterMatcher {
 					web.ServicePlansURL+"/**",
 					web.VisibilitiesURL+"/**",
 					web.NotificationsURL+"/**",
+					web.ConfigURL+"/**",
```
NickyMateev: You can remove `web.LoggingConfigURL+"/**"` below as this declaration already covers any subpaths

PR #333, opened 2019-09-18
Support token parameter for paging

DimitarPetrov:
## Motivation

All List endpoints MUST return the list of entities ordered by paging sequence (which is auto-increment field and implies creation order). Clients define the beginning of the result set by setting the token parameter. If this parameter is missing or empty, the first page is returned. To retrieves subsequent pages, the client has to provide the token value that was included in the previous page.
The Service Manager MUST provide an appropriate error if the provided token is invalid. In this case, the client has to reiterate the list from the beginning.

## Approach

1) Each entity has an additional serial column paging_sequence which is encoded in the token and used to determine where next page starts.
2) A new filter is introduced which attaches the criteria for the paging provided by the token and max_items query params.
3) Field-query and limit clauses are now applied only to the primary table (before joining labels) and order by clause is applied to both tables (before and after joining labels). This common logic for sub-querying is extracted and generalised from query builder.
4) Same sub-query from 3) is used for label-querying the joined labels tables. Each label query is modelled as two field-queries
5) Initial request criteria from query parameters provided by the user are now saved as user provided and can be retrieved separately from all criteria which might have additional system criterias (such as the ones added from paging filter).


dotchev:
From the motivation

> ... return the list of entities ordered by creation date and entity ID

Now the items are ordered by paging_sequence.

api/base_controller.go
```diff
@@ -187,34 +189,61 @@ func (c *BaseController) GetSingleObject(r *web.Request) (*web.Response, error)
...
+	userCriteria := query.UserCriteriaForContext(ctx)
+	count, err := c.repository.Count(ctx, c.objectType, userCriteria...)
+	if err != nil {
+		return nil, util.HandleStorageError(err, string(c.objectType))
+	}
+
+	limit := ctx.Value("limit").(int)
+	if limit == 0 {
+		log.C(ctx).Debugf("Returning only count of %s since max_items is 0", c.objectType)
 		page := struct {
```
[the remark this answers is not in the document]
DimitarPetrov: We don't want `items: nil` in the response, however when the num_items is 0 we want `items: []`, so `omitempty` is not an option.

api/filters/paging.go
```diff
@@ -52,22 +52,32 @@ func (*pagingFilter) Name() string {
...
+	rawToken := req.URL.Query().Get("token")
+	token, err := pf.parsePageToken(rawToken)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx = web.ContextWithPageLimit(ctx, limit)
 	if limit > 0 {
-		ctx, err = query.AddCriteria(ctx, query.LimitResultBy(limit))
+		ctx, err = query.AddCriteria(ctx, query.LimitResultBy(limit+1),
```
KirilKabakchiev: LimitResultsTo should be a more appropriate name for the method called

Adding +1 seems counterintuitive to what the method name states. Whats the reason?

api/filters/paging.go
```diff
@@ -52,22 +52,32 @@ func (*pagingFilter) Name() string {
...
+			query.OrderResultBy("paging_sequence", query.AscOrder),
+			query.ByField(query.GreaterThanOperator, "paging_sequence", strconv.Itoa(token)))
```
KirilKabakchiev: Atoi in parse method and Itoa here  probably cancel out - you could directly use the string but we should check how meaningful the error message will be

api/base_controller.go
```diff
@@ -273,3 +302,8 @@ func stripCredentials(ctx context.Context, object types.Object) {
 		log.C(ctx).Debugf("Object of type %s with id %s is not secured, so no credentials are cleaned up on response", object.GetType(), object.GetID())
 	}
 }
+
+func generateTokenForItem(obj types.Object) string {
+	nextPageToken := obj.(types.Pageable).GetPagingSequence()
```
KirilKabakchiev: It looks like you expect all Objects to be Pageable. You could embed Pageable in Object and put the PagingSequence field as well as Pageable implementation in Base to avoid type casting here and code duplication accross all types

api/base_controller.go
```diff
@@ -187,34 +189,61 @@ func (c *BaseController) GetSingleObject(r *web.Request) (*web.Response, error)
...
+
 	for i := 0; i < objectList.Len(); i++ {
-		page.ItemsCount++
 		obj := objectList.ItemAt(i)
 		stripCredentials(ctx, obj)
 		page.Items = append(page.Items, obj)
 	}
-	return util.NewJSONResponse(http.StatusOK, page)
+
+	var token string
```
KirilKabakchiev: token looks like an unnecessary variable?

api/base_controller.go
```diff
@@ -187,34 +189,61 @@ func (c *BaseController) GetSingleObject(r *web.Request) (*web.Response, error)
...
-			return nil, util.HandleStorageError(err, string(c.objectType))
-		}
+	userCriteria := query.UserCriteriaForContext(ctx)
+	count, err := c.repository.Count(ctx, c.objectType, userCriteria...)
+	if err != nil {
+		return nil, util.HandleStorageError(err, string(c.objectType))
+	}
+
+	limit := web.PageLimitFromContext(ctx)
+	if limit == 0 {
```
KirilKabakchiev: I thought limited is always set and if not provided default value is used. Limit being zero does not seem possible. If this check is to handle a case in which the PagingFilter's FilterMatchers does not include an entry for a resource that uses base_controller.go, it won't work as line 198 will panic

pkg/web/context_keys.go
```diff
@@ -32,3 +33,13 @@ func IsAuthorized(ctx context.Context) bool {
...
 }
+
+// ContextWithPageLimit sets the page size for list requests
+func ContextWithPageLimit(ctx context.Context, limit int) context.Context {
+	return context.WithValue(ctx, limitKey, limit)
+}
+
+// PageLimitFromContext retrieves the page size from the context
+func PageLimitFromContext(ctx context.Context) int {
+	return ctx.Value(limitKey).(int)
```
KirilKabakchiev: this would panic if the key is not present

test/list.go
```diff
@@ -335,22 +332,17 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase) bool {
...
 		By(fmt.Sprintf("[TEST]: Verifying expected status code %d is returned from list operation", listOpEntry.expectedStatusCode))
-		resp := req.
-			Expect().
-			Status(listOpEntry.expectedStatusCode)
 
 		if listOpEntry.expectedStatusCode != http.StatusOK {
 			By(fmt.Sprintf("[TEST]: Verifying error and description fields are returned after list operation"))
-
-			resp.JSON().Object().Keys().Contains("error", "description")
+			req := ctx.SMWithOAuth.GET(t.API)
```
KirilKabakchiev: this should probably be List and not Get and should happen before the status code branching, imo. Normally the method that the test tests is independant from the expectations (calling GET if you expect error and list otherwise seems a bit confusing

test/list.go
```diff
@@ -448,6 +442,37 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase) bool {
 						resp.Path("$.num_items").Number().Gt(0)
 					})
 				})
+				When("there is no more pages", func() {
```
KirilKabakchiev: there are no more pages

test/list.go
```diff
@@ -448,6 +442,37 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase) bool {
...
+						ctx.SMWithOAuth.GET(t.API).WithQuery("token", token).Expect().Status(http.StatusNotFound)
+					}
+					Context("no base64 encoded", func() {
+						It("returns 404", func() {
+							executeWithInvalidToken("invalid")
+						})
+					})
+					Context("non numerical", func() {
+						It("returns 404", func() {
+							token := base64.StdEncoding.EncodeToString([]byte("non-numerical"))
```
KirilKabakchiev: what about a negative or zero numerical ?

test/list.go
```diff
@@ -422,12 +414,14 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase) bool {
 
 			Context("Paging", func() {
 				Context("with max items query", func() {
-					It("returns smaller pages", func() {
+					It("returns smaller pages token and Link header", func() {
```
KirilKabakchiev: The way "Paging" tests context is added will not test a combination of fieldqueries + labelQueries + paging. Not sure if its a big deal but if for example we introduce a bug that causes having a labelquery and paging query (token/max_items) to return a wrong result or error, we wont catch that with these tests

api/filters/paging.go
```diff
@@ -52,22 +52,32 @@ func (*pagingFilter) Name() string {
 func (pf *pagingFilter) Run(req *web.Request, next web.Handler) (*web.Response, error) {
 	ctx := req.Context()
 	maxItems := req.URL.Query().Get("max_items")
-	limit, err := pf.validateMaxItemsQuery(maxItems)
+	limit, err := pf.parseMaxItemsQuery(maxItems)
```
KirilKabakchiev: currently there is also an antlr the idea of which is to translate rest queries to criterions - are we going to try to include the paging query parsing there too ? If so we should probably consider this while implementing the paging (for example if we wanted to reuse the antlr stuff, we would probably need a query paramater `pagingQuery=limit=5 and token=abc` and introduce a PagingCriteria type. (there are other approaches as well..)

storage/postgres/abstract.go
```diff
@@ -104,7 +104,7 @@ func columnsByTags(tags []tagType) map[string]bool {
 
 func validateFieldQueryParams(columns map[string]bool, criteria []query.Criterion) error {
 	for _, criterion := range criteria {
-		if criterion.Type == query.FieldQuery && !columns[criterion.LeftOp] {
+		if !columns[criterion.LeftOp] {
```
KirilKabakchiev: method still receives Criterion type so it should handle properly cases when a non field query is passed (for example ignore criterias with different type)

api/filters/paging.go
```diff
@@ -93,6 +103,30 @@ func (pf *pagingFilter) validateMaxItemsQuery(maxItems string) (int, error) {
...
 
+func (pf *pagingFilter) parsePageToken(token string) (int, error) {
+	var targetPageSequence int
+	if token != "" {
+		base64DecodedTokenBytes, err := base64.StdEncoding.DecodeString(token)
+		if err != nil {
+			return 0, &util.HTTPError{
+				ErrorType:   "TokenInvalid",
+				Description: fmt.Sprintf("Invalid token provided: %v", err),
+				StatusCode:  http.StatusNotFound,
```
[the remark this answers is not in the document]
DimitarPetrov: https://github.wdf.sap.corp/SvcManager/e2e-tests/pull/51
Here is defined as 404, maybe propose a spec change?

api/filters/paging.go
```diff
@@ -52,22 +52,32 @@ func (*pagingFilter) Name() string {
...
-	ctx = context.WithValue(ctx, "limit", limit)
+
+	rawToken := req.URL.Query().Get("token")
+	token, err := pf.parsePageToken(rawToken)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx = web.ContextWithPageLimit(ctx, limit)
 	if limit > 0 {
```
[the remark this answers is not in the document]
DimitarPetrov: By the spec, when the limit is 0 the client is interested in total number of items only. That's why we do not add a criteria if it is 0 and just go to the controller when if the limit is 0 only count is returned.

api/filters/paging.go
DimitarPetrov: Because if you set the limit as provided in the request there is no way to say afterwards if there is more items or not (if you need to generate and return a token in the response).

api/base_controller.go
DimitarPetrov: The Notification type is not Pageable.

api/base_controller.go
DimitarPetrov: By the spec limit 0 means that the client is interested in the total count of the items only.

test/list.go
DimitarPetrov: The List method expects the iteration to be successful and returns the array of items.
Here we want to assert the error is in the response which is different case and no need to iterate or make multiple calls. If this checks are moved to List and the status code expected is moved to the List it will not look like List anymore.

storage/postgres/abstract.go
DimitarPetrov: Maybe we need to return the check, as i don't remember a reason for removing it.

api/filters/paging.go
DimitarPetrov: We must clarify which is going first and which needs to be adopted because there will be conflicts for sure.

storage/postgres/migrations/20190913161000_paging.up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+ALTER TABLE brokers ADD COLUMN paging_sequence SERIAL;
+ALTER TABLE platforms ADD COLUMN paging_sequence SERIAL;
+ALTER TABLE visibilities ADD COLUMN paging_sequence SERIAL;
```
georgifarashev: While serial (4 bytes) is a big number over time it might not be sufficient, especially with visibilities. Wouldn't it be better to use bigserial (8 bytes)?
Also all variables "PagingSequence" should be int64.

dpanayotov:
General comment - this needs a page in the docs directory explaining how the paging works (implementation-specific token handling)

api/base_controller.go
```diff
@@ -187,34 +189,61 @@ func (c *BaseController) GetSingleObject(r *web.Request) (*web.Response, error)
...
 func (c *BaseController) ListObjects(r *web.Request) (*web.Response, error) {
 	ctx := r.Context()
-	if ctx.Value("limit").(int) == 0 {
-		log.C(ctx).Debugf("Getting count of %s since max_items is 0", c.objectType)
-		count, err := c.repository.Count(ctx, c.objectType, query.CriteriaForContext(ctx)...)
-		if err != nil {
-			return nil, util.HandleStorageError(err, string(c.objectType))
-		}
+	userCriteria := query.UserCriteriaForContext(ctx)
+	count, err := c.repository.Count(ctx, c.objectType, userCriteria...)
```
dpanayotov: By the time the objects are listed, the count may have changed. Describe this as a limitation in the document

api/filters/paging.go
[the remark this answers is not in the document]
dpanayotov: Bruteforcing SM to guess an internal representation is a security issue. It should be 404. Also, change the message from "Invalid token" to "No such token" and log a message for ourselves to know that it was actually invalid.

api/filters/paging.go
```diff
@@ -93,6 +103,30 @@ func (pf *pagingFilter) validateMaxItemsQuery(maxItems string) (int, error) {
...
+				Description: fmt.Sprintf("Invalid token provided: %v", err),
+				StatusCode:  http.StatusNotFound,
+			}
+		}
+		base64DecodedToken := string(base64DecodedTokenBytes)
+		targetPageSequence, err = strconv.Atoi(base64DecodedToken)
+		if err != nil {
+			return 0, &util.HTTPError{
+				ErrorType:   "TokenInvalid",
+				Description: fmt.Sprintf("Invalid token provided: %v", err),
```
dpanayotov: same about the message

storage/postgres/query_builder.go
```diff
@@ -44,88 +43,95 @@ func NewQueryBuilder(db pgDB) *QueryBuilder {
 }
 
 // NewQuery constructs new queries for the current query builder db
-func (qb *QueryBuilder) NewQuery() *pgQuery {
+func (qb *QueryBuilder) NewQuery(entity PostgresEntity) *pgQuery {
```
dpanayotov: General Comment: This whole query building looks like it's getting quite complicated for us to support. How about we discuss if we could benefit from researching some library for queries or drop the abstraction layer altogether and directly use a sql-building library.

pkg/types/platform.go
```diff
@@ -45,6 +46,10 @@ func (e *Platform) GetCredentials() *Credentials {
 	return e.Credentials
 }
 
+func (e *Platform) GetPagingSequence() int {
```
dpanayotov: This looks like it could be generated but is not because of the notifications. Can't they be `Pageable` also and this can go in the base resource with this being generated. `ListObjects` is not implemented for `Notification` so it won't be called anyways

api/base_controller.go
KirilKabakchiev: Even if it implemented Pageable it wouldnt matter as GET notifications does not return a direct response (its a WS connection open call). If we had a normal GET API for notifications they would be pageble also so i think it would be fine if they implemented it

api/filters/paging.go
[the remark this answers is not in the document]
KirilKabakchiev: I don't insist on Bad Request but 404 is not correct. This a continuation token, not a security token.

```
{
kind: "Status",
apiVersion: "v1",
metadata: { },
status: "Failure",
message: "invalid continue token: continue key is not valid: invalid character '\x00' looking for beginning of value",
reason: "BadRequest",
code: 400
}
```

storage/postgres/query_builder_test.go
```diff
@@ -245,9 +243,9 @@ LIMIT 10;`)))
 				criteria5 := query.ByLabel(query.InOperator, "left2", "right2", "right3")
 				criteria6 := query.ByLabel(query.NotEqualsOperator, "left3", "right4")
 
-				_, err := qb.NewQuery().
+				_, err := qb.NewQuery(entity).
```
KirilKabakchiev: The idea of the multiple criteria test is to document how a query with all possible criterias will look like. Add limit and order as well

storage/postgres/query_builder.go
NickyMateev: Something like this?
https://github.com/Masterminds/squirrel

storage/postgres/subquery.go
```diff
@@ -0,0 +1,140 @@
...
+	fieldCriteria []query.Criterion
+	orderByFields []orderRule
+	limit         string
+
+	whereClauseTreeCreator func([]query.Criterion) *whereClauseTree
+
+	err error
+}
+
+type selectSubQuery struct {
```
KirilKabakchiev: whats the point of having subQuery and selectSubQuery ? subQuery is used only in select

api/filters/paging.go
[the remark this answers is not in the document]
DimitarPetrov: If the request has reached this filter, then we assume the request is authenticated and there will be no security risks. So returning Bad Request should be ok?

test/list.go
DimitarPetrov: The original tests are still using paging and the result is aggregated by the List function, so if something is broken other test Entries will cover it, not the tests in the Paging context.

pkg/types/interfaces.go
```diff
@@ -31,8 +31,14 @@ type Secured interface {
 	GetCredentials() *Credentials
 }
 
+// Pageable interface indicates that an object can be listed page by page
+type Pageable interface {
```
dpanayotov: Not needed

api/base_controller.go
```diff
@@ -273,3 +310,84 @@ func stripCredentials(ctx context.Context, object types.Object) {
...
+	return limit, nil
+}
+
+func (c *BaseController) parsePageToken(token string) (string, error) {
+	targetPageSequence := "0"
+	if token != "" {
+		base64DecodedTokenBytes, err := base64.StdEncoding.DecodeString(token)
+		if err != nil {
+			return "", &util.HTTPError{
+				ErrorType:   "TokenInvalid",
```
dpanayotov: `TokenMissing` with log. Even though 404 is returned you're providing internal information in the body.

api/base_controller.go
```diff
@@ -273,3 +310,84 @@ func stripCredentials(ctx context.Context, object types.Object) {
...
+				Description: fmt.Sprintf("Invalid token provided: %v", err),
```
dpanayotov: Log the error and don't return it to the response.

PR #334, opened 2019-09-19
[WIP] paging client

dotchev:
## Motivation

Common client utility for iterating through pages

## Approach

Two usages are supported:
* iterator - load one page at a time, do not keep all the items in memory
* load all items - load all the items
 
## Pull Request status
* [x] Initial implementation
* [ ] Unit tests

### DIFF
--- a/pkg/util/list.go
+++ b/pkg/util/list.go
@@ -22,9 +22,9 @@ type ListIterator struct {
 }
 
 type listResponse struct {
-	Token    string          `json:"token"`
-	NumItems int64           `json:"num_items"`
-	Items    json.RawMessage `json:"items"`
+	Token    string      `json:"token"`
+	NumItems int64       `json:"num_items"`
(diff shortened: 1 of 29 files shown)


pkg/util/list.go
```diff
@@ -0,0 +1,106 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"reflect"
+	"strconv"
+)
+
+// ListIterator lists the objects from the given url by loading one page at a time
+type ListIterator struct {
```
DimitarPetrov: User provided query parameters currently seems to not being supported

dotchev: query parameters can be passed in the url
```go
u, err := url.Parse(baseurl)
q := u.Query()
q.Add("key", "value")
u.RawQuery = q.Encode()
fullurl := u.String()
```

dotchev:
replaced by #335

PR #335, opened 2019-09-25
Paging client

dotchev:
## Motivation

Common client utility for iterating through pages

## Approach

Two usages are supported:
* iterator - load one page at a time, do not keep all the items in memory
* load all items - load all the items in memory by iterating through all the pages
 
## Pull Request status
* [x] Initial implementation
* [x] Unit tests

### DIFF
--- a/pkg/util/list.go
+++ b/pkg/util/list.go
@@ -40,7 +40,7 @@ func (li *ListIterator) Next(ctx context.Context, items interface{}, maxItems in
 	}
 
 	if li.done {
-		return false, -1, errors.New("Iteration already complete")
+		return false, -1, errors.New("iteration already complete")
 	}
 
 	params := map[string]string{}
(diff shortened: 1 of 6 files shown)


pkg/util/list.go
```diff
@@ -0,0 +1,108 @@
...
+	Items    interface{} `json:"items"`
+}
+
+// Next loads the next page of items
+// items should be a pointer to a slice, that will be populated with the items from the current page,
+// if nil, only the total number is returned
+// maxItems is the maximum numbe of items to load with the next page, -1 - use server default, 0 - just get the count
+// more is true if there are more items
+// count is the total number of items, -1 if not available
+func (li *ListIterator) Next(ctx context.Context, items interface{}, maxItems int) (more bool, count int64, err error) {
```
KirilKabakchiev: using raw interfaces is against the current design idea for SM. Also, Go best practices advice to accept interfaces and we have designed the Object interface for this purpose. Can we try to accept ObjectList or []Object here and remove all the reflection ?

PR #336, opened 2019-09-30
Trim down interceptable repository code

NickyMateev:
# Trim down interceptable repository code

## Motivation

The implementation of _InterceptableTransactionalRepository_ is similar to _interceptableRepository_, but it still differed in some subtle ways, like introducing intermediary structs like * finalCreateObjectInterceptor*. This PR eliminates the intermediary structs and tries to make both repository implementations more similar.

Also added some docs to some package and public structs as I thought it would be beneficial for later on.

## Approach

Refactor _InterceptableTransactionalRepository_ to be similar to _interceptableRepository_ by removing unnecessary code.

## Pull Request status

For example:

* [x] Initial implementation
* [x] Refactoring

### DIFF
--- a/storage/interceptable_repository.go
+++ b/storage/interceptable_repository.go
@@ -469,17 +469,18 @@ func (itr *InterceptableTransactionalRepository) validateDeleteProviders(objectT
 }
 
 func validateProviderOrder(order InterceptorOrder, existingProviderNames []string, providerName string) {
-	positionAroundTx, aroundTxName := order.AroundTxPosition.PositionType, order.AroundTxPosition.Name
-	positionTx, nameTx := order.OnTxPosition.PositionType, order.OnTxPosition.Name
 	if providerWithNameExists(existingProviderNames, providerName) {
 		log.D().Panicf("%s create interceptor provider is already registered", providerName)
 	}
 
+	positionAroundTx, aroundTxName := order.AroundTxPosition.PositionType, order.AroundTxPosition.Name
 	if positionAroundTx != PositionNone {
 		if !providerWithNameExists(existingProviderNames, aroundTxName) {
 			log.D().Panicf("could not find interceptor with name %s", aroundTxName)
 		}
 	}
+
+	positionTx, nameTx := order.OnTxPosition.PositionType, order.OnTxPosition.Name
 	if positionTx != PositionNone {
 		if !providerWithNameExists(existingProviderNames, nameTx) {
 			log.D().Panicf("could not find interceptor with name %s", nameTx)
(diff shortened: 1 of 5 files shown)


storage/interceptable_repository.go
```diff
@@ -50,6 +50,7 @@ func newInterceptableRepository(repository Repository,
 	}
 }
 
+// InterceptableTransactionalRepository wraps a TransactionalRepository and holds sets of ordered interceptor providers for each object type
```
KirilKabakchiev: holds sets of ordered interceptors why ? if you add documentation make sure it brings value to whoever reads it

storage/interceptable_repository.go
```diff
@@ -372,58 +327,34 @@ func (itr *InterceptableTransactionalRepository) Delete(ctx context.Context, obj
...
-	var result types.Object
+	finalInterceptor := func(ctx context.Context, obj types.Object, labelChanges ...*query.LabelChange) (types.Object, error) {
+		var result types.Object
+		var err error
 
-	if err = final.repository.InTransaction(ctx, func(ctx context.Context, txStorage Repository) error {
-		interceptableRepository := newInterceptableRepository(txStorage, final.providedCreateInterceptors, final.providedUpdateInterceptors, final.providedDeleteInterceptors)
+		if err = itr.smStorageRepository.InTransaction(ctx, func(ctx context.Context, txStorage Repository) error {
+			interceptableRepository := newInterceptableRepository(txStorage, providedCreateInterceptors, providedUpdateInterceptors, providedDeleteInterceptors)
+			result, err = interceptableRepository.Update(ctx, obj, labelChanges)
```
KirilKabakchiev: I think that the criteria from the closure should also be passed here (we should double check)

storage/interceptable_repository.go
NickyMateev: In hindsight, I see that this doc doesn't seem to bring any additional value than what the code itself says.
Do you have any better ideas or should I remove the doc?

storage/interceptable_repository.go
NickyMateev: It seems that the criteria has been forgotten way back and hasn't been passed on from the controller level as well - https://github.com/Peripli/service-manager/blob/7920247e060da10cb2f8ea7dced2f99823ae9c01/api/base_controller.go#L296 

Even if we pass it everywhere it is currently not in use on the lowest level (Storage implementation) - https://github.com/Peripli/service-manager/blob/7920247e060da10cb2f8ea7dced2f99823ae9c01/storage/postgres/storage.go#L258

I think we can start passing it everywhere where it's missing and on the lowest level (Storage implementation) mark it as unused by replacing the var name with underscore.

NickyMateev: Another idea would be to discard the criteria argument from the interface all together.

PR #337, opened 2019-10-07
Introduce Access Level

KirilKabakchiev:
Add AccessLevel to user context
Authorizers now return accesslevel

### DIFF
--- a/api/filters/basic_authentication.go
+++ b/api/filters/basic_authentication.go
@@ -85,7 +85,7 @@ func (a *BasicAuthenticator) Authenticate(request *http.Request) (*web.UserConte
 		},
 		AuthenticationType: web.Basic,
 		Name:               username,
-		Attributes:         make(map[string]string),
+		AccessLevel:        web.GlobalAccess,
 	}, httpsec.Allow, nil
 }
 
--- a/api/filters/basic_authentication.go
+++ b/api/filters/basic_authentication.go
@@ -85,7 +85,7 @@ func (a *BasicAuthenticator) Authenticate(request *http.Request) (*web.UserConte
 		},
 		AuthenticationType: web.Basic,
 		Name:               username,
-		AccessLevel:        web.GlobalAccess,
+		AccessLevel:        web.NoAccess,
 	}, httpsec.Allow, nil
 }
 
--- a/api/filters/basic_authentication.go
+++ b/api/filters/basic_authentication.go
@@ -85,6 +85,7 @@ func (a *BasicAuthenticator) Authenticate(request *http.Request) (*web.UserConte
 		},
 		AuthenticationType: web.Basic,
 		Name:               username,
+		Attributes:         make(map[string]string),
 	}, httpsec.Allow, nil
 }
 
--- a/pkg/security/filters/middlewares/authrization.go
+++ b/pkg/security/filters/middlewares/authrization.go
@@ -45,10 +45,12 @@ func (m *Authorization) Run(request *web.Request, next web.Handler) (*web.Respon
 		if !found {
 			return nil, security.UnauthorizedHTTPError("authorization failed due to missing authentication")
 		}
-		if accessLevel != web.DefaultAccess {
-			userContext.AccessLevel = accessLevel
-			request.Request = request.WithContext(web.ContextWithUser(ctx, userContext))
+		if accessLevel == web.DefaultAccess {
+			return nil, security.UnauthorizedHTTPError("authorization failed due to missing access level")
 		}
+
+		userContext.AccessLevel = accessLevel
+		request.Request = request.WithContext(web.ContextWithUser(ctx, userContext))
 
 		if !web.IsAuthorized(ctx) {
 			request.Request = request.WithContext(web.ContextWithAuthorization(ctx))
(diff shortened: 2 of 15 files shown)


pkg/web/user.go
```diff
@@ -17,13 +17,47 @@
...
+	// Such access might be granted to systems that need to manage global resources	GlobalAccess AccessLevel = iota
+	GlobalAccess
+
+	// TenantAccess means access was granted to manage the tenant's own resources. Such access might be granted to a user
+	// so the he can manage his own data
+	TenantAccess
+
+	// AllTenantAccess means access was granted to manage the resources of all tenants. Such access might be granted
+	// to systems that have to manage data across multiple tenants
+	AllTenantAccess
```
dpanayotov: how do `AllTenantAccess` and `GlobalAccess` relate? Do you get `AllTenantAccess` if you have `GlobalAccess`? If not then both of these give you so to say `FullAccess`, so does it make sense to allow multiple access levels in a user context?

pkg/web/user.go
```diff
@@ -17,13 +17,47 @@
...
+const (
+	//DefaultAccess is the default value for access level - it is used when a component does not expliciting set an access level
+	DefaultAccess AccessLevel = iota
+
+	// GlobalAccess means access was granted to manage global resources (such resources are not scoped or associated with tenant.
+	// Such access might be granted to systems that need to manage global resources	GlobalAccess AccessLevel = iota
+	GlobalAccess
+
+	// TenantAccess means access was granted to manage the tenant's own resources. Such access might be granted to a user
+	// so the he can manage his own data
```
dpanayotov: `so that`

pkg/web/user.go
```diff
@@ -17,13 +17,47 @@
...
+type AccessLevel int
+
+var levels = []string{"DefaultAccess", "GlobalAccess", "SingleTenantAccess", "AllTenantAccess"}
+
+const (
+	//DefaultAccess is the default value for access level - it is used when a component does not expliciting set an access level
+	DefaultAccess AccessLevel = iota
+
+	// GlobalAccess means access was granted to manage global resources (such resources are not scoped or associated with tenant.
+	// Such access might be granted to systems that need to manage global resources	GlobalAccess AccessLevel = iota
```
dpanayotov: `GlobalAccess AccessLevel = iota` leftover?

pkg/security/http/auth.go
```diff
@@ -58,9 +58,9 @@ type Authenticator interface {
 // returns a decision if the authorization passed
 //go:generate counterfeiter . Authorizer
 type Authorizer interface {
-	// Authorize returns decision specifying
-	// whether the authorizer ran or not and an error if one occurs
-	Authorize(req *web.Request) (Decision, error)
+	// Authorize returns decision specifying whether the authorizer allowed, denied or abstained from giving access,
+	// the access level associated with the decision and an error if one occurs
```
dpanayotov: Which access level should I return if I'm implementing an authorizer? The one in the user context or the one required to perform the operation? What am I expected to do with this access level if I'm a consumer of an authorizer?

pkg/security/filters/middlewares/authrization.go
```diff
@@ -41,6 +41,17 @@ func (m *Authorization) Run(request *web.Request, next web.Handler) (*web.Respon
 	switch decision {
 	case http.Allow:
 		ctx := request.Context()
+		userContext, found := web.UserFromContext(ctx)
+		if !found {
+			return nil, security.UnauthorizedHTTPError("authorization failed due to missing authentication")
+		}
+		if accessLevel == web.DefaultAccess {
+			return nil, security.UnauthorizedHTTPError("authorization failed due to missing access level")
```
dpanayotov: It's strange to get `allowed` and then still get `denied` because of access level.
Also, this error message is somewhat confusing. I have default access, it's not missing.

api/filters/basic_authentication.go
```diff
@@ -85,6 +85,7 @@ func (a *BasicAuthenticator) Authenticate(request *http.Request) (*web.UserConte
 		},
 		AuthenticationType: web.Basic,
 		Name:               username,
+		AccessLevel:        web.GlobalAccess,
```
dpanayotov: well.. yes and no... For now this might be ok

dpanayotov: Should the authenticators set that access level at all or should it be left to the authorizers?

pkg/web/user.go
KirilKabakchiev: AllTenantAccess  allows to see and manage only tenant scoped resources (but for all tenants)

GlobalAccess allows to manage all resources (listing with global access will return both tenant and non-tenant resources). Basically GlobalAccess means a tenantID should not be  looked for at all (there will be no tenant id in criterias)

api/filters/basic_authentication.go
KirilKabakchiev: I was really thinking about it (how it should look for OSS SM). Probably authenticators should put the default access level (zero value)

pkg/security/http/auth.go
KirilKabakchiev: the consumer is the authz filter that sets it in the context currently. So in the context that your authorizer would accept the only possible value for the AccessLevel will be whatever is the zero value used (or whatever the authentication layer set when creating the inital context)

PR #338, opened 2019-10-07
Document db migrations

dotchev:
Describe db migrations and how to upgrade/downgrade manually.

### DIFF
--- a/docs/development/db-schema.md
+++ b/docs/development/db-schema.md
@@ -38,15 +38,15 @@ On Windows adapt the shell variable syntax accordingly.
 
 Get the curent schema version:
 ```sh
-migrate -source "file://$MIGRATIONS" -database $DATABASE version
+migrate -source "file://$MIGRATIONS" -database "$DATABASE" version
 ```
 Go to specific version:
 ```sh
-migrate -source "file://$MIGRATIONS" -database $DATABASE goto $SCHEMA_VERSION
+migrate -source "file://$MIGRATIONS" -database "$DATABASE" goto $SCHEMA_VERSION
 ```
 
 ## Troubleshooting
-A common problem when switching SM versions is this error during startup"
+A common problem when switching SM versions is this error during startup:
 ```
 ERR panic: unbale to construct service-manager builder: error opening storage: error opening storage: could not update database schema: file does not exist
 ```
--- a/docs/development/db-schema.md
+++ b/docs/development/db-schema.md
@@ -1,0 +1,58 @@
+# Database Schema
+
+Currently Service Manager supports only PostgreSQL as a database.
+
+The database schema is built using a sequence of SQL scripts located in
+[storage/postgres/migrations](../../storage/postgres/migrations).
+For each schema version there are two scripts - _up_ and _down_.
+So it is possible also to downgrade the schema to a an older version.
+Each file name starts with some numbers - these represent the schema version.
+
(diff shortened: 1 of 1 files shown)


docs/development/db-schema.md
```diff
@@ -0,0 +1,60 @@
+# Database Schema
+
+Currently Service Manager supports only PostgreSQL as a database.
+
+The database schema is built using a sequence of SQL scripts located in
+[storage/postgres/migrations](../../storage/postgres/migrations).
+For each schema version there are two scripts - _up_ and _down_.
+So it is possible also to downgrade the schema to a an older version.
+Each file name starts with some numbers - these represent the schema version.
```
dzahariev: Most recent one are timestamps, as the migrations are now applied based on the timestamp ordering, instead numbers. This was introduced to allow database migrations to be added easily from SM extensions.

docs/development/db-schema.md
```diff
@@ -0,0 +1,60 @@
...
+
+These scripts are executed in a specific or by [migrate](https://github.com/golang-migrate/migrate) library
+during SM start up. This library records in the database the current schema version,
+so subsequent executions do not attempt to recreate the same db objects.
+It also uses db locks to prevent parallel migrations in case multple instances of SM start at the same time.
+
+## Upgrade / downgrade db schema
+Normally SM upgrade the db schea automatically during startup.
```
dzahariev: schea -> schema

docs/development/db-schema.md
```diff
@@ -0,0 +1,60 @@
...
+Still in rare cases it is necessary to do this manually. Here is how to do it.
```
dzahariev: I think one comma need to be added here:
Still in rare cases -> Still, in rare cases

docs/development/db-schema.md
```diff
@@ -0,0 +1,60 @@
...
+```sh
+migrate -source "file://$MIGRATIONS" -database "$DATABASE" version
+```
+Go to specific version:
+```sh
+migrate -source "file://$MIGRATIONS" -database "$DATABASE" goto $SCHEMA_VERSION
+```
+
+## Troubleshooting
+A common problem when switching SM versions is this error during startup:
```
dzahariev: is this error -> is the following error

docs/development/db-schema.md
```diff
@@ -0,0 +1,60 @@
...
+ERR panic: unbale to construct service-manager builder: error opening storage: error opening storage: could not update database schema: file does not exist
+```
+The reason usually is that an older version of SM was started after a newer one has updated the db schema.
+The older version is unaware of the new schema version and bails out with this error.
+See [this issue](https://github.com/golang-migrate/migrate/issues/79) in _migrate_.
+
+Normally SM should be updated only to newer versions.
+If for some reason it is necessary to downgrade it to an older version,
+then the db schame should be downgraded to the corresponding version too.
+See above how to do that.
```
dzahariev: I think this section need to be moved up, inside "Upgrade / downgrade db schema" section, as it explains the need of such operation.

docs/development/db-schema.md
dotchev: yes, now I installed the spell check plugin in vscode and it caught a typo in the SM error message ;)

docs/development/db-schema.md
dotchev: This section explains how to address the particular issue in Troubleshooting.
Also there might be different reasons to change the db schema manually.
So I would keep it here.

PR #339, opened 2019-10-11
Manage platform status during ping/pong flow

NickyMateev:
# Manage platform status during ping/pong flow

## Motivation

Currently managing platform status is done when WS connection is established and terminated with a platform (proxy agent). This leads to health check problems during B/G deployment when Service Manager runs on multiple nodes.

## Approach

Stop managing platform statuses on WS connection establishment/termination and instead manage statuses during ping/pong flow.

## Pull Request status

* [x] Initial implementation
* [x] Tests

### DIFF
--- a/test/ws_notification_test/ws_notification_test.go
+++ b/test/ws_notification_test/ws_notification_test.go
@@ -292,7 +292,7 @@ var _ = Describe("WS", func() {
 				By(fmt.Sprintf("Ping received and active status is true, then when %v ping timeout passes, active status should be set to false", pingTimeout))
 
 				ctx, _ := context.WithTimeout(context.TODO(), pingTimeout+time.Second)
-				ticker := time.NewTicker(500 * time.Millisecond)
+				ticker := time.NewTicker(pingTimeout / 3)
 				for {
 					select {
 					case <-ticker.C:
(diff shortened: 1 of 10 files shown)


test/ws_notification_test/ws_notification_test.go
```diff
@@ -235,52 +236,77 @@ var _ = Describe("WS", func() {
...
+		Context("when ping is not received", func() {
+			It("should switch platform's active status to false", func() {
+				Expect(conn.WriteControl(websocket.PingMessage, []byte("pingping"), time.Now().Add(pingTimeout))).ShouldNot(HaveOccurred())
+				Eventually(pongCh).Should(BeClosed()) // wait for a pong message
+				assertPlatformIsActive()
+
+				By(fmt.Sprintf("Ping received and active status is true, then when %v ping timeout passes, active status should be set to false", pingTimeout))
+
+				ctx, _ := context.WithTimeout(context.TODO(), pingTimeout+time.Second)
+				ticker := time.NewTicker(500 * time.Millisecond)
```
georgifarashev: Can we do this a fraction of the `pingTimeout`? Let's say pingTimeout/3 for example.
Minor: Also is it possible to reduce the pingTimeout to lets say 1/2 sec.? It might speed up the tests.

DimitarPetrov: It is 1 second currently the BeforeEach overrides it (https://github.com/Peripli/service-manager/pull/339/files/84a78c676dc2d747851c0ec5a7a9099a0bb672ff#diff-234cfb3ff7d03d66a6db4f491358a003R74)

PR #340, opened 2019-10-14
Allow settings ssl when connecting with notificator

pankrator:
## Prerequisites

The following error occurs when connecting to postgres for notifications: 

level=error msg="DB connection for notifications closed" component="postgres/notificator.go:108" correlation_id=- error="pq: SSL is not enabled on the server"

## Approach

Allow settings the connection to be without ssl if configured as such.

### DIFF
--- a/storage/postgres/notification_connection.go
+++ b/storage/postgres/notification_connection.go
@@ -32,12 +32,17 @@ type notificationConnectionCreator interface {
 
 type notificationConnectionCreatorImpl struct {
 	storageURI           string
+	skipSSSLValidation   bool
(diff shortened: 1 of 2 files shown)


storage/postgres/notification_connection.go
```diff
@@ -32,12 +32,17 @@ type notificationConnectionCreator interface {
 
 type notificationConnectionCreatorImpl struct {
 	storageURI           string
+	skipSSSLValidation   bool
```
dzahariev: skipSSLValidation

PR #342, opened 2019-10-17
Rework SQL queries

KirilKabakchiev:
SQL queries that were currently built contained bugs when paging is used. Also they were not optimal (new queries are leaner and faster)


storage/postgres/query_builder.go
```diff
@@ -44,336 +84,328 @@ func NewQueryBuilder(db pgDB) *QueryBuilder {
...
+	return fmt.Sprintf(" WHERE %s", whereSQL)
+}
 
-	if len(pgq.returningFields) == 1 {
-		pgq.sql.WriteString(fmt.Sprintf(" RETURNING " + pgq.returningFields[0]))
-	} else if len(pgq.returningFields) > 0 {
-		pgq.sql.WriteString(" RETURNING " + strings.Join(pgq.returningFields, ", "))
+func (pq *pgQuery) returningSQL() string {
+	fieldsCount := len(pq.returningFields)
+	switch fieldsCount {
```
DimitarPetrov: The whole switch is redundant since this is handled in `string.Join(...)`
```
func Join(a []string, sep string) string {
	switch len(a) {
	case 0:
		return ""
	case 1:
		return a[0]
...
```

storage/postgres/query_builder.go
```diff
@@ -44,336 +97,339 @@ func NewQueryBuilder(db pgDB) *QueryBuilder {
...
 }
 
-func (pgq *pgQuery) orderBySQL() *pgQuery {
-	if sql, err := orderBySQL(pgq.orderByFields); err != nil {
-		pgq.err = err
-		return pgq
-	} else {
-		pgq.sql.WriteString(sql)
+// joinSQL is a performance improvement - queries can work with LEFT JOIN always but are slower
+// JOIN is used when a label quert is present, LEFT JOIN is used when no label query is present so that resultset includes
```
georgifarashev: quert => query?

storage/postgres/query_builder.go
```diff
@@ -12,23 +15,74 @@ import (
...
-	current = strings.Replace(current, old, new, 1)
-	qsb.WriteString(current)
-}
+const PrimaryKeyColumn = "id"
+
+const COUNTWithoutCriteriaTemplate = `
+SELECT COUNT(DISTINCT {{.ENTITY_TABLE}}.{{.PRIMARY_KEY}})
+FROM {{.ENTITY_TABLE}}
+{{.FOR_SHARE_OF}}
+{{.LIMIT}};`
```
DimitarPetrov: Try to remove the limit from count queries by calculating a specific set of data for template resolving per operation instead of one with all the clauses for each operation.

PR #344, opened 2019-10-24
Assert service and plan catalog names are CLI-friendly

DimitarPetrov:
## Motivation
Service Manager does not validate the broker catalog (as per OSB-spec) when a broker is registered. Currently only general checks are performed.
This results in broker registration failures in the respective platforms that already validate.
The names must be "CLI-friendly" as described in the OSB spec https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#body

## Approach
Match service and plan names with regex for CLI-friendly names as part of the validation and return error if not matched.


pkg/util/api_test.go
```diff
@@ -86,6 +86,35 @@ var _ = Describe("Utils test", func() {
...
+			}
+			for _, v := range cases {
+				name := v
+				It(fmt.Sprintf("%s should be considered not CLI-friendly", name), func() {
+					Expect(util.IsCLIFriendly(name)).To(Equal(false))
+				})
+			}
+		})
+		Context("name with only alaphanumeric, period or hyphens characters", func() {
+			cases := []string{
```
dpanayotov: Should starting and ending with `_`,`-`,`.` be considered valid?

pkg/types/service_offering.go
```diff
@@ -63,6 +63,11 @@ func (e *ServiceOffering) Validate() error {
 	if e.BrokerID == "" {
 		return fmt.Errorf("service offering broker id missing")
 	}
+	if cliFriendly, err := util.IsCLIFriendly(e.CatalogName); err != nil {
+		return err
+	} else if !cliFriendly {
+		return fmt.Errorf("service offering catalog name is not CLI-friendly")
```
dpanayotov: can you extract this error and include what actually CLI-friendly means in the response.

pkg/types/service_plan.go
```diff
@@ -59,6 +59,11 @@ func (e *ServicePlan) Validate() error {
 	if e.ServiceOfferingID == "" {
 		return fmt.Errorf("service plan service offering id missing")
 	}
+	if cliFriendly, err := util.IsCLIFriendly(e.CatalogName); err != nil {
+		return err
```
dpanayotov: this is actually an internal error and should not be returned. This will return internal error as 400. Maybe log the actual error and return something generic such as `Could not validate ___ name due to internal error. Check logs for more information.`

KirilKabakchiev: its **not** an internal error - a user registers his broker and this user knows his broker credentials so this user can access the catalog in the first place - the broker as well as the broker catalog are user INPUT. Disclosing information about the catalog does not disclose anything internal as the user can access the catalog in the first place.

If this broker's catalog has a not-cli-friendly entity name we should let the user know exactly whats wrong with his broker catalog and not make him open a ticket in order to figure that out.

dpanayotov: The error is actually not being able to compile the regex. If the name is valid or not is returned as a boolean. That’s why it’s internal error.

pkg/util/api_test.go
DimitarPetrov: From K8S documentation:

> Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between.

So i assume it is not considered valid.

DimitarPetrov:
We must clarify if this approach is desired and if it is desired for broker names as well.

DimitarPetrov:
This is no longer needed since Service Catalog addressed this issue.

PR #351, opened 2019-11-18
Provide information about service instances

NickyMateev:
# Provide information about service instances

## Motivation

Service Manager should start storing information about service instances and in this way provide a central place where:
* Service Instances can be managed and retrieved
* Enable scenarios such as out-of-platform consumption
* Enforce constraints between other resources (example: platforms or service plans should not be deleted if there are associated service instances created already)
* Central quota check about service consumption
* And others...

## Description

This is the topic branch for this task/story.

## Tasks:
* [Introduce Operation resource](https://github.com/Peripli/service-manager/pull/352) ✅
* [Introduce Service Instance resource](https://github.com/Peripli/service-manager/pull/353) ✅
* [Introduce Service Instance API](https://github.com/Peripli/service-manager/pull/354) ✅
* [Secure Service Instances API](https://github.com/Peripli/service-manager/pull/359) ✅
* [Introduce Service Instance labelling interceptor](https://github.com/Peripli/service-manager/pull/356) ✅
* [Store instance information on proxied OSB requests](https://github.com/Peripli/service-manager/pull/361) ✅
* Fail delete service plan if instances exist ✅
* Fail delete platform if instances exist ✅

PR #353, opened 2019-11-18
Introduce Service Instance resource

NickyMateev:
# Introduce Service Instance resource

## Motivation

Introduces Service Instance resource so Service Manager can start manage such entities.

## Pull Request status

* [x] Initial implementation

## Topic branch:
https://github.com/Peripli/service-manager/pull/351

### DIFF
--- a/pkg/types/service_instance.go
+++ b/pkg/types/service_instance.go
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2018 The Service Manager Authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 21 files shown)


pkg/types/service_instance.go
```diff
@@ -0,0 +1,60 @@
...
+type ServiceInstance struct {
+	Base
+	Name            string          `json:"name"`
+	ServicePlanID   string          `json:"service_plan_id"`
+	PlatformID      string          `json:"platform_id"`
+	MaintenanceInfo json.RawMessage `json:"maintenance_info"`
+	Context         json.RawMessage `json:"-"`
+	PreviousValues  json.RawMessage `json:"-"`
+	Ready           bool            `json:"ready"`
+	Usable          bool            `json:"usable"`
```
dzahariev: What is the semantic of this field

NickyMateev: When a broker returns an error that the instance is no longer usable (due to an error during async update or deprovision) we also store that information in SM.

NickyMateev:
@KirilKabakchiev 
How do you propose component testing GET/LIST when there are no Create/Update/Delete APIs yet? 
There's no direct way to create a resource, there's also no indirect way (eg. services and plans).

Also, the state of the controller right now is plain-vanilla base controller logic, nothing special about, which already has tests.

KirilKabakchiev:
This is a new API and if you don't enable the list/get component tests its hard to guarantee that querying is working properly for all fields. I would suggest to implement the blueprint for instances by using the repository that is provided in the context

PR #356, opened 2019-11-20
Introduce Service Instance labelling interceptor

NickyMateev:
# Introduce Service Instance labelling interceptor

## Motivation

Service Manager needs to label all instances with the tenant in which it is created.

## Approach

To do this, there needs to be an AroundTx interceptor that adds a <tenant_identifier> label to the service instance. The value of the <tenant_identifier> label can be obtained from the context field in the service instance.

## Pull Request status

* [x] Initial implementation
* [x] Integration tests

### DIFF
--- a/pkg/sm/sm.go
+++ b/pkg/sm/sm.go
@@ -406,5 +406,8 @@ func (smb *ServiceManagerBuilder) EnableMultitenancy(labelKey string, extractTen
 
 	multitenancyFilters := filters.NewMultitenancyFilters(labelKey, extractTenantFunc)
 	smb.RegisterFiltersAfter(filters.ProtectedLabelsFilterName, multitenancyFilters...)
+	smb.WithCreateInterceptorProvider(types.ServiceInstanceType, &interceptors.ServiceInstanceCreateInsterceptorProvider{
+		TenantIdentifier: labelKey,
+	}).Register()
 	return smb
 }
--- a/pkg/sm/sm.go
+++ b/pkg/sm/sm.go
@@ -406,7 +406,7 @@ func (smb *ServiceManagerBuilder) EnableMultitenancy(labelKey string, extractTen
 
 	multitenancyFilters := filters.NewMultitenancyFilters(labelKey, extractTenantFunc)
 	smb.RegisterFiltersAfter(filters.ProtectedLabelsFilterName, multitenancyFilters...)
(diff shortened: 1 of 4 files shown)


storage/interceptors/service_instance_create_interceptor.go
```diff
@@ -0,0 +1,77 @@
...
+func (c *serviceInstanceCreateInterceptor) AroundTxCreate(h storage.InterceptCreateAroundTxFunc) storage.InterceptCreateAroundTxFunc {
+	return func(ctx context.Context, obj types.Object) (types.Object, error) {
+		serviceInstance := obj.(*types.ServiceInstance)
+
+		bytes, err := json.Marshal(&serviceInstance.Context)
+		if err != nil {
+			return nil, err
+		}
+
+		tenantID := gjson.Get(string(bytes), c.TenantIdentifier)
```
pankrator: Use directly `gjson.GetBytes`

pkg/sm/sm.go
```diff
@@ -406,5 +406,8 @@ func (smb *ServiceManagerBuilder) EnableMultitenancy(labelKey string, extractTen
 
 	multitenancyFilters := filters.NewMultitenancyFilters(labelKey, extractTenantFunc)
 	smb.RegisterFiltersAfter(filters.ProtectedLabelsFilterName, multitenancyFilters...)
+	smb.WithCreateInterceptorProvider(types.ServiceInstanceType, &interceptors.ServiceInstanceCreateInsterceptorProvider{
+		TenantIdentifier: labelKey,
```
pankrator: Is the TenantIdentifier (meaning the property in the context) the same key as the key used in MultitenancyFilter?

NickyMateev: Yes it is.

storage/interceptors/service_instance_create_interceptor.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		labels[c.TenantIdentifier] = []string{tenantID.String()}
+
+		serviceInstance.SetLabels(labels)
+
+		return h(ctx, serviceInstance)
+	}
+}
+
+func (c *serviceInstanceCreateInterceptor) OnTxCreate(f storage.InterceptCreateOnTxFunc) storage.InterceptCreateOnTxFunc {
```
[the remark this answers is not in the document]
NickyMateev: good point

PR #359, opened 2019-11-21
Add authentication for service instance endpoint

georgifarashev:
# Add authentication for service instance endpoint

## Motivation

The service instance api should be used only by authenticated users.

## Approach

Extend the existing authentication mechanism by adding the service instance url.

## Pull Request status

For example:

* [x] Implementation
* [x] Tests

### DIFF
--- a/test/service_instance_test/service_instance_test.go
+++ b/test/service_instance_test/service_instance_test.go
@@ -19,12 +19,13 @@ package service_test
 import (
 	"context"
 	"fmt"
-	"github.com/Peripli/service-manager/pkg/query"
-	"github.com/Peripli/service-manager/pkg/types"
-	"github.com/gofrs/uuid"
 	"net/http"
 	"testing"
 	"time"
+
+	"github.com/Peripli/service-manager/pkg/query"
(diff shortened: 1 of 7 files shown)


test/service_instance_test/service_instance_test.go
```diff
@@ -125,7 +128,10 @@ var _ = test.DescribeTestsFor(test.TestCase{
 
 func blueprint(ctx *common.TestContext, auth *common.SMExpect) common.Object {
 	serviceInstance := prepareServiceInstance(ctx, auth, fmt.Sprintf(`{"%s":"%s"}`, TenantIdentifier, TenantValue))
-	ctx.SMRepository.Create(context.Background(), serviceInstance)
+	_, err := ctx.SMRepository.Create(context.Background(), serviceInstance)
+	if err != nil {
```
KirilKabakchiev: Expect err to not have occurred instead ?

georgifarashev: At this point the fail handler is not yet registered. Not quite sure why.

PR #365, opened 2019-12-03
Introduce Resource Validator [API-level validation]

NickyMateev:
# Introduce Resource Validator [API-level validation]

## Motivation [related to #358]

Current validation of entities approach is very limited and hidden. It allows us to pretty much check only entity properties for validity.

The intent of this change is to expand on the term "validate" and allow making API-level validation, or in other words to validate whether specific POST/PATCH/DELETE requests are valid.

This way we can make custom validation checks prior to actually executing the POST/PATCH/DELETE operation.

Examples of such use cases are:
- Platform entities should not be deleted if Service Instance entities exist
- Broker entities should not be deleted if Service Instance entities associated with one of the broker's catalog exist
- Broker's catalog should not allow removal of one of its Service Plan entities if Service Instance entities associated with this service plan exist 

## Approach

A new interface `ResourceValidator` is introduced. Every base controller will have one.

```go
type ResourceValidator interface {
	ValidateCreate(ctx context.Context, repository storage.Repository, object types.Object) error
	ValidateUpdate(ctx context.Context, repository storage.Repository, object types.Object) error
	ValidateDelete(ctx context.Context, repository storage.Repository, object types.Object) error
}
```

The idea is that just before passing the current object for _Creation/Update/Deletion_ to a controller's repository to execute the request, the corresponding `ValidateCreate`, `ValidateUpdate` and `ValidateDelete` will be called.

There is a `DefaultResourceValidator` for controllers which don't require custom validation logic.

If someone wants to create a custom API validator for some resource, he only has to implement `ResourceValidator` and provide that implementation to the specific API controller. 

## Use case example

Broker's catalog should not allow removal of one of its Service Plan entities if Service Instance entities associated with this service plan exist. This means we need to introduce a `BrokerValidator` and define it's `ValidateUpdate` method to check what's needed.

Here's an example error message if the validation fails:

```json
{
  "error": "ExistingReferenceEntity",
  "description": "Error executing ServiceBroker request: Could not delete ServicePlan due to one or more existing ServiceInstance reference entities with ID(s): [57bd8c39-d736-4089-aa5c-8a28e0f5d749,e3f8db5b-4425-4a4c-8bc5-95825f785d26,095d0237-ad89-4ef5-8f15-c09216bb52c9]"
}
```

## Pros/Cons of such an API-level validation approach

Pros:
* Can have custom validation logic for each API type
* Gives access to violation IDs (in cases such as foreign key violation constraints, eg. deleting a Platform with Service Instances) 
* Save valuable time troubleshooting

Cons:
* Possible race conditions - validation DB calls are not in the same transaction as the requested CUD operation. **[we should think of a way to move this API-level validation a little further down (eg. somewhere right after the start of the interceptor chain) - this way we can ensure that no race conditions will occur]**. If we cannot mitigate this - we can at least combine this approach with #358 and in this way when a race condition does occur, the client will receive 400 Bad request with both entity and violation entity - they just won't have the violation IDs in the response message.
* Extra DB calls (unavoidable, requiring to do custom validation pretty much guarantees that this will happen)
* Will have to call Validate method manually in certain places

## Note

This change allows us to remove the hidden `Validate()` invocation from the `BytesToObject()` implementation. If this new validation approach is adopted, the other repos need to be updated.

## Pull Request status

* [x] Initial implementation
* [x] Integration tests

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -20,6 +20,7 @@ package api
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	"github.com/Peripli/service-manager/pkg/env"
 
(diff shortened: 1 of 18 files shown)


NickyMateev:
Closing this as team decided the feature is not yet needed.

PR #369, opened 2019-12-09
Delete notifications to not return result

KirilKabakchiev:
Notifications payload is often large and loading deleted notifications inmemory is not necessary. In general in most places we delete stuff, loading it inmemory is not needed. The PR addresses these points

### DIFF
--- a/storage/interfaces.go
+++ b/storage/interfaces.go
@@ -51,23 +51,25 @@ var (
 
 // Settings type to be loaded from the environment
 type Settings struct {
-	URI                string                `mapstructure:"uri" description:"URI of the storage"`
-	MigrationsURL      string                `mapstructure:"migrations_url" description:"location of a directory containing sql migrations scripts"`
-	EncryptionKey      string                `mapstructure:"encryption_key" description:"key to use for encrypting database entries"`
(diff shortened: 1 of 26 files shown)


storage/postgres/query_builder.go
```diff
@@ -147,12 +148,25 @@ func (pq *pgQuery) Count(ctx context.Context) (int, error) {
 	return count, nil
 }
 
-func (pq *pgQuery) Delete(ctx context.Context) (*sqlx.Rows, error) {
+func (pq *pgQuery) Delete(ctx context.Context) (*sqlx.Rows, sql.Result, error) {
```
georgifarashev: Is it possible to split this into two functions. Delete and DeleteReturning?

storage/interfaces.go
```diff
@@ -163,8 +163,11 @@ type Repository interface {
 	// Count retrieves number of objects of particular type in SM DB
 	Count(ctx context.Context, objectType types.ObjectType, criteria ...query.Criterion) (int, error)
 
-	// Delete deletes an object from SM DB
-	Delete(ctx context.Context, objectType types.ObjectType, criteria ...query.Criterion) (types.ObjectList, error)
+	// DeleteReturning deletes objects from SM DB
```
NickyMateev: And returns them as result

PR #370, opened 2019-12-10
Batch fetching details for visibility notifications

dotchev:
# Pull Request Template

## Motivation

Deleting a visibility creates the corresponding notification. The notification requires also additional data, so for each visibility SM loads its plan, service offering and broker - 3 db calls for each visibility.
In case multiple visibilities are deleted in one operation this produces a lot of db calls and the operation may take 10+ seconds.

## Approach

Now the additional data is loaded for all deleted visibilities with just 3 db calls independent of the visibility count.

## Pull Request status

* [x] Refactoring


NickyMateev:
Needs rebase.

PR #373, opened 2019-12-15
Async API Engine

NickyMateev:
# Async API Engine

## Motivation

Currently the SM API is fully synchronous. However, this is no longer sufficient.
- Some of SM's current APIs (namely CU /v1/service_brokers) take a long time to execute on bigger environments (close to 60 seconds) because a lot of things are happening under the hood
- The current synchronous broker API does not wait for confirmation that the notifications that are created during broker registration/update have been processed
- SM has to eventually behave as an OSB platform - meaning SM should have "lifecycle APIs" for managing service instances and service bindings directly in SM

The general idea behind the SM async APIs is described in https://github.com/Peripli/specification/pull/31

## Approach

A new `operations` package is added. It introduces abstractions for:
- **JobScheduler** - schedules async API jobs
- **WorkerPool** - executes schedules async API jobs
- **Maintainer** - cleans up operations older than specified time and also ensures no orphan operations are left out

The abstraction for a `Controller` is extended and now controller can provide a `JobScheduler` - for APIs which are not meant to be executed asynchronously an empty implementation can be provided.

## Pull Request status

* [x] Initial implementation
* [x] Refactoring
* [x] Unit tests
* [x] Integration tests


operations/scheduler.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Run starts the DefaultScheduler's worker pool enabling him to start polling
+// for scheduled jobs
+func (ds *DefaultScheduler) Run() {
+	ds.workerPool.Run()
+}
+
+// SchedulerCreate schedules a Create job in the worker pool
+func (ds *DefaultScheduler) ScheduleCreate(ctx context.Context, object types.Object, operationID string) {
+	reqCtxClone := util.RequestContextCopy{Context: ctx}
```
NickyMateev: Context still gets canceled as soon as the original request context gets canceled..

operations/worker_pool.go
```diff
@@ -0,0 +1,70 @@
...
+func (wp *WorkerPool) Run() {
+	go wp.processJobs()
+}
+
+// processJobs polls the currently scheduled jobs and as long as there
+// are available workers it assigns a worker to execute for each available scheduled job
+func (wp *WorkerPool) processJobs() {
+	for {
+		job := <-wp.jobs
+		for wp.currentWorkers >= wp.poolSize {
```
NickyMateev: Need to do something cleaner

operations/jobs.go
```diff
@@ -0,0 +1,166 @@
...
+
+	operation.State = state
+	operation.UpdatedAt = time.Now()
+
+	if opErr != nil {
+		bytes, err := json.Marshal(opErr)
+		if err != nil {
+			return err
+		}
+		operation.Errors = json.RawMessage(bytes)
```
NickyMateev: Doesn't get set in DB for some reason

pkg/web/controller.go
```diff
@@ -20,6 +20,14 @@ package web
 type Controller interface {
 	// Routes returns the set of routes for this controller
 	Routes() []Route
+
+	// Scheduler returns the Async job scheduler for this controller (and whether it has one)
+	Scheduler() (bool, JobScheduler)
```
pankrator: swap the position of the return types

api/base_controller.go
```diff
@@ -55,17 +60,27 @@ type BaseController struct {
 }
 
 // NewController returns a new base controller
-func NewController(repository storage.Repository, resourceBaseURL string, objectType types.ObjectType, objectBlueprint func() types.Object, defaultPageSize, maxPageSize int) *BaseController {
+func NewController(ctx context.Context, repository storage.Repository, resourceBaseURL string, objectType types.ObjectType, objectBlueprint func() types.Object, options *Options, supportsAsync bool) *BaseController {
+	var scheduler *operations.DefaultScheduler
+	if supportsAsync {
+		scheduler = operations.NewScheduler(ctx, repository, options.OperationSettings)
```
pankrator: Perhaps it will be easier to maintain one scheduler per Service Manager.

operations/worker_pool.go
```diff
@@ -0,0 +1,59 @@
...
+		repository: repository,
+		jobs:       make(chan ExecutableJob, options.PoolSize),
+		workers:    make(chan struct{}, options.PoolSize),
+		jobTimeout: options.JobTimeout,
+	}
+}
+
+// Run starts the worker pool so it can start polling for scheduled jobs
+func (wp *WorkerPool) Run() {
+	go wp.processJobs()
```
pankrator: There are jobs also in the scheduler. No jobs are put in the worker pool. I suppose it is a typo?

api/api.go
```diff
@@ -94,19 +95,19 @@ func New(ctx context.Context, e env.Environment, options *Options) (*web.API, er
 	return &web.API{
 		// Default controllers - more filters can be registered using the relevant API methods
 		Controllers: []web.Controller{
-			NewController(options.Repository, web.ServiceBrokersURL, types.ServiceBrokerType, func() types.Object {
+			NewController(ctx, options.Repository, web.ServiceBrokersURL, types.ServiceBrokerType, func() types.Object {
```
DimitarPetrov: Maybe different constructors оn the basis of the last boolean property. Something like NewController and NewAsyncController, since this bool is a little confusing right now.

operations/maintainer.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/Peripli/service-manager/pkg/types"
+	"github.com/Peripli/service-manager/pkg/util"
+	"github.com/Peripli/service-manager/storage"
+	"time"
+)
+
+// OperationMaintainer ensures that operations old enough are deleted
+// and that no stuck (orphan) operations are left in the DB due to crashes/restarts of SM
+type OperationMaintainer struct {
+	repository      storage.Repository
```
pankrator: Use transactional repository here and wrap the Execute in InTransaction

operations/worker_pool.go
```diff
@@ -0,0 +1,59 @@
...
+
+		go func() {
+			defer func() {
+				<-wp.workers
+			}()
+
+			ctxWithTimeout, cancel := context.WithTimeout(wp.smCtx, wp.jobTimeout)
+			defer cancel()
+
+			operationID, err := job.Execute(ctxWithTimeout, wp.repository)
```
pankrator: Pass a repository in transaction, or pass a TransactionalRepository. Then in the execute wrap the repository calls in transaction

pkg/sm/sm.go
```diff
@@ -191,6 +197,17 @@ func (smb *ServiceManagerBuilder) Build() *ServiceManager {
 	srv := server.New(smb.cfg.Server, smb.API)
 	srv.Use(filters.NewRecoveryMiddleware())
 
+	// start the operation maintainer
+	smb.OperationMaintainer.Run()
+
+	// start each Controller scheduler's worker pool
+	for _, controller := range smb.API.Controllers {
```
pankrator: Provide different configurations for each scheduler's pool when creating it. It will allow for fine grained configuration per resource

api/base_controller.go
```diff
@@ -55,17 +60,27 @@ type BaseController struct {
...
 		objectType:      objectType,
-		DefaultPageSize: defaultPageSize,
-		MaxPageSize:     maxPageSize,
+		DefaultPageSize: options.APISettings.DefaultPageSize,
+		MaxPageSize:     options.APISettings.MaxPageSize,
 	}
 }
 
+func (c *BaseController) Scheduler() (bool, web.JobScheduler) {
+	return c.supportsAsync(), c.scheduler
```
DimitarPetrov: For me the scheduler first and bool second is more intuitive. Like the build in map operations.

api/base_controller.go
```diff
@@ -114,6 +129,10 @@ func (c *BaseController) Routes() []web.Route {
 	}
 }
 
+func (c *BaseController) supportsAsync() bool {
```
DimitarPetrov: Seems like unnecessary function to me

operations/config.go
```diff
@@ -0,0 +1,30 @@
...
+func DefaultSettings() *Settings {
+	return &Settings{
+		PoolSize:        1000,
+		JobTimeout:      10 * time.Minute,
+		CleanupInterval: 60 * time.Minute,
+	}
+}
+
+// Validate validates the Operations settings
+func (s *Settings) Validate() error {
```
DimitarPetrov: Maybe JobTimeout and CleanupInterval should be validated too because with a really small values this can make SM non operational

operations/worker_pool.go
```diff
@@ -0,0 +1,59 @@
...
+			if err != nil {
+				log.D().Debugf("Error occurred during execution of operation with ID (%s): %s", operationID, err.Error())
```
DimitarPetrov: you can return here and remove the else

operations/maintainer.go
```diff
@@ -0,0 +1,89 @@
...
+func (om *OperationMaintainer) Run() {
+	go om.cleanupOperations()
+	go om.cleanupStuckOperations()
+}
+
+// cleanOperations cleans up periodically all operations but the last
+// for each C/U/D operation for every resource_id which are older than some specified time
+func (om *OperationMaintainer) cleanupOperations() {
+	ticker := time.NewTicker(om.cleanupInterval)
+	terminate := make(chan struct{})
```
DimitarPetrov: who writes in this channel and terminates the maintainer. Probably it is better to have a context here and use it's done chan

operations/jobs.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	_, err = repository.Update(ctx, operation, query.LabelChanges{})
+	if err != nil {
+		log.D().Debugf("Failed to update state of operation with id (%s) to SUCCEEDED", operationID)
+		return err
+	}
+
+	log.D().Debugf("Successfully updated state of operation with id (%s) to SUCCEEDED", operationID)
```
DimitarPetrov: `log.D().Debugf("Failed to update state of operation with id (%s) to %s", operationID, state)`
It is not only SUCCEEDED

operations/maintainer.go
NickyMateev: Discussed that it's not necessary.

operations/worker_pool.go
NickyMateev: Discussed that it's not necessary.

operations/maintainer.go
NickyMateev: yep, will use the smCtx here

api/base_controller.go
NickyMateev: We discussed it and came to the conclusion that there are no significant benefits and that there would actually be more implications of such an approach.

operations/worker_pool.go
NickyMateev: When constructing the scheduler we assign to it the worker pool's job channel. Both components communicate through it afterwards.

pkg/sm/sm.go
NickyMateev: I'll have to think about a **clear** way to provide such a configuration.

dpanayotov:
Two general things we might think on while designing this:
* Does it make sense to provide an API to view all of the running jobs?
* is there a way that a job may get stuck? If yes, there needs to be a way to cancel the job externally (e.g. via an API)

NickyMateev:
Things to add before finishing this POC:
- Introduce ClaimedAt property to operation & fix the updateOrphanOperations function
- Update operations to failed if a job takes too long

KirilKabakchiev:
Operation has to be created as soon as possible (before returning 202 accepted) as the job might not be picked up for proceesing immediately if there are no free resources (keep in mind the client starts polling immediately and expects to get  a meaningful response on polling)

operations/jobs.go
NickyMateev: Fixed

NickyMateev:
The core of the implementation is in place. There are things that we as a team need to discuss before proceeding. Among which:

1. How do we determine which operations are stuck/orphaned due to SM crash? Current design had us marking operations created longer than the last `job_timeout` period as `FAILED`, but this is not a good indicator, as an operation may have not been claimed for processing due to unavailable workers:
    - **Option 1:** Introduce ClaimedAt property to operations resource. We'll mark operation as claimed once it starts executing. [Proposal PR](https://github.com/Peripli/service-manager/pull/378)
    - **Option 2:** Categorise operations older than **X** number of `job_timeout` periods. This way we don't introduce new properties.
2. Discarding executing Create/Update operations if they are followed immediately by Delete operation.
    - This is doable, but will require a pgnotify-like implementation in order for all nodes of SM to get notified when a new operation is stored. All currently executing jobs on every node would be able to check if the new operation stored is a `DELETE` operation for the same resource and could cancel the current `CREATE` or `UPDATE` operation.
    - It would be best if we could reuse our `Notificator` abstraction. It's not as generic as we'd like though, as it's very coupled with the our SM notifications. So this would require some time spent refactoring.
3. Consider worker pool logic:
    1. Using an in-memory jobs channel processed by worker pool (as is the current implementation)
    2. Constantly pinging DB for new operations processed by worker pool (no in-memory jobs channel, very costly, much overkill)
    3. _pgnotify-like_ implementation where worker pool listens for new operation entities (requires refactoring to `Notificator` like mentioned above)
    4. No worker pool and no in-memory jobs channel. Proposal PR](https://github.com/Peripli/service-manager/pull/379)
        - Why do we need the worker pool? (to have a dedicated configurable abstraction over execution of asynchronous jobs)
        - Why separate pool per resource? (some requests might take much longer than other to process; spawn a whole bunch of them and you could use up all available workers and block other potentially faster requests)
        - How do we determine the optimal pool size? If we set the pool size too low, the jobs will be delayed unnecessarily. (experience will show for itself)

api/base_controller.go
```diff
@@ -293,13 +377,42 @@ func (c *BaseController) PatchObject(r *web.Request) (*web.Response, error) {
...
+		if err := c.checkAsyncSupport(); err != nil {
+			return nil, err
+		}
+
+		operation, err := c.buildOperation(ctx, c.repository, types.IN_PROGRESS, types.UPDATE, objFromDB.GetID(), log.CorrelationIDFromContext(ctx))
+		if err != nil {
+			return nil, err
+		}
+
+		operationID, err := c.scheduler.Schedule(operations.Job{
```
KirilKabakchiev: .Schedule can accept ctx as first argument and internally only use the values and create a new contextWithtimeout. probably the whole Stateecoontext struct and reqctx field on job can be removed

operations/jobs.go
```diff
@@ -0,0 +1,102 @@
...
+	}()
+
+	if _, err = j.operationFunc(reqCtx, repository); err != nil {
+		log.D().Debugf("Failed to execute %s operation with id (%s) for %s entity", j.operation.Type, j.operation.ID, j.objectType)
+
+		if timedOut {
+			err = errors.New("job timed out")
+		}
+
+		if opErr := updateOperationState(opCtx, repository, j.operation.ID, types.FAILED, &OperationError{Message: err.Error()}); opErr != nil {
```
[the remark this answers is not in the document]
NickyMateev: I added this, but not sure how we can test it..

NickyMateev:
image

PR #374, opened 2019-12-18
Check instance ownership on bind requests

DimitarPetrov:
## Motivation

During an OSB bind request, SM must check if the instance owner is the same as the one requesting the bind operation as someone might call SM directly with the broker credentials and bypass the platform's security.

## Approach
A new plugin is introduced which checks if the tenant identifier of the caller is the same as the tenant identifier of the owner of the service instance on bind and update service instance. If this two does not match 404 is returned.

Another plugin is introduced to check the platform_id associated with the provided credentials match the platform_id of the instance on which the operation is performed.

## Pull Request status

* [x] Implementation
* [x] Tests

### DIFF
--- a/api/plugins/check_instance_owner_plugin.go
+++ b/api/plugins/check_instance_owner_plugin.go
@@ -19,7 +19,7 @@ type checkInstanceOwnerPlugin struct {
 	tenantIdentifier string
 }
 
-// NewCheckInstanceOwnerPlugin creates new plugin that checks the owner of the instance on bind requests
+// NewCheckInstanceOwnerPlugin creates new plugin that checks the owner of the instance
 func NewCheckInstanceOwnerPlugin(repository storage.Repository, tenantIdentifier string) *checkInstanceOwnerPlugin {
 	return &checkInstanceOwnerPlugin{
 		repository:       repository,
--- a/api/plugins/check_instance_owner_plugin.go
+++ b/api/plugins/check_instance_owner_plugin.go
@@ -45,6 +45,11 @@ func (p *checkInstanceOwnerPlugin) UpdateService(req *web.Request, next web.Hand
 func (p *checkInstanceOwnerPlugin) assertOwner(req *web.Request, next web.Handler) (*web.Response, error) {
 	ctx := req.Context()
 	callerTenantID := gjson.GetBytes(req.Body, "context."+p.tenantIdentifier).String()
+	if len(callerTenantID) == 0 {
+		log.C(ctx).Info("Tenant identifier not found in request context.")
+		return next.Handle(req)
+	}
+
 	instanceID := req.PathParams["instance_id"]
 	byID := query.ByField(query.EqualsOperator, "id", instanceID)
 	object, err := p.repository.Get(ctx, types.ServiceInstanceType, byID)
@@ -56,20 +61,22 @@ func (p *checkInstanceOwnerPlugin) assertOwner(req *web.Request, next web.Handle
 	}
 	instance := object.(*types.ServiceInstance)
 
-	if instance.Labels == nil {
-		return nil, fmt.Errorf("could not determine owner of service instance with id: %s", instanceID)
+	var instanceOwnerTenantID string
+	if instance.Labels != nil {
+		if tenantIDLabel, ok := instance.Labels[p.tenantIdentifier]; ok {
+			instanceOwnerTenantID = tenantIDLabel[0]
+		}
 	}
-	tenantIDLabel, ok := instance.Labels[p.tenantIdentifier]
-	if !ok {
-		return nil, fmt.Errorf("could not determine owner of service instance with id: %s", instanceID)
+	if instanceOwnerTenantID == "" {
+		log.C(ctx).Infof("Tenant label for instance with id %s is missing.", instanceID)
+		return next.Handle(req)
 	}
-	instanceOwnerTenantID := tenantIDLabel[0]
 
 	if instanceOwnerTenantID != callerTenantID {
 		log.C(ctx).Errorf("Instance owner %s is not the same as the caller %s", instanceOwnerTenantID, callerTenantID)
 		return nil, &util.HTTPError{
 			ErrorType:   "NotFound",
-			Description: "Service instance not found or not visible to the current user.",
+			Description: fmt.Sprintf("could not find such %s", string(types.ServiceInstanceType)),
 			StatusCode:  http.StatusNotFound,
 		}
 	}
--- a/api/plugins/check_instance_owner_plugin.go
+++ b/api/plugins/check_instance_owner_plugin.go
@@ -1,8 +1,15 @@
 package plugins
 
 import (
+	"github.com/Peripli/service-manager/pkg/log"
+	"github.com/Peripli/service-manager/pkg/query"
+	"github.com/Peripli/service-manager/pkg/types"
+	"github.com/Peripli/service-manager/pkg/util"
 	"github.com/Peripli/service-manager/pkg/web"
 	"github.com/Peripli/service-manager/storage"
+	"github.com/tidwall/gjson"
+	"net/http"
+	"strings"
 )
 
 const CheckInstanceOwnerPluginName = "CheckInstanceOwnerPlugin"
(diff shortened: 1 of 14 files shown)


api/plugins/check_instance_owner_plugin.go
```diff
@@ -0,0 +1,69 @@
...
+func (p *checkInstanceOwnerPlugin) Name() string {
+	return CheckInstanceOwnerPluginName
+}
+
+// Bind intercepts bind requests and check if the instance owner is the same as the one requesting the bind operation
+func (p *checkInstanceOwnerPlugin) Bind(req *web.Request, next web.Handler) (*web.Response, error) {
+	ctx := req.Context()
+	callerTenantID := gjson.GetBytes(req.Body, "context."+p.tenantIdentifier).String()
+	pathSegments := strings.Split(req.URL.Path, "/")
+	instanceID := pathSegments[len(pathSegments)-3] // /v2/service_instances/:instance_id/service_bindings/:binding_id
```
pankrator: You can use instead: `req.PathParams["instance_id"]`
this is the route we attach: ` baseURL + "/v2/service_instances/{instance_id}/service_bindings/{binding_id}"`

dpanayotov:
I agree multi tenancy check makes sense when enabled, but platform_id check makes sense in both cases (MT and not).

api/plugins/check_instance_owner_plugin.go
```diff
@@ -0,0 +1,69 @@
...
+	byID := query.ByField(query.EqualsOperator, "id", instanceID)
+	object, err := p.repository.Get(ctx, types.ServiceInstanceType, byID)
+	if err != nil {
+		if err != util.ErrNotFoundInStorage {
+			return nil, util.HandleStorageError(err, string(types.ServiceInstanceType))
```
NickyMateev: We should not fail if the instance is not found. It's quite possible for a CC instance to have existed before we started storing instances.

You should just call the next handler.

api/plugins/check_instance_owner_plugin.go
```diff
@@ -0,0 +1,69 @@
...
+		} else {
+			return next.Handle(req)
+		}
+	}
+	instance := object.(*types.ServiceInstance)
+
+	var instanceOwnerTenantID string
+	if instance.Labels != nil {
```
NickyMateev: Not sure we need this check as we always ensure that an instance is labelled, but I do understand that we want to be on the safe side.

What do you think about checking if the `instance.Labels` are nil, and if they are, then returning some error. This way we could catch if bugs which might have caused us not to label the instance. The error could say something like: _"Could not determine service instance owner"_

This would be returned even if there are labels, but the `tenantIdentifier` key is missing.

api/plugins/check_instance_owner_plugin.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+}
+
+// Name returns the name of the plugin
+func (p *checkInstanceOwnerPlugin) Name() string {
+	return CheckInstanceOwnerPluginName
+}
+
+// Bind intercepts bind requests and check if the instance owner is the same as the one requesting the bind operation
+func (p *checkInstanceOwnerPlugin) Bind(req *web.Request, next web.Handler) (*web.Response, error) {
```
NickyMateev: Validation should be done for all binding related operations (not just for Bind) - bind, unbind, fetch binding, poll last binding operation.

api/plugins/check_instance_owner_plugin.go
DimitarPetrov: Yes, this check does exactly this, if the error is different from NotFound process it, else proceed with the next handler.

DimitarPetrov:
> I agree multi tenancy check makes sense when enabled, but platform_id check makes sense in both cases (MT and not).

What do you mean by `platform_id check`?

dpanayotov:
@DimitarPetrov each instance has platform_id field. Each OSB request comes with credentials mapped to a platform. These 2 must match.

api/plugins/check_instance_owner_plugin.go
NickyMateev: My mistake, sorry :D

api/plugins/check_instance_owner_plugin.go
```diff
@@ -0,0 +1,78 @@
...
+		return nil, util.HandleStorageError(err, string(types.ServiceInstanceType))
+	}
+	instance := object.(*types.ServiceInstance)
+
+	if instance.Labels == nil {
+		return nil, fmt.Errorf("could not determine owner of service instance with id: %s", instanceID)
+	}
+	tenantIDLabel, ok := instance.Labels[p.tenantIdentifier]
+	if !ok {
+		return nil, fmt.Errorf("could not determine owner of service instance with id: %s", instanceID)
```
KirilKabakchiev: this will be a problem for instances in whitelisted orgs

api/plugins/check_instance_owner_plugin.go
```diff
@@ -0,0 +1,78 @@
...
+	object, err := p.repository.Get(ctx, types.ServiceInstanceType, byID)
+	if err != nil {
+		if err == util.ErrNotFoundInStorage {
+			return next.Handle(req)
+		}
+		return nil, util.HandleStorageError(err, string(types.ServiceInstanceType))
+	}
+	instance := object.(*types.ServiceInstance)
+
+	if instance.Labels == nil {
```
KirilKabakchiev: this will be a problem for instances in whitelisted orgs

api/plugins/check_instance_owner_plugin.go
```diff
@@ -0,0 +1,78 @@
...
+	if !ok {
+		return nil, fmt.Errorf("could not determine owner of service instance with id: %s", instanceID)
+	}
+	instanceOwnerTenantID := tenantIDLabel[0]
+
+	if instanceOwnerTenantID != callerTenantID {
+		log.C(ctx).Errorf("Instance owner %s is not the same as the caller %s", instanceOwnerTenantID, callerTenantID)
+		return nil, &util.HTTPError{
+			ErrorType:   "NotFound",
+			Description: "Service instance not found or not visible to the current user.",
```
KirilKabakchiev: prolly do something similar to the descriptions in other not found errors - fmt.Sprintf("could not find such %s", string(types.ServiceInstanceType))

api/plugins/check_instance_owner_plugin.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// UpdateService intercepts update service instance requests and check if the instance owner is the same as the one requesting the operation
+func (p *checkInstanceOwnerPlugin) UpdateService(req *web.Request, next web.Handler) (*web.Response, error) {
+	return p.assertOwner(req, next)
+}
+
+func (p *checkInstanceOwnerPlugin) assertOwner(req *web.Request, next web.Handler) (*web.Response, error) {
+	ctx := req.Context()
+	callerTenantID := gjson.GetBytes(req.Body, "context."+p.tenantIdentifier).String()
```
KirilKabakchiev: do we need more explicit handling/logs for the cases when context is missing or tenantIdentifier is missing from context?

api/plugins/check_platform_id_plugin.go
```diff
@@ -0,0 +1,91 @@
...
+
+// PollBinding intercepts poll binding operation requests and check if the instance is in the platform from where the request comes
+func (p *checkPlatformIDPlugin) PollBinding(req *web.Request, next web.Handler) (*web.Response, error) {
+	return p.assertPlatformID(req, next)
+}
+
+func (p *checkPlatformIDPlugin) assertPlatformID(req *web.Request, next web.Handler) (*web.Response, error) {
+	ctx := req.Context()
+	user, _ := web.UserFromContext(ctx)
+	platform := &types.Platform{}
```
KirilKabakchiev: you can do a platform.Validate here to make sure the user from context is what you expect it to be

pkg/sm/sm.go
```diff
@@ -155,8 +153,9 @@ func New(ctx context.Context, cancel context.CancelFunc, e env.Environment, cfg
 		cfg:                 cfg,
 	}
 
-	smb.RegisterPlugins(plugins.NewCatalogFilterByVisibilityPlugin(interceptableRepository))
-	smb.RegisterPlugins(osb.NewStoreServiceInstancesPlugin(interceptableRepository))
+	smb.RegisterPlugins(osb.NewCatalogFilterByVisibilityPlugin(interceptableRepository))
+	smb.RegisterPluginsBefore(osb.CheckInstanceOwnerPluginName, osb.NewStoreServiceInstancesPlugin(interceptableRepository))
```
KirilKabakchiev: we now make calling EnableMultitenancy mandatory - otherwise SM will panic on bootstrap ( not finding the checkInstanceOwner plugin). This defeats the purpose for having EnableMultitenancy as a separate method.

PR #376, opened 2019-12-27
Handle 410 Gone on Last Operation requests

NickyMateev:
# Handle 410 Gone on Last Operation requests

## Motivation

As per OSB spec a response to a last operation call can return 410 Gone: https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-1

In the case of an async delete operation, this has to be interpreted as the instances has been deleted in the original service broker.

Currently we do not handle this scenario as we're always expecting a 200 OK response: https://github.com/Peripli/service-manager/blob/8598d3c656897e687b8b443ef256839e0792f42a/api/osb/store_instances_plugin.go#L354

The consequences are that we have orphan instances in our DB, which furthermore complicate things as those instances block the deletion(or catalog update) of the corresponding service broker in SM. For such cases, we have to go directly in the DB and delete those instances (until we have a DELETE API).

## Approach

Handle the case where 410 Gone is returned on last operation requests.

## Pull Request status

* [x] Initial implementation
* [x] Integration tests

### DIFF
--- a/api/osb/store_instances_plugin.go
+++ b/api/osb/store_instances_plugin.go
@@ -351,7 +351,7 @@ func (ssi *StoreServiceInstancePlugin) PollInstance(request *web.Request, next w
 		return nil, err
 	}
 
-	if response.StatusCode != http.StatusOK {
+	if response.StatusCode != http.StatusOK || response.StatusCode != http.StatusGone {
 		return response, nil
 	}
 
@@ -382,6 +382,13 @@ func (ssi *StoreServiceInstancePlugin) PollInstance(request *web.Request, next w
 		}
 
 		operationFromDB := op.(*types.Operation)
+		if response.StatusCode == http.StatusGone {
+			if operationFromDB.Type != types.DELETE {
+				return nil
+			}
+			resp.State = types.SUCCEEDED
+		}
+
 		if operationFromDB.State != resp.State {
 			switch operationFromDB.Type {
 			case types.CREATE:
--- a/api/osb/store_instances_plugin.go
+++ b/api/osb/store_instances_plugin.go
@@ -351,7 +351,7 @@ func (ssi *StoreServiceInstancePlugin) PollInstance(request *web.Request, next w
 		return nil, err
 	}
 
-	if response.StatusCode != http.StatusOK || response.StatusCode != http.StatusGone {
+	if response.StatusCode != http.StatusOK && response.StatusCode != http.StatusGone {
 		return response, nil
 	}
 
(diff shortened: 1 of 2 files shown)


api/osb/store_instances_plugin.go
```diff
@@ -382,6 +382,13 @@ func (ssi *StoreServiceInstancePlugin) PollInstance(request *web.Request, next w
 		}
 
 		operationFromDB := op.(*types.Operation)
+		if response.StatusCode == http.StatusGone {
+			if operationFromDB.Type != types.DELETE {
+				return nil
```
pankrator: Shouldn't we proceed with the below logic or return here? I am not sure, looking at the code it seems to me that it should proceed

NickyMateev: A returned 410 Gone response will have no body, thus we cannot make the subsequent check to compare the state in the response with the last state in the DB. Also what would be the meaning of 410 Gone response to a CREATE/UPDATE operation?

And finally:

https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-1

"_Returning this while the Platform is polling for create or update operations SHOULD be interpreted as an invalid response and the Platform SHOULD continue polling._"

test/osb_test/poll_instance_last_operation_test.go
```diff
@@ -403,6 +415,18 @@ var _ = Describe("Get Service Instance Last Operation", func() {
 				})
 			})
 		})
+
+		Context("that returns 410 gone", func() {
+			BeforeEach(func() {
+				brokerServer.ServiceInstanceLastOpHandler = parameterizedHandler(http.StatusGone, `{}`)
+			})
+
+			It("returns 410", func() {
```
KirilKabakchiev: you could also verify that the operation/instance is not updated

api/osb/store_instances_plugin.go
```diff
@@ -382,6 +382,13 @@ func (ssi *StoreServiceInstancePlugin) PollInstance(request *web.Request, next w
 		}
 
 		operationFromDB := op.(*types.Operation)
+		if response.StatusCode == http.StatusGone {
```
KirilKabakchiev: i guess there is no easy way to avoid comparing status codes twice ? (once here and once on 354) ? One way I can think of would be to have a status code switch ontop and in each case process only the relevant stuff (for example 410 gone would process only DELETE op) but this would require extracting fetching op in a common func and probably a few other refactorings and I am not sure if the end result will be better

NickyMateev: Here's what I was thinking about: https://github.com/Peripli/service-manager/compare/fix-orphan-instances...fix-orphan-instances-v2

I think it gets kinda nasty with the triple switch and I'd rather have it like it is.

KirilKabakchiev: With the proposal you still check status code twice - you have to move the top level switch  to 354 -but yes if you prefer just keep it as is

PR #378, opened 2020-01-01
Introduce ClaimedAt operations property [async engine story]

NickyMateev:
# Introduce ClaimedAt property [async engine story]

## Note

This PR was created in the hopes to ease the discussion process of whether we need such a change. Until the team has agreed upon such an approach, this change only serves as a reference implementation.

## Motivation

The operation maintainer currently updates operations older than job_timeout to failed. It does so by making a diff between the current time and the created_at timestamp of the operation, but this is somewhat flawed - it's possible for an operation to be in the "job queue" and may have not been claimed at. Such operations are not stuck or orphans, they are just waiting for their turn to come when they're executed.

This PR proposes an introduction of a `claimed_at` property to operation entities, which will be set first thing when an operation (job) execution starts running.

This way the maintainer can differentiate which operations have indeed been stuck/orphaned due to SM node crash (for example) and which are waiting to be claimed.

### DIFF
--- a/operations/jobs.go
+++ b/operations/jobs.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Peripli/service-manager/pkg/types"
 	"github.com/Peripli/service-manager/pkg/util"
 	"github.com/Peripli/service-manager/storage"
+	"time"
 )
 
 // ExecutableJob represents a DB operation that has to be executed.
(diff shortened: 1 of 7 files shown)


NickyMateev:
Team agreed to this concept. Merging.

PR #379, opened 2020-01-01
Async engine without worker pool [async engine story]

NickyMateev:
# Async engine without worker pool [async engine story]

## Note

This PR was created in the hopes to ease the discussion process of whether we need such a change. Until the team has agreed upon such an approach, this change only serves as a reference implementation.

## Motivation

Motivation for this change was @dotchev 's idea of an async engine without a worker pool born out of the following questions:

> Why do we need the worker pool? Why separate pool per resource? How do we determine the optimal pool size? If we set the pool size too low, the jobs will be delayed unnecessarily.<br>
Jobs seem to be too coupled with the controller - we have ScheduleCreate, ScheduleUpdate and ScheduleDelete, which are very similar. <br>
The operation is maintained in different places. The controller creates the operation in db then it schedules the job. Each job updates the operation state. I would expect this to be done in a common place, e.g. the JobScheduler.<br>
I would expect to have a generic scheduler.ScheduleJob(job func(ctx Context) error) (string, error), which creates the operation in the db and returns its id. Then it waits for the job function to finish and updates the operation state.

Original comment: https://github.com/Peripli/service-manager/pull/373#pullrequestreview-335912222


NickyMateev:
Has a nasty import cycle that needs fixing..

NickyMateev:
Implemented into original branch with this commit: https://github.com/Peripli/service-manager/commit/53aac20cbcc810ac55833f87512fe241f037c113

PR #383, opened 2020-01-06
Get operation sub-API

KirilKabakchiev:


test/platform_test/platform_test.go
```diff
@@ -56,6 +57,7 @@ var _ = test.DescribeTestsFor(test.TestCase{
 			"zid": "tenantID",
 		},
 	},
+	SupportsAsyncOperations:                true,
```
NickyMateev: You can remove this and from visibilities also as we decided that only brokers will be able to processed asynchronously.

test/get.go
```diff
@@ -28,97 +38,229 @@ import (
...
+			var testResource common.Object
+			var testResourceID string
 
-		Context(fmt.Sprintf("Existing resource of type %s", t.API), func() {
-			createTestResourceWithAuth := func(auth *common.SMExpect) {
-				testResource = t.ResourceBlueprint(ctx, auth)
-				By(fmt.Sprintf("[SETUP]: Verifying that test resource %v is not empty", testResource))
-				Expect(testResource).ToNot(BeEmpty())
+			Context(fmt.Sprintf("Existing resource of type %s", t.API), func() {
+				createTestResourceWithAuth := func(auth *common.SMExpect) (common.Object, string) {
```
NickyMateev: `createTestResourceWithAuth` there's no auth

KirilKabakchiev:
If we decide to support it, since a bulk delete is esencially one transaction/one operation right now(there is no way to have a partically successful bulk delete), the correct way for the time being would be to create one operation in the database that has multiple resource_ids. We could strech a one-to-many relationship to a new table or we can use labels(model reaource_id as label with multiple values) or we can store the array as string with proper validation on the column (csv or json array)

api/base_controller.go
```diff
@@ -83,6 +84,13 @@ func (c *BaseController) Routes() []web.Route {
 			},
 			Handler: c.GetSingleObject,
 		},
+		{
+			Endpoint: web.Endpoint{
+				Method: http.MethodGet,
+				Path:   fmt.Sprintf("%s/{%s}%s/{%s}", c.resourceBaseURL, PathParamResourceID, web.OperationsURL, PathParamID),
```
pankrator: It is a bit confusing, because `PathParamResourceID` and `PathParamID` are with changed positions. 
My proposal is to put first `PathParamID` so it is the same as on the other endpoints. 
The other way is to change all other endpoints to use `PathParamResourceID`

api/base_controller.go
```diff
@@ -195,6 +203,30 @@ func (c *BaseController) GetSingleObject(r *web.Request) (*web.Response, error)
 	return util.NewJSONResponse(http.StatusOK, object)
 }
 
+// GetOperation handles the fetching of a single operation with the id specified for the specified resource
+func (c *BaseController) GetOperation(r *web.Request) (*web.Response, error) {
```
pankrator: Should we implement ListOperation for given objectID? For example: operations for all brokers?

api/filters/plans_filter_by_visibility.go
```diff
@@ -50,7 +50,7 @@ func (vf *PlanFilterByVisibility) FilterMatchers() []web.FilterMatcher {
 	return []web.FilterMatcher{
 		{
 			Matchers: []web.Matcher{
-				web.Path(web.ServicePlansURL + "/**"),
+				web.Path(web.ServicePlansURL + "/*"),
```
pankrator: Why is the second star removed? This filter will not be attached for getting single plan. Same goes for others.

KirilKabakchiev: Well according to docs it should be attached - a single star means up to the next separator /. Its removed because we dont want these filters on the operations subapi

api/base_controller.go
KirilKabakchiev: Not needed for the async engine part (it will be introduced for admin purposes eventually when we get there i guess ?). And for admin purposes i see it as a too level list operations api secured with its own operations scopes as we dont wqnt to hand it out in the location header

api/filters/plans_filter_by_visibility.go
pankrator: yes, I missed the part that we don't need them on the operations subapi

pankrator:
About bulk delete:

I think it would be practically inefficient to store many IDs in one db table field. May be it will be better to put other IDs as labels. This will make uglier to query for operations for given resource, but worth trying I think.

test/platform_test/platform_test.go
pankrator: Is this because we don't have a mechanism to return the platform credentials other way?

api/base_controller.go
```diff
@@ -195,6 +203,30 @@ func (c *BaseController) GetSingleObject(r *web.Request) (*web.Response, error)
...
+
+	byOperationID := query.ByField(query.EqualsOperator, "id", operationID)
+	byObjectID := query.ByField(query.EqualsOperator, "resource_id", objectID)
+	var err error
+	ctx, err = query.AddCriteria(ctx, byObjectID, byOperationID)
+	if err != nil {
+		return nil, err
+	}
+	criteria := query.CriteriaForContext(ctx)
+	operation, err := c.repository.Get(ctx, types.OperationType, criteria...)
```
NickyMateev: You have to strip the tenant identifier label if it's in the current criteria, otherwise the DB operation will fail with non-existant entity.

KirilKabakchiev: Well I actually rely on it to be there so that calls with tenant access cannot see global operations. See how operations for tenant scoped resources are created in the authz tests for reference

NickyMateev: Yeah, sure, but it still breaks the DB Get, so it has to be stripped away before making the DB operation

KirilKabakchiev: Its needed in order to disallow tenant access to see operations for global resources

PR #384, opened 2020-01-06
Get Operation API (top-level)

KirilKabakchiev:


api/filters/basic_authentication.go
```diff
@@ -138,5 +138,11 @@ func basicAuthnMatchers() []web.FilterMatcher {
 				web.Path(web.NotificationsURL + "/**"),
 			},
 		},
+		{
+			Matchers: []web.Matcher{
```
pankrator: Do we need the operations API to be available with basic authentication?

KirilKabakchiev: I dont think so - i mainly wanted to showcase the additional effort required in the internal repo (filter and authz tests) - i will clean up if we pick this approach

PR #388, opened 2020-01-14
Persist dashboard_url for Service Instances

NickyMateev:
# Persist dashboard_url for Service Instances

## Motivation

The dashboard_url property is a top-level one which should be returned when fetching instances.

## Pull Request status

* [x] Initial implementation
* [x] Unit tests
* [x] Integration tests

### DIFF
--- a/storage/postgres/migrations/20200114101000_dashboard_url.up.sql
+++ b/storage/postgres/migrations/20200114101000_dashboard_url.up.sql
@@ -1,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE service_instances ADD COLUMN dashboard_url varchar(100);
+
+COMMIT;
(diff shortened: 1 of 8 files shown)


storage/postgres/migrations/20200114101000_dashboard_url.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE service_instances ADD COLUMN dashboard_url varchar(100);
```
pankrator: Not sure that 100 will be enough for all services. Check whether OSB spec says something. At first glance it does not say anything about length as far as I checked

NickyMateev: It's 16000 chars in CC DB, I can make it the same.

PR #389, opened 2020-01-14
Return last operation for entity

pankrator:
# Motivation

Last operation for resource is needed for the CLI to display properly the state of a resource

## Approach

Return last operation when query parameter `display_op` is true and entity supports async

## Pull Request status

* [x] Initial implementation
* [x ] Unit tests

### DIFF
--- a/api/base_controller.go
+++ b/api/base_controller.go
@@ -283,13 +283,8 @@ func (c *BaseController) GetSingleObject(r *web.Request) (*web.Response, error) 
 	log.C(ctx).Debugf("Getting %s with id %s", c.objectType, objectID)
 
 	byID := query.ByField(query.EqualsOperator, "id", objectID)
-	var err error
-	ctx, err = query.AddCriteria(ctx, byID)
-	if err != nil {
-		return nil, err
-	}
 	criteria := query.CriteriaForContext(ctx)
-	object, err := c.repository.Get(ctx, c.objectType, criteria...)
+	object, err := c.repository.Get(ctx, c.objectType, append(criteria, byID)...)
 	if err != nil {
 		return nil, util.HandleStorageError(err, c.objectType.String())
 	}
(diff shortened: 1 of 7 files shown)


pkg/types/interfaces.go
```diff
@@ -38,6 +38,11 @@ type Secured interface {
 	GetCredentials() *Credentials
 }
 
+type Operationable interface {
```
NickyMateev: Every type in this file has godoc, might as well continue to convention

api/base_controller.go
```diff
@@ -293,6 +294,12 @@ func (c *BaseController) GetSingleObject(r *web.Request) (*web.Response, error)
 	}
 
 	stripCredentials(ctx, object)
+	displayOp := r.URL.Query().Get("display_op")
```
NickyMateev: How about simply `last_op`, it's a well-known term, and also whichever you choose, extract it as constant so that we don't run into yesterday's problem with hardcoded parameter names in filters which are outdated.

api/base_controller.go
```diff
@@ -458,6 +465,32 @@ func (c *BaseController) PatchObject(r *web.Request) (*web.Response, error) {
 	return util.NewJSONResponse(http.StatusOK, object)
 }
 
+func attachLastOperation(ctx context.Context, objectID string, object types.Object, r *web.Request, repository storage.Repository) error {
+	operationable, ok := object.(types.Operationable)
+	if ok {
```
NickyMateev:
```go
if operationable, ok := object.(types.Operationable); ok {
...
}
```

pkg/types/service_offering.go
```diff
@@ -46,6 +46,8 @@ type ServiceOffering struct {
 	CatalogName string `json:"catalog_name"`
 
 	Plans []*ServicePlan `json:"plans"`
+
+	LastOperation *Operation `json:"last_operation,omitempty"`
```
NickyMateev: Why do we need last operation for offerings and plans?

test/get.go
```diff
@@ -67,6 +67,29 @@ func DescribeGetTestsfor(ctx *common.TestContext, t TestCase, responseMode Respo
...
+							})
+						}
+
+						if !t.SupportsAsyncOperations {
+							Context("when resource does not support async and query param display_op is true", func() {
+								It("returns error", func() {
+									ctx.SMWithOAuth.GET(fmt.Sprintf("%s/%s", t.API, testResourceID)).
+										WithQuery("display_op", "true").
+										Expect().
+										Status(http.StatusBadRequest).JSON().Object().Value("description").Equal("last operation is not supported for type Platform")
```
NickyMateev: these are generic tests, remove **Platform**. Also, how did the broker tests pass with expectation for **Platform** in string?

pankrator: ops, this is mistake, the tests does not pass even

pkg/types/service_offering.go
pankrator: Not really sure, which entity support async, but I will remove this

NickyMateev:
🔨

dpanayotov:
It says that the client supports OSB v2.11,12,13 Will it be enough in our case? v2.15 is already released. Also the import path is `github.com/kubernetes-sigs/go-open-service-broker-client`as it's moved under k8s-sigs

PR #391, opened 2020-01-14
Introduce service binding GET apis

georgifarashev:
# Introduce service binding GET apis

## Status

* [x] Storage layer
* [x] API layer
* [x] Tests


pkg/types/service_binding.go
```diff
@@ -0,0 +1,100 @@
...
+	Secured           `json:"-"`
+	Name              string          `json:"name"`
+	ServiceInstanceID string          `json:"service_instance_id"`
+	SyslogDrainURL    string          `json:"syslog_drain_url,omitempty"`
+	RouteServiceURL   string          `json:"route_service_url,omitempty"`
+	VolumeMounts      json.RawMessage `json:"volume_mounts,omitempty"`
+	Endpoints         json.RawMessage `json:"endpoints,omitempty"`
+	Context           json.RawMessage `json:"-"`
+	BindResource      json.RawMessage `json:"-"`
+	Credentials       string          `json:"credentials"`
```
pankrator: May be it will be better for this to be of type RawMessage or []byte because we don't know what the broker will put there

pankrator: As discussed this is a limitation of the storage layer, because it cannot Scan into RawMessage. Also because it is encrypted in the database it cannot be any other db field, but byte array

pkg/types/service_binding.go
```diff
@@ -0,0 +1,100 @@
...
+// ServiceBinding struct
+type ServiceBinding struct {
+	Base
+	Secured           `json:"-"`
+	Name              string          `json:"name"`
+	ServiceInstanceID string          `json:"service_instance_id"`
+	SyslogDrainURL    string          `json:"syslog_drain_url,omitempty"`
+	RouteServiceURL   string          `json:"route_service_url,omitempty"`
+	VolumeMounts      json.RawMessage `json:"volume_mounts,omitempty"`
+	Endpoints         json.RawMessage `json:"endpoints,omitempty"`
```
pankrator: I suppose that endpoints might contain URLs with credentials in it. Should we encrypt them also?

pkg/types/service_binding.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return true
+}
+
+// Validate implements InputValidator and verifies all mandatory fields are populated
+func (e *ServiceBinding) Validate() error {
+	if util.HasRFC3986ReservedSymbols(e.ID) {
+		return fmt.Errorf("%s contains invalid character(s)", e.ID)
+	}
+	if e.Name == "" {
```
pankrator: Service instance id should also be checked for existence

storage/interceptors/secured_generate_credentials_interceptor.go
```diff
@@ -50,7 +51,11 @@ func (c *generateCredentialsInterceptor) AroundTxCreate(h storage.InterceptCreat
 			log.C(ctx).Error("Could not generate credentials for platform")
 			return nil, err
 		}
-		(obj.(types.Secured)).SetCredentials(credentials)
+		platform, ok := obj.(*types.Platform)
```
pankrator: This interceptor is assigned only to platform type, but the interceptor is generic in order to allow generation of credentials for arbitrary type.
One option is to make this interceptor only for platforms (name of the filter).
Another option is to introduce another interface to set credentials of certain types object and here check for this interface implementation. If implemented set the credentials of the object via the same interface.

georgifarashev: I will rename the filter so it is platform specific

pkg/types/service_binding.go
georgifarashev: This is a good point. This should be investigated.

api/service_binding_controller.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/Peripli/service-manager/pkg/types"
+	"github.com/Peripli/service-manager/pkg/web"
+)
+
+// ServiceBindingController implements api.Controller by providing service instances API logic
```
NickyMateev: "providing service bindings API logic"

test/service_binding_test/service_binding_test.go
```diff
@@ -0,0 +1,102 @@
...
+		_, err = ctx.SMRepository.Update(context.Background(), sb, patchLabels)
+		if err != nil {
+			Fail(fmt.Sprintf("unable to update resource %s: %s", resourceType, err))
+		}
+	},
+	AdditionalTests: func(ctx *common.TestContext) {},
+})
+
+func blueprint(ctx *common.TestContext, auth *common.SMExpect, _ bool) common.Object {
+	_, serviceInstance := service_instance.Prepare(ctx, ctx.TestPlatform.ID, "", fmt.Sprintf(`{"%s":"%s"}`, TenantIdentifier, TenantValue))
```
NickyMateev: You can use the Create Instances API now

PR #392, opened 2020-01-14
Introduce Service Instance C/U/D APIs

NickyMateev:
# Introduce Service Instance C/U/D APIs

## Motivation

SM has to be able to provision and manage instances directly.

## Pull Request status

* [x] Initial implementation
* [x] Integration tests


test/service_instance_test/service_instance_test.go
```diff
@@ -50,7 +52,7 @@ const (
 var _ = test.DescribeTestsFor(test.TestCase{
 	API: web.ServiceInstancesURL,
 	SupportedOps: []test.Op{
-		test.Get, test.List,
+		test.Get, test.List, test.Delete, test.DeleteList, test.Patch,
```
KirilKabakchiev: Post missing. Also post and patch tests are not parameterized

test/service_instance_test/service_instance_test.go
```diff
@@ -65,8 +67,8 @@ var _ = test.DescribeTestsFor(test.TestCase{
 	ResourceType:                           types.ServiceInstanceType,
 	SupportsAsyncOperations:                true,
 	DisableTenantResources:                 true,
-	ResourceBlueprint:                      blueprint,
-	ResourceWithoutNullableFieldsBlueprint: blueprint,
+	ResourceBlueprint:                      blueprint(),
+	ResourceWithoutNullableFieldsBlueprint: blueprint(),
```
KirilKabakchiev: This blueprint should return the object with only mandatory stuff set as far as i remember

NickyMateev: Yes, but there's no nullable properties that a client can ignore for instances, like there is for brokers and platforms (`description` property).

api/filters/service_instance_filter.go
```diff
@@ -0,0 +1,81 @@
...
+	return ServiceInstanceValidationFilterName
+}
+
+func (*ServiceInstanceValidationFilter) Run(req *web.Request, next web.Handler) (*web.Response, error) {
+	platformID := gjson.GetBytes(req.Body, platformIDProperty).Str
+
+	if platformID != "" {
+		return nil, &util.HTTPError{
+			ErrorType:   "BadRequest",
+			Description: fmt.Sprintf("Providing %s property during provisioning/updating of a service instance is forbidden", platformIDProperty),
```
dpanayotov: What if I provide `service-manager`? Should this be allowed? We could just ignore whatever is provided and overwrite it with `service-manager`?

NickyMateev: addressed

PR #394, opened 2020-01-14
Introduce flexible OSB Client constructor

NickyMateev:
# Introduce flexible OSB Client constructor

## Motivation

Will be needed for OSB interceptor which will call broker endpoints.

## Pull Request status

* [x] Initial implementation

### DIFF
--- a/pkg/util/osb.go
+++ b/pkg/util/osb.go
@@ -2,6 +2,7 @@ package util
 
 import osbc "github.com/pmorie/go-open-service-broker-client/v2"
 
+// NewOSBCient returns a function which constructs an OSB client based on a provided configuration
 func NewOSBClient(skipSsl bool) osbc.CreateFunc {
 	return func(configuration *osbc.ClientConfiguration) (osbc.Client, error) {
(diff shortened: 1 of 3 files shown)


pkg/util/osb.go
```diff
@@ -0,0 +1,10 @@
+package util
+
+import osbc "github.com/pmorie/go-open-service-broker-client/v2"
+
+func NewOSBClient(skipSsl bool) osbc.CreateFunc {
```
pankrator: Add godoc.

dpanayotov:
It says that the client supports OSB v2.11,12,13 Will it be enough in our case? v2.15 is already released. Also the import path is `github.com/kubernetes-sigs/go-open-service-broker-client`as it's moved under k8s-sigs

KirilKabakchiev:
We should double check but since this will be the SM as a platform flow, it might be ok ?  brokers on 2.15 should work fine for calls with a 2.13 client side i think?

pkg/util/osb.go
NickyMateev: Of course :D

PR #395, opened 2020-01-15
Log error when no plan visibility is found

dotchev:
## Motivation

In case the service plan is not visible, service provisioning fails and we could see only this error in the log
```
HTTPError: could not find such types.ServicePlan
```
The actual reason was not clear.

## Approach

Log explicitly that the plan is not visible.

## Pull Request status

* [x] Initial implementation

### DIFF
--- a/api/osb/check_visibility_plugin.go
+++ b/api/osb/check_visibility_plugin.go
@@ -2,7 +2,6 @@ package osb
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http"
 
 	"github.com/Peripli/service-manager/pkg/log"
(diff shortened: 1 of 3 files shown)


api/osb/check_visibility_plugin.go
```diff
@@ -114,23 +113,25 @@ func (p *checkVisibilityPlugin) checkVisibility(req *web.Request, next web.Handl
 				}
 			}
 		}
+		log.C(ctx).Errorf("Service plan %v is not visible on platform %v", planID, platform.ID)
```
pankrator: May be log also the plan name, as it will be easier to find out is this the plan you are looking for. The same for the other log message

pankrator: Now I saw, that we don't have the plan name here...

PR #398, opened 2020-01-16
Fix instance dashboard url serialization

georgifarashev:
# Fix instance dashboard url serialization

Service instance dashboard url is serialized from db into string.
There are some cases where this is actually null in the database (after db migration of existing service instances).
Therefore NullString should be used.


KirilKabakchiev:
can we just add DEFAULT '' in the sql migration and use normal string in the code ?

georgifarashev:
> can we just add DEFAULT '' in the sql migration and use normal string in the code ?

If there are existing records in the db setting default values will not apply to them. It is possible to set all dasboard_urls to '' and then set the default but I don't see that much benefit in doing so.

KirilKabakchiev:
I think not null default “ will also set the value in the old rows - initially i thought thats how we do it everywhere (so i wanted to do it for conaiatency) but then i saw we have both approaches in the code so i am fine with either.

PR #402, opened 2020-01-21
Add checkBrokerCredentials filter

ataleksandrov:
# Pull Request Template

## Motivation

There may be cases when a user updates an existing service broker and provides only URL without the credentials, thus getting the real credentials of the broker.

## Approach

Introduce checkBrokerCredentials filter.

## Pull Request status

* [x] Initial implementation
* [x] Integration tests

### DIFF
--- a/api/filters/checkBrokerCredentials.go
+++ b/api/filters/checkBrokerCredentials.go
@@ -25,10 +25,10 @@ func (*CheckBrokerCredentialsFilter) Name() string {
 func (*CheckBrokerCredentialsFilter) Run(req *web.Request, next web.Handler) (*web.Response, error) {
 	fields := gjson.GetManyBytes(req.Body, "broker_url", fmt.Sprintf(credentialsPath, "username"), fmt.Sprintf(credentialsPath, "password"))
 
-	if fields[0].Exists() && !fields[1].Exists() && !fields[2].Exists() {
+	if fields[0].Exists() && (!fields[1].Exists() || !fields[2].Exists()) {
 		return nil, &util.HTTPError{
 			ErrorType:   "BadRequest",
-			Description: "Updating an URl of a broker requires its basic credentials",
+			Description: "Updating an URL of a broker requires its basic credentials",
 			StatusCode:  http.StatusBadRequest,
 		}
 	}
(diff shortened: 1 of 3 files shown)


api/filters/checkBrokerCredentials.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (*CheckBrokerCredentialsFilter) Name() string {
+	return CheckBrokerCredentialsFilterName
+}
+
+func (*CheckBrokerCredentialsFilter) Run(req *web.Request, next web.Handler) (*web.Response, error) {
+	fields := gjson.GetManyBytes(req.Body, "broker_url", fmt.Sprintf(credentialsPath, "username"), fmt.Sprintf(credentialsPath, "password"))
+
+	if fields[0].Exists() && !fields[1].Exists() && !fields[2].Exists() {
```
DimitarPetrov: With check like this if for example one provides URL and Username the respective expression will be:
`true && false && ...` Therefore the evaluation would stop when it sees that an username is provided even tough that password may be missing.
Something like `fields[0].Exists() && !(fields[1].Exists && fields[2].Exists)`

test/broker_test/broker_test.go
```diff
@@ -804,6 +812,65 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+
+							assertInvocationCount(brokerServer.CatalogEndpointRequests, 0)
+
+							ctx.SMWithOAuth.GET(web.ServiceBrokersURL+"/"+brokerID).
+								Expect().
+								Status(http.StatusOK).
+								JSON().Object().
+								ContainsMap(expectedBrokerResponse).
+								Keys().NotContains("services", "credentials")
+						}
```
DimitarPetrov: This checks seems unnecessary. With this said the function is unnecessary itself and tests could just make a PATCH call and expect BadRequest.

test/broker_test/broker_test.go
```diff
@@ -804,6 +812,65 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+						Context("password is missing", func() {
+							BeforeEach(func() {
+								updatedBrokerJSON["credentials"] = common.Object{
+									"basic": common.Object{
+										"username": "a",
+									},
+								}
+							})
+
+							It("returns 400", func() {
```
DimitarPetrov: Same as the test in 841 line.

PR #404, opened 2020-01-21
Allow to create instances if there is public visibility for the plan

pankrator:
# Pull Request Template

## Motivation

- When plan has public SM visibility, one should be able to create instances
- When plan does not have any visibility and OAuth without tenant is used, one should not be able to create instances

### DIFF
--- a/api/filters/check_instance_visibility_filter.go
+++ b/api/filters/check_instance_visibility_filter.go
@@ -17,6 +17,7 @@
 package filters
 
 import (
+	"github.com/Peripli/service-manager/pkg/util/slice"
 	"github.com/Peripli/service-manager/pkg/log"
 	"github.com/Peripli/service-manager/pkg/query"
 	"github.com/Peripli/service-manager/pkg/types"
@@ -92,10 +93,8 @@ func (f *serviceInstanceVisibilityFilter) Run(req *web.Request, next web.Handler
 		}
 		tenantLabels, ok := v.Labels[f.tenantIdentifier]
 		if ok {
-			for _, tl := range tenantLabels {
-				if tl == tenantID {
-					return next.Handle(req)
-				}
+			if slice.StringsAnyEquals(tenantLabels, tenantID) {
+				return next.Handle(req)
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


api/filters/check_instance_visibility_filter.go
```diff
@@ -60,32 +60,47 @@ func (f *serviceInstanceVisibilityFilter) Run(req *web.Request, next web.Handler
...
-
-	_, err := f.repository.Get(ctx, types.VisibilityType, criteria...)
-	if err != nil {
-		if err == util.ErrNotFoundInStorage {
-			return nil, &util.HTTPError{
-				ErrorType:   "NotFound",
-				Description: "could not find such service plan",
-				StatusCode:  http.StatusNotFound,
+	visibilities := list.(*types.Visibilities).Visibilities
+	for _, v := range visibilities {
```
NickyMateev: Do we have to have the loop primitive? In reality there's either going to be 0 visibilities or 1 (public or one with platform_id=SM)

Can we just return visibility error if 0, and otherwise get the 0-th index visibility and directly work on it? Because it's somewhat confusing having this loop, at first I thought that there would indeed be lots of visibilities being traversed

api/filters/check_instance_visibility_filter.go
```diff
@@ -60,32 +60,47 @@ func (f *serviceInstanceVisibilityFilter) Run(req *web.Request, next web.Handler
...
+		if len(v.PlatformID) == 0 {
+			return next.Handle(req)
+		}
+		tenantLabels, ok := v.Labels[f.tenantIdentifier]
+		if ok {
+			for _, tl := range tenantLabels {
```
NickyMateev: use `slice.StringsContaining`

api/filters/check_instance_visibility_filter.go
```diff
@@ -86,17 +86,16 @@ func (f *serviceInstanceVisibilityFilter) Run(req *web.Request, next web.Handler
...
-	for _, v := range visibilities {
-		if len(v.PlatformID) == 0 {
+	// There may be at most one visibility for the query - for SM platform or public for this plan
+	visibility := list.ItemAt(0).(*types.Visibility)
+	if len(visibility.PlatformID) == 0 { // public visibility
+		return next.Handle(req)
+	}
+	tenantLabels, ok := visibility.Labels[f.tenantIdentifier]
+	if ok {
+		if slice.StringsAnyEquals(tenantLabels, tenantID) {
```
NickyMateev: if ok && slice.StringsAnyEquals(tenantLabels, tenantID)

PR #406, opened 2020-01-23
Reject requests with no tenant for service and bindings

pankrator:
# Pull Request Template

## Motivation

* Return bad request for requests with no tenant identifier for instances and bindings

### DIFF
--- a/test/test.go
+++ b/test/test.go
@@ -80,10 +80,11 @@ type TestCase struct {
 
 	MultitenancySettings   *MultitenancySettings
 	DisableTenantResources bool
+	StrictlyTenantScoped   bool
 
 	ResourceBlueprint                      func(ctx *common.TestContext, smClient *common.SMExpect, async bool) common.Object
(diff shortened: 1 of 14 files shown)


test/test.go
```diff
@@ -80,10 +80,11 @@ type TestCase struct {
 
 	MultitenancySettings   *MultitenancySettings
 	DisableTenantResources bool
+	StrictlyTenantScoped   bool
```
NickyMateev: I'm not sure if we can completely get rid of this flag as I see there are some tests that are some generic tests executed only if StrictlyTenantScoped is true.

But there were numerous places where we do the exact same request just a different SMExpect (either SMWithOAuth, or SMWithAuthForTenant) - what do you think about extending the TestCase struct to contain also the SMExpect with which the tests should be executed? This way we can remove some of those if/else checks. Not sure if this can help somehow with the StrictlyTenantScoped flag.

PR #407, opened 2020-01-23
Return better message when plan is not visible

dotchev:
## Motivation

If a client tries to access a plan that is not visible in the current context, he gets the error
`could not find such service plan`
which is confusing

## Approach

Now we return error
`service plan not found or not accessible`
which describes better the reason

## Pull Request status

* [x] Initial implementation

### DIFF
--- a/api/osb/check_visibility_plugin.go
+++ b/api/osb/check_visibility_plugin.go
@@ -16,9 +16,9 @@ import (
 const CheckVisibilityPluginName = "CheckVisibilityPlugin"
 
 var errPlanNotAccessible = &util.HTTPError{
-	ErrorType:   "ServicePlanNotAccessible",
-	Description: "service plan is not accessible",
-	StatusCode:  http.StatusForbidden,
+	ErrorType:   "ServicePlanNotFound",
+	Description: "service plan not found or not accessible",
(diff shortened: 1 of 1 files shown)


api/osb/check_visibility_plugin.go
```diff
@@ -15,6 +15,12 @@ import (
 
 const CheckVisibilityPluginName = "CheckVisibilityPlugin"
 
+const errPlanNotAccessible = &util.HTTPError{
+	ErrorType:   "ServicePlanNotAccessible",
+	Description: "service plan is not accessible",
+	StatusCode:  http.StatusForbidden,
```
KirilKabakchiev: im not sure about this - for a client that has no visibility this practically means that he should probably not know that such plan exists

dotchev: We discussed this with @dpanayotov and decided to change the message to avoid confusion as clients usually do know that this plan exists. Sometimes they enable service access manually in CF and try to create an instance.

dpanayotov: Maybe it's better to say `service either not found or not accessible` with status 404? Isn't it an issue to directly say that there's such a plan but you don't have access?

PR #408, opened 2020-01-24
Public plan visibility composition based on plan supported platforms + broker platform notifications

NickyMateev:
# Public plan visibility composition based on plan supported platforms + broker platform notifications

## Motivation

We need to distinguish public plans which plans are supported in which platforms to ensure adequate behaviour of the Provisioning API.

This PR also evolves the notifications interceptors so that respective broker notifications are create for each supported platform.

### Case one (composition)
When a public plan is supported in all platforms (`plan.metadata.supportedPlatforms` is empty or not provided) we'll do things as before -> one visibility with empty `platform_id`.

### Case two (decomposition)
When a public plan has explicitly specified supported platforms, we'll create a visibility (with no labels) for each platform in our DB of the specified platform types. 

**Example:** _Public plan is supported only in CF and K8S environments - we'll create a visibility **for each** CF platform and **for each** K8S platform in our DB._  

## Pull Request status

* [x] Initial implementation
* [x] Integration tests


dotchev:
Also when a new platform is registered and it is among the supported platforms for some plan, the respective visibility should be created.

storage/interceptors/notifications_interceptor.go
```diff
@@ -199,3 +236,30 @@ func CreateNotification(ctx context.Context, repository storage.Repository, op t
...
+
+func determinePlatformIDs(oldPlatformIDs, updatedPlatformIDs []string) ([]string, []string, []string) {
+	preexistingPlatformIDs := slice.StringsIntersection(oldPlatformIDs, updatedPlatformIDs)
+	addedPlatformIDs := findDistinctStrings(updatedPlatformIDs, preexistingPlatformIDs)
+	removedPlatformIDs := findDistinctStrings(oldPlatformIDs, preexistingPlatformIDs)
+
+	return preexistingPlatformIDs, addedPlatformIDs, removedPlatformIDs
+}
+
+func findDistinctStrings(str1, str2 []string) []string {
```
pankrator: Move this function to the utility package

storage/interceptors/notifications_interceptor.go
```diff
@@ -82,18 +101,19 @@ func (ni *NotificationsInterceptor) OnTxUpdate(h storage.InterceptUpdateOnTxFunc
 		}
 		oldObject.SetLabels(nil)
 
-		// if the resource update contains change in the platform ID field this means that the notification would be processed by
-		// two platforms - one needs to perform a delete operation and the other needs to perform a create operation.
-		if oldPlatformID != updatedPlatformID {
-			if err := CreateNotification(ctx, repository, types.CREATED, updatedObject.GetType(), updatedPlatformID, &Payload{
+		for _, platformID := range addedPlatformIDs {
```
pankrator: I think the if above about removing the labels of the updated object should be below this for statement.
This is because if the supported platform is new, it does not know anything about this notification for now and it will need all labels for this object. This has to be double checked of course.

storage/interceptors/notifications_interceptor.go
```diff
@@ -103,18 +123,21 @@ func (ni *NotificationsInterceptor) OnTxUpdate(h storage.InterceptUpdateOnTxFunc
...
-				Additional: additionalDetails,
-			},
-			Old: &ObjectPayload{
-				Resource:   oldObject,
-				Additional: additionalDetails,
-			},
-			LabelChanges: labelChanges,
-		}); err != nil {
-			return nil, err
+		modifiedPlatformIDs := append(preexistingPlatformIDs, addedPlatformIDs...)
```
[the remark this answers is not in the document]
NickyMateev: Discussed.

storage/interceptors/notifications_interceptor.go
NickyMateev: Discussed.

PR #411, opened 2020-01-28
Add originating platform ID to operation

desislavaa:
# Pull Request Template

## Motivation

Currently, the Service Manager cleans up all operations after the configured interval passes.
In cases of frequent server restarts, where the server is restarted before the time for clean up has come, the operations are never cleaned up.

## Approach

New column `platform_id` is added to the operations which can be used for differentiating operations that came from other platforms and operations that belong to SMaaP. 

The maintainer also gets an [initial run](https://github.com/Peripli/service-manager/pull/411/files#diff-640bb52cdfeed2e49e630b57071b7e17R53) when the Service Manager is started.
 
## Pull Request status

* [x] Initial implementation
* [x] Integration tests


test/operations_test/operations_test.go
```diff
@@ -195,28 +195,27 @@ var _ = Describe("Operations", func() {
...
 					Expect(err).To(BeNil())
+					Expect(count).To(Equal(1))
 
-					return count
-				}, cleanupInterval*2).Should(Equal(0))
+					time.Sleep(cleanupInterval + time.Second)
+					count, err = ctx.SMRepository.Count(context.Background(), types.OperationType, byID)
+					Expect(err).To(BeNil())
+					Expect(count).Should(Equal(1))
+				})
```
dzahariev: Add test for internal operations.

operations/config.go
```diff
@@ -22,15 +22,21 @@ import (
 )
 
 const (
-	minTimePeriod     = time.Nanosecond
-	defaultJobTimeout = 5 * time.Minute
+	minTimePeriod              = time.Nanosecond
+	
+	defaultMarkOrphansInterval = 24 * time.Hour
+	defaultJobTimeout          = 7*24*time.Hour - 1*time.Hour
```
desislavaa: the job timeout must be less than the cleanup interval

PR #424, opened 2020-02-02
Authz has scope

eyalnathan:
## Motivation

In some cases filters/interceptors need to know if a request has a specific scope, regardless of the request access level. At the moment there is no API for checking that. 

## Approach

the new utility function in the authz package provides simple API for checking if a request user has a specific scope, abstracting the actual scope retrieval from the token. 

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [x] Unit tests
* [ ] Integration tests

### DIFF
--- a/pkg/security/http/authz/oauth_scopes.go
+++ b/pkg/security/http/authz/oauth_scopes.go
@@ -65,11 +65,5 @@ func HasScope(user *web.UserContext, scope string) (bool, error) {
 	}
 	userScopes := claims.Scopes
 
-	for _, userScope := range userScopes {
-		if scope == userScope {
-			return true, nil
-		}
-	}
-
-	return false, nil
+	return slice.StringsAnyEquals(userScopes, scope), nil
 }
--- a/pkg/security/http/authz/oauth_scopes.go
+++ b/pkg/security/http/authz/oauth_scopes.go
@@ -53,3 +53,23 @@ func findMostRestrictiveAccessLevel(levels []web.AccessLevel) web.AccessLevel {
 	}
 	return min
 }
+
+// Checks whether the user has the requested scope
+func HasScope(user *web.UserContext, scope string) (bool, error) {
+	var claims struct {
+		Scopes []string `json:"scope"`
+	}
+
+	if err := user.Data(&claims); err != nil {
+		return false, fmt.Errorf("could not extract scopes from token: %v", err)
+	}
+	userScopes := claims.Scopes
+
+	for _, userScope := range userScopes {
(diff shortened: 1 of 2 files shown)


pkg/security/http/authz/oauth_scopes.go
```diff
@@ -53,3 +53,23 @@ func findMostRestrictiveAccessLevel(levels []web.AccessLevel) web.AccessLevel {
...
+	var claims struct {
+		Scopes []string `json:"scope"`
+	}
+
+	if err := user.Data(&claims); err != nil {
+		return false, fmt.Errorf("could not extract scopes from token: %v", err)
+	}
+	userScopes := claims.Scopes
+
+	for _, userScope := range userScopes {
```
NickyMateev: Use `slice.StringsContaining()` instead

pkg/security/http/authz/oauth_scopes_test.go
```diff
@@ -1,51 +1,125 @@
...
+			user, ok := web.UserFromContext(ctx)
+			if !ok {
+				return nil, fmt.Errorf("Failed to retrieve user from context")
+			}
+
+			return user, nil
+		}
+
+		When("no scopes are available", func() {
+			user, err := getUserContextWithToken(tokenDataNoScopes, nil)
```
NickyMateev: Instead of writing this same line for every test, consider having `JustBeforeEach()` on top level which does this.
Then every test (inside every `When`) you can have a `BeforeEach()` which only sets up which token & err values to use in the `JustBeforeEach()`.

Also I see that every test starts with `Expect(err).ToNot(HaveOccurred())` - move that immediately after the `getUserContextWithToken()` invocation as that's where it makes more sense to be - closer to the actual function invocation itself, asserting that the completed function does not expect to have an error occurred.

pkg/security/http/authz/oauth_scopes.go
eyalnathan: Thanks! I switched to slice.StringsAnyEquals

pkg/security/http/authz/oauth_scopes_test.go
eyalnathan: I tried to refactor to use Just/BeforeEach, but I think it turned out uglier.. it requires more test-scoped variables, and 4 lines of code to replace each single line I have now. 
I did refactor it a bit, for example removed the error checks for the user context creation as there are no errors expected there, can you have another look? If that's not the Go way of doing things I will change it, but I find the current implementation more readable than the BeforeEach() option.

PR #429, opened 2020-02-05
Merge master in master-smaap

NickyMateev:


NickyMateev:
@georgifarashev , @pankrator , @KirilKabakchiev ,

There was a new Peripli mirgation in the `master` which had a date newer than all of the newly introduced migrations in the `master-smaap` branch. I renamed/changed this new migration's date to be before all of the new `master-smaap` migrations, but still newer than all the previous migrations.

We should be fine, right?

operations/maintainer.go
```diff
@@ -29,27 +29,32 @@ import (
...
+		cleanupInterval:         options.CleanupInterval,
+		operationExpirationTime: options.ExpirationTime,
 	}
 }
 
 // Run starts the two recurring jobs responsible for cleaning up operations which are too old
 // and deleting orphan operations
 func (om *Maintainer) Run() {
+	om.cleanUpOldOperations()
+	om.markOrphanOperationsFailed()
```
NickyMateev: @georgifarashev , @pankrator , @KirilKabakchiev 
Any idea why we trigger these immediately on SM start up and not wait until the first ticker to trigger them?

KirilKabakchiev: because if SM restarts often enough and the initial interval never passes, we theoretically might never do a cleanup

NickyMateev:
I've reordered our migrations to be after the newly added ones to the `master` branch.

PR #433, opened 2020-02-11
Operations Maintainer 

NickyMateev:
# Operations Maintainer 

## Motivation

The Maintainer performs maintenance on operations that need it. Maintenance includes cleaning up operations that are no longer necessary or scheduling/rescheduling operations that should be repeated with the appropriate action (the action might differ from the operation type. The different "things" the maintainer does are referred to as maintenance procedures.

The Maintainer runs as part of the SM process which means that at any given moment there are as many maintainers as SM nodes are started. So, we need to make sure that multiple maintainers do not pick up the same operation for processing (there needs to be some locking mechanism in place). 

## Pull Request status

* [x] Initial implementation
* [x] Integration tests


operations/maintainer.go
```diff
@@ -27,89 +28,303 @@ import (
...
+		logger := log.ForContext(om.smCtx).WithField(log.FieldCorrelationID, operation.CorrelationID)
+
+		var action storageAction
+
+		switch operation.Type {
+		case types.CREATE:
+			object, err := om.repository.Get(om.smCtx, operation.ResourceType, query.ByField(query.EqualsOperator, "id", operation.ResourceID))
+			if err != nil {
+				logger.Warnf("Failed to fetch resource with ID (%s) for operation with ID (%s): %s", operation.ResourceID, operation.ID, err)
+				return
```
dpanayotov: why break the whole rescheduling? leave this and try the next one?

operations/maintainer.go
```diff
@@ -27,89 +28,303 @@ import (
...
+	operations := objectList.(*types.Operations)
+	for i := 0; i < operations.Len(); i++ {
+		operation := operations.ItemAt(i).(*types.Operation)
+		logger := log.ForContext(om.smCtx).WithField(log.FieldCorrelationID, operation.CorrelationID)
+
+		byID := query.ByField(query.EqualsOperator, "id", operation.ResourceID)
+
+		action := func(ctx context.Context, repository storage.Repository) (types.Object, error) {
+			err := repository.Delete(ctx, operation.ResourceType, byID)
+			if err != nil {
```
dpanayotov: same

operations/maintainer.go
```diff
@@ -121,12 +336,31 @@ func (om *Maintainer) markOrphanOperationsFailed() {
...
 		if _, err := om.repository.Update(om.smCtx, operation, query.LabelChanges{}); err != nil {
-			log.D().Debugf("Failed to update orphan operation with ID (%s) state to FAILED: %s", operation.ID, err)
+			logger.Warnf("Failed to update orphan operation with ID (%s) state to FAILED: %s", operation.ID, err)
+			continue
+		}
+
+		byID := query.ByField(query.EqualsOperator, "id", operation.ResourceID)
+		action := func(ctx context.Context, repository storage.Repository) (types.Object, error) {
+			err := repository.Delete(ctx, operation.ResourceType, byID)
+			if err != nil {
```
dpanayotov: same

storage/encrypting_repository.go
```diff
@@ -14,14 +14,23 @@ import (
...
+// LockerCreatorFunc is a function building a storage.Locker with a specific advisory index
+type LockerCreatorFunc func(advisoryIndex int) Locker
+
+// Locker provides basic Lock/Unlock functionality
+type Locker interface {
 	// Lock locks the storage so that only one process can manipulate the encryption key. Returns an error if the process has already acquired the lock
 	Lock(ctx context.Context) error
 
+	// TryLock tries to lock the storage so that only one process can manipulate the encryption key. Returns an error if the process has already acquired the lock
+	TryLock(ctx context.Context) error
```
dpanayotov: huh? this has the same declaration and godoc, from an interface perspective I don't understand which to use

operations/maintainer.go
```diff
@@ -27,89 +28,303 @@ import (
 	"github.com/Peripli/service-manager/storage"
 )
 
+const (
+	initialOperationsLockIndex = 200
+	ZeroTime                   = "0001-01-01 00:00:00+00"
+)
+
+// MaintainerFunctor represents a named maintainer function which runs over a pre-defined period
+type MaintainerFunctor struct {
```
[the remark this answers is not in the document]
pankrator: no reason, will make it private

storage/encrypting_repository.go
pankrator: Will be revised

PR #438, opened 2020-02-19
Generalize instance visibility checks

eyalnathan:
## Motivation

Allow customizing visibility checks by providing a visibility metadata provider function

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [ ] Unit tests
* [x] Integration tests

### DIFF
--- a/api/filters/check_instance_visibility_filter.go
+++ b/api/filters/check_instance_visibility_filter.go
@@ -104,6 +104,11 @@ func (f *serviceInstanceVisibilityFilter) Run(req *web.Request, next web.Handler
 	if len(visibility.PlatformID) == 0 { // public visibility
 		return next.Handle(req)
 	}
+
+	if visibilityMetadata.LabelKey == "" {
+		return next.Handle(req)
+	}
+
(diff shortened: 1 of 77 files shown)


api/filters/check_instance_visibility_filter.go
```diff
@@ -100,8 +104,18 @@ func (f *serviceInstanceVisibilityFilter) Run(req *web.Request, next web.Handler
...
+	if visibilityMetadata.LabelKey == "" {
+		return next.Handle(req)
+	}
+
+	tenantLabels, ok := visibility.Labels[visibilityMetadata.LabelKey]
+	if ok && slice.StringsAnyEquals(tenantLabels, visibilityMetadata.LabelValue) {
+		return next.Handle(req)
+	}
+
+	if !ok {
```
NickyMateev: Why is this not tested?
When finding bugs, we make sure to add tests for the new bug, so that we can be confident going forward that it doesn't reoccur.

eyalnathan: added test for this specific case - plan with public visibility but that supports only specific platform

PR #440, opened 2020-02-21
Secured objects integrity

dpanayotov:
## Motivation

Prevent someone to change part(s) of the integral data of a secured object in the database and thus tricking the application to send sensitive information to a malicious application.

## Approach

Each secured object defines which is its integral data and a processor calculates an integrity to be stored next to the object. Upon retrieval of the object, the integrity is calculated and is compared to the previously stored one. 

## Open questions
* [ ] What do we do with existing secured objects? Create different application that sets the integrity? Leave this code here and remove it after some time? have some flag in db that says "intregration_setup_done"?
* [ ] Does it make sense to check integrity of all secured objects or only for brokers? 
* [ ] Does it make sense to allow extending the integral data outside of the open-source?
* [x] Does it make sense to allow configuration of integrity processor (e.g. different hashing algorithm in a wrapper)? - integrity processor is in storage config
* [x] Should the integrity always be 32 byte slice or allow variable size? - integrity is now a byte slice


pkg/sm/sm.go
```diff
@@ -259,6 +266,10 @@ func (smb *ServiceManagerBuilder) Build() *ServiceManager {
 		log.C(smb.ctx).Panic(err)
 	}
 
+	if err := smb.calculateIntegrity(); err != nil {
```
pankrator: In case we decide to update the integrity as part of the application: we might try to acquire a lock, before the update, so only one instance of SM updates the integrity

pkg/types/interfaces.go
```diff
@@ -42,6 +42,9 @@ type Strip interface {
 type Secured interface {
 	Encrypt(context.Context, func(context.Context, []byte) ([]byte, error)) error
 	Decrypt(context.Context, func(context.Context, []byte) ([]byte, error)) error
+	IntegralData() []byte
```
pankrator: The following 3 methods are only used in isolation from the above 2, so I will move the new ones in different interface. This will allow us to check integrity for example on visibilities also. In case the labels are changed outside application knowing.

pkg/sm/sm.go
pankrator: Also, the first time integrity is calculated, other instances should not start, but wait for lock. Otherwise they will start to serve requests with empty integrity for some resources

dpanayotov: How can this happen? Say node1 starts before node2:
* node1 starts processing integrity
* node2 starts processing the same integrity
* node1 finishes and starts processing requests - here all resources are with set integrity
* node2 continues to update integrity - it might do unnecessary db calls to update the integrity which should be okay.

pankrator: You are right

PR #441, opened 2020-02-23
Enable mTLS to ensure secure traffic between SM and Service brokers

alexwo:
## Motivation
Enable mTLS to ensure that traffic is secure and trusted between the service manager and the service brokers. 

mTLS SM->Broker flows
  - Catalog fetcher [Fetch catalog]
  - OSB calls [All]
  - SM as a platform [create/update instance / binding]

## Approach
The broker API endpoints are extended with additional mTLS  client_certificate and client_key optional fields. SM will use mTLS for any connectivity to service brokers that are registered or updated with mTLS settings. Basic broker auth credentials can still be provided in addition to mTLS transport security settings

https://github.com/Peripli/specification/blob/sm-ops/api.md#registering-a-service-broker

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [x] Unit tests
* [ ] Integration tests


pkg/types/service_broker.go
```diff
@@ -32,17 +35,51 @@ const maxNameLength = 255
...
+	BrokerURL     string             `json:"broker_url"`
+	Credentials   *Credentials       `json:"credentials,omitempty"`
+	Catalog       json.RawMessage    `json:"-"`
+	Services      []*ServiceOffering `json:"-"`
+	LastOperation *Operation         `json:"last_operation,omitempty"`
+}
+
+func (e *ServiceBroker) Interpret() error {
+	if e.Credentials != nil && e.Credentials.TLS != nil {
+		if val, err := base64.StdEncoding.DecodeString(e.Credentials.TLS.Certificate); err == nil {
```
liorokman: Consider moving to raw encoding, makes it easier to receive the configuration.

pkg/util/api.go
```diff
@@ -91,6 +96,15 @@ func BytesToObject(bytes []byte, object interface{}) error {
 	if err := unmarshal(bytes, object); err != nil {
 		return err
 	}
+
+	if err := interpret(object); err != nil {
+		return &HTTPError{
+			ErrorType:   "BadRequest",
+			Description: "Unable to decode TLS credentials: " + err.Error(),
```
liorokman: Should be "Could not interpret", not "unable to decode TLS credentials".

Also: do you really need the additional `HTTPError`  here? Why not return the error from `interpret()`?

pkg/util/client.go
```diff
@@ -30,21 +31,39 @@ import (
...
+		client.Transport = &transport
+
+		return reqFunc(req, client)
 	}
 }
 
 // SendRequest sends a request to the specified client and the provided URL with the specified parameters and body.
-func SendRequest(ctx context.Context, doRequest DoRequestFunc, method, url string, params map[string]string, body interface{}) (*http.Response, error) {
-	return SendRequestWithHeaders(ctx, doRequest, method, url, params, body, map[string]string{})
+func SendRequest(client *http.Client, ctx context.Context, doRequest DoRequestOsbFunc, method, url string, params map[string]string, body interface{}) (*http.Response, error) {
```
liorokman: `Context` is supposed to be the first parameter by convention.

pkg/util/list.go
```diff
@@ -53,7 +53,7 @@ func (li *ListIterator) Next(ctx context.Context, items interface{}, maxItems in
 
 	method := http.MethodGet
 	url := li.URL
-	response, err := SendRequest(ctx, li.DoRequest, method, url, params, nil)
+	response, err := SendRequest(http.DefaultClient, ctx, li.DoRequest, method, url, params, nil)
```
liorokman: `Context` should be first

storage/postgres/migrations/20200219214200_broker_tls.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE brokers ADD COLUMN tls_client_key bytea;
+ALTER TABLE brokers ADD COLUMN tls_client_certificate  VARCHAR(1000);
```
liorokman: Consider `bytea` instead of `varchar(1000)`

test/common/common.go
```diff
@@ -512,8 +512,8 @@ func Closer(s string) io.ReadCloser {
 	return NopCloser{bytes.NewBufferString(s)}
 }
 
-func DoHTTP(reaction *HTTPReaction, checks *HTTPExpectations) func(*http.Request) (*http.Response, error) {
-	return func(request *http.Request) (*http.Response, error) {
+func DoHTTP(reaction *HTTPReaction, checks *HTTPExpectations) func(request *http.Request, client *http.Client) (*http.Response, error) {
```
liorokman: Use the type instead of rewriting the function.

test/common/common.go
```diff
@@ -559,10 +559,10 @@ type HTTPCouple struct {
 	Reaction     *HTTPReaction
 }
 
-func DoHTTPSequence(sequence []HTTPCouple) func(*http.Request) (*http.Response, error) {
+func DoHTTPSequence(sequence []HTTPCouple) func(*http.Request, *http.Client) (*http.Response, error) {
```
liorokman: Use the type instead of the explicit signature as a return value.

PR #443, opened 2020-02-24
Credentials API

NickyMateev:
# Credentials API

## Motivation

We'll need to rotate the broker platform credentials for security reasons. This PR introduces a new Credentials API, which SM Proxies (Agents) will make requests to with hashed broker platform credentials.

## Pull Request status

* [x] Initial implementation
* [x] Integration tests


storage/interceptors/notifications_interceptor.go
```diff
@@ -189,6 +194,10 @@ func (ni *NotificationsInterceptor) OnTxDelete(h storage.InterceptDeleteOnTxFunc
 				}); err != nil {
 					return err
 				}
+
+				if err := ni.DeletePostConditionFunc(ctx, oldObject, repository, platformID); err != nil {
```
KirilKabakchiev: is this needed

storage/interceptors/notifications_interceptor.go
```diff
@@ -126,6 +127,10 @@ func (ni *NotificationsInterceptor) OnTxUpdate(h storage.InterceptUpdateOnTxFunc
 			}); err != nil {
 				return nil, err
 			}
+
```
KirilKabakchiev: if the notification is not applied in the platform and we delete the credentials in smdb, the platform credentials will now work once again because of the fallback behaviour. is this ok ? More generally can we revise when the platform credentials should be usable and add tests for the cases?

I think platform credentials should be written as old_credentials/new_credentials initially (probably with migration?) and once they fall out due to rotations, fall backing to them should no longer work

api/credentials_controller.go
```diff
@@ -0,0 +1,168 @@
...
+
+func NewCredentialsController(ctx context.Context, options *Options) *CredentialsController {
+	return &CredentialsController{
+		BaseController: NewController(ctx, options, web.BrokerPlatformCredentialsURL, types.BrokerPlatformCredentialType, func() types.Object {
+			return &types.BrokerPlatformCredential{}
+		}),
+	}
+}
+
+func (c *CredentialsController) Routes() []web.Route {
```
KirilKabakchiev: what if multiple calls to this api are done and applying the notification result by the agent in CC fails ?
From a high level i think that the last known credentials in CC should be still valid in SM

pkg/security/authenticators/basic_authentication.go
```diff
@@ -57,7 +70,77 @@ func (a *Basic) Authenticate(request *web.Request) (*web.UserContext, httpsec.De
...
 	}
 
-	bytes, err := json.Marshal(platform)
+	return buildResponse(username, platform)
+}
+
+//BasicOSBAuthenticator attempts to authenticate basic auth requests with provided broker platform credentials
+func BasicOSBAuthenticator(request *web.Request, repository storage.Repository, username, password string) (*web.UserContext, httpsec.Decision, error) {
+	ctx := request.Context()
+	log.C(ctx).Debugf("Attempting to authenticate broker platform credentials")
```
KirilKabakchiev: maybe log username and broker id too ?

pkg/security/authenticators/basic_authentication.go
```diff
@@ -57,7 +70,77 @@ func (a *Basic) Authenticate(request *web.Request) (*web.UserContext, httpsec.De
...
+
+		byUsername.LeftOp = "old_username"
+		credentialsList, err = repository.List(ctx, types.BrokerPlatformCredentialType, byBrokerID, byUsername)
+		if err != nil {
+			return nil, httpsec.Abstain, fmt.Errorf("could not get credentials entity from storage: %s", err)
+		}
+
+		if credentialsList.Len() != 1 {
+			log.C(ctx).Debugf("Authenticating broker platform credentials failed - will try to fallback to platform credentials authentication")
+			return BasicPlatformAuthenticator(request, repository, username, password)
```
KirilKabakchiev: this should probably go away 

I think platform credentials should be written as old_credentials/new_credentials initially (probably with migration?) and once they fall out due to rotations, fall backing to them should no longer work

pkg/security/authenticators/basic_authentication.go
```diff
@@ -57,7 +70,77 @@ func (a *Basic) Authenticate(request *web.Request) (*web.UserContext, httpsec.De
...
+
+	byBrokerID := query.ByField(query.EqualsOperator, "broker_id", brokerID)
+	byUsername := query.ByField(query.EqualsOperator, "username", username)
+
+	credentialsList, err := repository.List(ctx, types.BrokerPlatformCredentialType, byBrokerID, byUsername)
+	if err != nil {
+		return nil, httpsec.Abstain, fmt.Errorf("could not get credentials entity from storage: %s", err)
+	}
+
+	useOldCredentials := false
```
KirilKabakchiev: probably instead of this bool you can have the actual struct for holding credentials here and just use it with the correct values assigned later on. Then you can remove `if useOldCredentials {
		passwordHash = credentials.OldPasswordHash
	}`

api/credentials_controller.go
```diff
@@ -0,0 +1,168 @@
...
+			query.ByField(query.EqualsOperator, "resource", string(types.ServiceBrokerType)),
+		}
+
+		obj, err := c.repository.Get(ctx, types.NotificationType, criteria...)
+		if err != nil {
+			if err == util.ErrNotFoundInStorage {
+				return nil, &util.HTTPError{
+					ErrorType:   "CredentialsError",
+					Description: fmt.Sprintf("Invalid notification ID %s - request rejected", body.NotificationID),
+					StatusCode:  http.StatusConflict,
```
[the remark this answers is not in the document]
NickyMateev: Bad request/Unauthorized are pretty common request errors - we wanted to make sure that we return something more unique, so that the proxy client could interpret it and know how to handle it and know that it's a "happy path" scenario instead of some real issue. Also the idea for Conflict was that the credentials are already in SM DB.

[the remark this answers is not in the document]
KirilKabakchiev: my understanding was that full resync wont do credentials rotation as the flow is not initiated by sm?

api/credentials_controller.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	objFromDB, err := c.repository.Get(ctx, c.objectType, criteria...)
+	if err != nil {
+		if err == util.ErrNotFoundInStorage {
+			return c.registerCredentials(ctx, body)
+		}
+		return nil, util.HandleStorageError(err, c.objectType.String())
+	}
+
+	if body.NotificationID == "" {
```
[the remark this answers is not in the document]
DimitarPetrov: It should be like this, because in the full resync scenario we have one happy path, which is if there aren't any broker credentials we must create them, and if already exists full resync doe s not update them. This flow does not include notificationID in the body.

storage/interceptors/notifications_interceptor.go
NickyMateev: no

pkg/security/authenticators/basic_authentication.go
```diff
@@ -19,19 +19,25 @@ package authenticators
...
 )
 
+//BasicAuthenticatorFunc defines a function which attempts to authenticate a basic auth request
+type BasicAuthenticatorFunc func(request *web.Request, repository storage.Repository, username, password string) (*web.UserContext, httpsec.Decision, error)
+
 // Basic for basic security
 type Basic struct {
-	Repository storage.Repository
+	Repository             storage.Repository
+	BasicAuthenticatorFunc BasicAuthenticatorFunc
```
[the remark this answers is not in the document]
NickyMateev: This struct was from before. We need it to inject the storage in some way - we cannot provide into the Authenticate function as it will not implement Authenticator then.

api/credentials_controller.go
NickyMateev: We've discussed ways in which to keep the last valid credentials in SM, like for example - not allowing rotation of credentials, if the current ones have not been used yet (old_username & old_password are not ""), but there are numerous bad consequences of this like for example we'll never be able to rotate credentials ever again.

PR #445, opened 2020-02-26
Migrating to Go Modules

alexwo:
# Pull Request Template

## Prerequisites

## Motivation

## Approach

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [ ] Unit tests
* [ ] Integration tests

## Third-party code

## Work in Progress label


Dockerfile
```diff
@@ -7,14 +7,11 @@ FROM golang:1.12.7-alpine3.10 AS builder
...
 RUN apk --no-cache add git
-RUN go get github.com/golang/dep/cmd/dep
 
 # Directory in workspace
 WORKDIR "/go/src/github.com/Peripli/service-manager"
 
-# Copy dep files only and ensure dependencies are satisfied
-COPY Gopkg.lock Gopkg.toml ./
-RUN dep ensure --vendor-only -v
+RUN go mod vendor
```
liorokman: Why do you need this?

go build will work just fine even if you don't create the vendor directory.

Dockerfile
```diff
@@ -7,14 +7,11 @@ FROM golang:1.12.7-alpine3.10 AS builder
```
liorokman: You should really move from golang 1.12 to the latest version of golang. 

1.14 was released on February 25, 2020. Since golang versions are supported for two major releases (https://endoflife.date/go), this means that Golang 1.12 is not supported anymore - it's passed its end-of-life.

PR #446, opened 2020-02-27
Notifications ready

pankrator:
## Motivation

*Currently for each resource a notification is created whether it is ready or not.
The proposed solution takes into account whether the resource is ready and only then a notification will be created for it.*

## Pull Request status

* [x] Initial implementation
* [x] Unit tests
* [ ] Integration tests


storage/interceptors/broker_notifications_interceptor.go
```diff
@@ -52,15 +53,58 @@ func NewBrokerNotificationsInterceptor() *NotificationsInterceptor {
...
-					Services: broker.Services,
+					Services: services,
 				}
 			}
 			return details, nil
 		},
 	}
 }
 
+func loadServicesFromStorage(ctx context.Context, broker *types.ServiceBroker, repository storage.Repository) ([]*types.ServiceOffering, error) {
```
DimitarPetrov: You can use https://github.com/Peripli/service-manager/blob/9bcb485ef331994537ec3a7e82a564a7d119a47f/storage/catalog/loader.go

storage/interceptors/notifications_interceptor.go
```diff
@@ -96,6 +102,21 @@ func (ni *NotificationsInterceptor) OnTxUpdate(h storage.InterceptUpdateOnTxFunc
 			return nil, err
 		}
 
+		// If the updated object has just become ready, then create a CREATED notification for it
+		if !oldObject.GetReady() && updatedObject.GetReady() {
```
NickyMateev: It seems like you've forgotten to strip the updatedObject's labels like it's done a few lines below.

test/notification_test/notification_test.go
```diff
@@ -596,5 +599,91 @@ var _ = Describe("Notifications Suite", func() {
...
+	return func(ctx context.Context, obj types.Object) (types.Object, error) {
+		robj, err := h(ctx, obj)
+		if err != nil {
+			return nil, err
+		}
+		return robj, errors.New("test test")
+	}
+}
+
+func (p *testCreateInterceptor) OnTxCreate(f storage.InterceptCreateOnTxFunc) storage.InterceptCreateOnTxFunc {
```
NickyMateev: You can remove this and just make it a `CreateAroundTxInterceptor`

storage/interceptors/broker_notifications_interceptor.go
```diff
@@ -52,8 +54,17 @@ func NewBrokerNotificationsInterceptor() *NotificationsInterceptor {
 			details := make(objectDetails, objects.Len())
 			for i := 0; i < objects.Len(); i++ {
 				broker := objects.ItemAt(i).(*types.ServiceBroker)
+				services := broker.Services
+				if len(services) == 0 {
+					var err error
```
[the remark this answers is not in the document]
pankrator: yes, exactly

storage/interceptors/notifications_interceptor.go
pankrator: This is because, when we create an object we don't have label changes, thus we need the created object with all its labels so it can be correctly applied by the agent

pkg/types/operation.go
```diff
@@ -53,19 +53,27 @@ const (
 	FAILED OperationState = "failed"
 )
 
+type RelatedType struct {
+	ID            string            `json:"id"`
```
DimitarPetrov: Isn't this should be omitempty also in the case where Criteria is used for identification?

DimitarPetrov:
Since other changes which include migrations are already in master, it is reasonable to change the time of your migrations to be the last ones.

pkg/types/operation.go
pankrator: I was wondering for the same thing, but left it as is. May be it will be better to be omitempty, as it will be cleaner

PR #447, opened 2020-02-27
Scheduler to not accept operations that are older than ROT

KirilKabakchiev:


operations/scheduler.go
```diff
@@ -515,6 +515,14 @@ func (s *Scheduler) executeOperationPreconditions(ctx context.Context, operation
 		return fmt.Errorf("scheduling for operations of type %s is not allowed", string(types.SUCCEEDED))
 	}
 
+	if time.Now().UTC().After(operation.CreatedAt.Add(s.reconciliationOperationTimeout)) {
+		return &util.HTTPError{
```
NickyMateev: An http error is returned, but we shouldn't expect this to happen as a result of a REST call, right? It will be a result of the maintainer rescheduling some operation.

test/operations_test/operations_test.go
```diff
@@ -332,12 +354,13 @@ var _ = Describe("Operations", func() {
 			})
 		})
 
-		When("Specified job timeout passes", func() {
+		When("Specified action timeout passes", func() {
```
NickyMateev: While you're at it, could you also replace all _job_ occurrences with _action_?

test/operations_test/operations_test.go
```diff
@@ -54,48 +54,71 @@ var _ = Describe("Operations", func() {
...
+			When("reconciliation timeout runs out", func() {
+				BeforeEach(func() {
+					postHook := func(e env.Environment, servers map[string]FakeServer) {
+						e.Set("operations.reconciliation_operation_timeout", 2*time.Nanosecond)
+					}
+
+					ctx = NewTestContextBuilder().WithEnvPostExtensions(postHook).SkipBasicAuthClientSetup(true).Build()
+				})
+
+				It("marks operation as failed", func() {
```
NickyMateev: Like I said above, I don't see this as a realistic case for a REST call to fail with this, but I guess it's fine to test it like so

operations/scheduler.go
KirilKabakchiev: yes unless at some point its necessary to allow rescheduling operations from a remote system

test/operations_test/operations_test.go
KirilKabakchiev: i added a commit where the error is added to the operation errors if the operation exists. since operations are returned to end users, operation errors have to be non-internal a.k.a. operation errors are http errors

PR #448, opened 2020-02-27
Transitive resources

pankrator:
## Motivation

Each operation associated with a resource might create/update/delete other resources.
For example when updating broker, new plans might be created/deleted and existing updated. Also visiblities and notifications might be created along the way.
These resources should be saved in the operation associated with the main resource, meaning the broker itself.

## Pull Request status

* [x] Initial implementation
* [x] Refactoring
* [x] Unit tests
* [ ] Integration tests

### DIFF
--- a/operations/scheduler.go
+++ b/operations/scheduler.go
@@ -25,6 +25,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/Peripli/service-manager/operations/opcontext"
+
 	"github.com/Peripli/service-manager/pkg/log"
 	"github.com/Peripli/service-manager/pkg/query"
 	"github.com/Peripli/service-manager/pkg/types"
@@ -386,6 +388,7 @@ func (s *Scheduler) handleActionResponse(ctx context.Context, actionObject types
 	if err != nil {
 		return nil, err
 	}
+	opAfterJob.TransitiveResources = opBeforeJob.TransitiveResources
 
 	// if an action error has occurred we mark the operation as failed and check if deletion has to be scheduled
 	if actionError != nil {
@@ -486,6 +489,20 @@ func (s *Scheduler) handleActionResponseSuccess(ctx context.Context, actionObjec
 			}); err != nil {
 				return err
 			}
+
+			for _, trR := range opAfterJob.TransitiveResources {
+				if trR.OperationType != types.CREATE {
+					continue
+				}
+				resource, err := storage.Get(ctx, trR.Type, query.ByField(query.EqualsOperator, "id", trR.ID))
+				if err != nil {
+					return err
+				}
+				resource.SetReady(true)
+				if _, err := storage.Update(ctx, resource, query.LabelChanges{}); err != nil {
+					return err
+				}
+			}
 		}
 		return nil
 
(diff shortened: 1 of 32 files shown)


test/broker_test/broker_test.go
```diff
@@ -1806,6 +1882,46 @@ func blueprint(setNullFieldsValues bool) func(ctx *TestContext, auth *SMExpect,
...
+			// Do not count the notifications and resources which are not for created
+			if tr.Object().Value("type").String().Raw() != types.NotificationType.String() &&
+				tr.Object().Value("operation_type").String().Raw() == string(types.CREATE) {
+				transitiveResourcesActualCount++
+			}
+		}
+		if transitiveResourcesActualCount != transitiveResourcesExpectedCount {
+			Fail(fmt.Sprintf("Expected transitive resources %d but got %d", transitiveResourcesExpectedCount, transitiveResourcesActualCount))
+		}
+
```
DimitarPetrov: This logic seems like a test case to me, not part of the resource creation blueprint.

operations/scheduler.go
```diff
@@ -315,6 +317,18 @@ func (s *Scheduler) storeOrUpdateOperation(ctx context.Context, operation, lastO
 	return nil
 }
 
+func updateTransitiveResources(ctx context.Context, storage storage.Repository, resources []*types.RelatedType, updateFunc func(obj types.Object)) error {
+	for _, trR := range resources {
+		if trR.OperationType != types.CREATE {
+			continue
+		}
+		if err := fetchAndUpdateResource(ctx, storage, trR.ID, trR.Type, updateFunc); err != nil {
```
DimitarPetrov: This could be in the if above if you change it to be `trR.OperationType == types.CREATE`

pkg/types/operation.go
```diff
@@ -53,19 +53,27 @@ const (
 	FAILED OperationState = "failed"
 )
 
+type RelatedType struct {
+	ID            string            `json:"id"`
+	Criteria      interface{}       `json:"criterion"`
```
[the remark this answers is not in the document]
pankrator: This is an array of criterias, but I don't remember now why it has to be interface{}. I think it was because it cannot unmarshal it otherwise. 
I will change the name to Criterion

storage/postgres/migrations/20200226131000_operations_resources.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE operations ADD COLUMN IF NOT EXISTS transitive_resources json;
```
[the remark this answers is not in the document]
pankrator: It is yes, it should be `'[]'`

pkg/types/operation.go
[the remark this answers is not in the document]
DimitarPetrov: If it is array of criterions, probably criteria is better naming for the json tag.

operations/scheduler.go
pankrator: I will leave it nested if, because otherwise the logic is hidden too far from the beginning of line

test/broker_test/broker_test.go
pankrator: Sure, I will try to make it as a test case

PR #449, opened 2020-02-28
Credentials integrity startup

dpanayotov:
Adding InTransaction as part of the credentials integrity makes the application unable to start when the integrity of existing objects is calculated. Also, maintainer may start work on objects which don't have set integrity thus failing


storage/integrity_repository.go
```diff
@@ -107,30 +107,24 @@ func (cr *integrityRepository) Delete(ctx context.Context, objectType types.Obje
...
 
-// InTransaction wraps repository passed in the transaction to also validate integrity
 func (cr *TransactionalIntegrityRepository) InTransaction(ctx context.Context, f func(ctx context.Context, storage Repository) error) error {
-	return cr.repository.InTransaction(ctx, func(ctx context.Context, storage Repository) error {
-		return f(ctx, &integrityRepository{
-			repository:         storage,
-			integrityProcessor: cr.integrityProcessor,
-		})
-	})
+	return cr.repository.InTransaction(ctx, f)
```
dpanayotov: disabling integrity validation in transaction is not a problem ATM as we don't have transitive integral resources that require manipulation in a transaction opened for the top resource

dpanayotov:
@pankrator @NickyMateev please review these changes. unit, component and e2e tests are passing

PR #453, opened 2020-03-04
Credentials integrity

dpanayotov:
## Motivation

Prevent someone to change part(s) of the integral data of a secured object in the database and thus tricking the application to send sensitive information to a malicious application.

## Approach

Each secured object defines which is its integral data and a processor calculates an integrity to be stored next to the object. Upon retrieval of the object, the integrity is calculated and is compared to the previously stored one.


dpanayotov:
@NickyMateev @pankrator please review again

PR #455, opened 2020-03-06
Proxy only sm supported encodings in osb calls

alexwo:
## Motivation

The Accept-Encoding field is used in an HTTP request to indicate which encodings are accepted in a response message to the request. This field is used to specify the values of gzip, deflate, compress and identity.  A value of identity indicates that the data must not be compressed. A wildcard of * indicates that any encoding is accepted. The Accept-Encoding field can also be empty, indicating that no encoding is accepted.  SM needs to support the handling of broker responses when the Accept-Encoding header is present and provided by the OSB clients as svcat.

## Approach

The proposed changes add an Accept-Encoding header filter for OSB calls and filters out the Accept-Encoding header incase the request encoding is not supported.

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [x ] Unit tests
* [ ] Integration tests

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -135,6 +135,7 @@ func New(ctx context.Context, e env.Environment, options *Options) (*web.API, er
 		// Default filters - more filters can be registered using the relevant API methods
 		Filters: []web.Filter{
 			&filters.Logging{},
+			&filters.SupportedEncodingsFilter{},
 			&filters.SelectionCriteria{},
 			&filters.ServiceInstanceStripFilter{},
 			&filters.ServiceBindingStripFilter{},
--- a/api/filters/supported_encodings_filter.go
+++ b/api/filters/supported_encodings_filter.go
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2018 The Service Manager Authors
+ *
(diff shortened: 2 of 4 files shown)


alexwo:
> According to rfc2616:
> "The "identity" content-coding is always acceptable, unless
> specifically refused because the Accept-Encoding field includes
> "identity;q=0", or because the field includes "*;q=0" and does
> not explicitly include the "identity" content-coding. If the
> Accept-Encoding field-value is empty, then only the "identity"
> encoding is acceptable."
> 
> Maybe we should check if accept encoding is _identity;q=0_ or somehow excludes "identity".
According to the spec excluding an encoding type is optional and up to the server. 
Its sats the following:
Even if both the client and the server supports the same compression algorithms, the server may choose not to compress the body of a response

As well we are not required to respond with 406 but rather we want to always fulfill the client request.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406

the spec has the following comment:
In practice, this error is very rarely used. Instead of responding using this error code, which would be cryptic for the end-user and difficult to fix, servers ignore the relevant header and serve an actual page to the user. It is assumed that even if the user won't be completely happy, they will prefer this to an error code.

PR #456, opened 2020-03-07
Fix Rescheduling of Instance/Binding operations

NickyMateev:
# Fix Rescheduling of Instance/Binding operations

## Motivation

Rescheduling of CREATE/UPDATE operation by maintainer fails for instance/binding resources due to unique constraint violation as the interceptors enforcing uniqueness of instance/binding names find the instance/binding resources in SM DB (they are with ready=false) when trying to reschedule the operation.

This change tweaks the two uniqueness interceptors, but also more importantly enables the previously disabled maintainer tests (which test the rescheduling mechanism), which had been previously failing remotely on Travis. Due to those tests having been disabled we weren't able to detect the rescheduling bug when the unique instance/binding name interceptors were initially introduced.

## Pull Request status

* [x] Initial implementation
* [x] Integration tests

### DIFF
--- a/storage/interceptors/unique_binding_names_iterceptor.go
+++ b/storage/interceptors/unique_binding_names_iterceptor.go
@@ -65,7 +65,7 @@ func (c *uniqueBindingNameInterceptor) AroundTxCreate(h storage.InterceptCreateA
 
 func (c *uniqueBindingNameInterceptor) checkUniqueName(ctx context.Context, binding *types.ServiceBinding) error {
 	operation, operationFound := operations.GetFromContext(ctx)
-	if operationFound {
+	if !operationFound {
 		log.C(ctx).Debug("operation missing from context")
 	}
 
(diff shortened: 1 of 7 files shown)


storage/interceptors/unique_binding_names_iterceptor.go
```diff
@@ -62,21 +64,32 @@ func (c *uniqueBindingNameInterceptor) AroundTxCreate(h storage.InterceptCreateA
 }
 
 func (c *uniqueBindingNameInterceptor) checkUniqueName(ctx context.Context, binding *types.ServiceBinding) error {
+	operation, operationFound := operations.GetFromContext(ctx)
+	if !operationFound {
+		log.C(ctx).Debug("operation missing from context")
+	}
+
+	if operationFound && operation.Reschedule {
+		log.C(ctx).Debug("skipping unique check of binding name as this is a rescheduled operation")
```
pankrator: Better log it on info level, as we don't expect this to happen too often. Same for the other interceptor

NickyMateev: addressed

PR #465, opened 2020-03-13
Adding support for broker tls based authentication

alexwo:
## Motivation

Support more secure communication between sm and service brokers over TLS with client authentication.

## Approach
Allow service brokers to register with TLS settings ( client key and certificate) PEM format.

SM will pass broker TLS settings at all sm to broker communication flows as:

  - Catalog fetcher [Fetch catalog]
  - OSB calls [All]
  - SM as a platform [create/update instance / binding]

The service broker entity
- Extended to store client certificates and keys

DB migration
- Tear up and down SQL scripts are added to add the tls_client key and certificate properties to the broker's entity

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [x] Unit tests
* [x] Integration tests


pkg/types/service_broker.go
```diff
@@ -77,6 +97,14 @@ func (e *ServiceBroker) transform(ctx context.Context, transformationFunc func(c
...
 		return err
 	}
+
+	if e.Credentials.TLS != nil && e.Credentials.TLS.Key != "" {
+		transformedPrivateKey, err := transformationFunc(ctx, []byte(e.Credentials.TLS.Key))
+		if err != nil {
+			return err
+		}
+		e.Credentials.TLS.Key = string(transformedPrivateKey)
+	}
```
dotchev: move this section so it is not in the middle of password transformation

api/osb/osb_controller.go
```diff
@@ -162,13 +170,40 @@ func (c *Controller) proxy(r *web.Request, logger *logrus.Entry, broker *types.S
...
+	if err != nil {
+		return nil, &util.HTTPError{
+			ErrorType:   "BadRequest",
+			Description: "unable to get tls configuration: " + err.Error(),
+			StatusCode:  http.StatusBadRequest,
+		}
+	}
+
+	if len(tlsConfig.Certificates) > 0 {
+		proxyTransport, ok := proxy.Transport.(*http.Transport)
```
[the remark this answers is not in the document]
alexwo: component tests with OSB requests should be extended, we will add it.

api/osb/osb_controller.go
```diff
@@ -178,13 +177,13 @@ func buildProxy(targetBrokerURL *url.URL, logger *logrus.Entry, broker *types.Se
...
-	tlsConfig, err := broker.GetTlsConfig()
+	tlsConfig, err := broker.GetTLSConfig()
 
 	if err != nil {
 		return nil, &util.HTTPError{
-			ErrorType:   "BadRequest",
+			ErrorType:   "InternalError",
 			Description: "unable to get tls configuration: " + err.Error(),
-			StatusCode:  http.StatusBadRequest,
+			StatusCode:  http.StatusInternalServerError,
```
dotchev: just return a normal error and the error handler will return status 500, see https://github.com/Peripli/service-manager/blob/master/pkg/util/errors.go#L77
same thing below

api/osb/osb_controller.go
```diff
@@ -162,13 +169,29 @@ func (c *Controller) proxy(r *web.Request, logger *logrus.Entry, broker *types.S
...
-func buildProxy(targetBrokerURL *url.URL, logger *logrus.Entry, broker *types.ServiceBroker) *httputil.ReverseProxy {
+func buildProxy(targetBrokerURL *url.URL, logger *logrus.Entry, broker *types.ServiceBroker) (*httputil.ReverseProxy, error) {
 	proxy := httputil.NewSingleHostReverseProxy(targetBrokerURL)
 	director := proxy.Director
 	proxy.Director = func(request *http.Request) {
 		director(request)
 		logger.Infof("Forwarded OSB request to service broker %s at %s", broker.Name, request.URL)
 	}
+
+	brokerClient, err := (&client.BrokerClient{}).New(broker, nil)
```
pankrator: It is more Go oriented to have a function `New` in the package which returns a new `BrokerClient`, which is configured

alexwo: 10x will adjust

api/osb/osb_controller.go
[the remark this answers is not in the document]
alexwo: Fixed based on feedback, now that looks as:
`useBrokerTLSConfig, transportWithTLS := brokerClient.GetTransportWithTLS()
	if useBrokerTLSConfig {
		proxy.Transport = transportWithTLS
	}`

Added this unit to validate it behaves correctly:
https://github.com/Peripli/service-manager/blob/15b83f391b7608bcfb025790ee0ecd0c76671776/api/osb/osb_controller_test.go

pankrator:
Add at least one test to test the happy path of the scenario in: `test/osb_test/osb_suite_test.go`.
There must be at least one component test that tests the tls functionality is working.

test/broker_test/broker_test.go
```diff
@@ -423,6 +515,29 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+								Expect().
+								Status(http.StatusBadRequest).
+								JSON().Object().Keys().Contains("error", "description")
+
+							assertInvocationCount(brokerServer.CatalogEndpointRequests, 0)
+						})
+					})
+
+					Context("when tls credentials are invalid", func() {
+						It("returns 400", func() {
```
dotchev: test description does not match actual test

test/broker_test/broker_test.go
```diff
@@ -423,6 +515,29 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+							reply := ctx.SMWithOAuth.POST(web.ServiceBrokersURL).WithJSON(postBrokerRequestWithTls).
+								Expect().
+								Status(http.StatusCreated).
+								JSON().Object()
+							reply.ContainsMap(expectedBrokerResponseTls)
+
+							assertInvocationCount(brokerServer.CatalogEndpointRequests, 1)
```
dotchev: verify that broker receives the client certificate

test/broker_test/broker_test.go
```diff
@@ -423,6 +515,29 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+						})
+					})
```
dotchev: test all possible combinations of basic and/or tls - 4

DanFrei: Done!

pkg/types/credentials.go
```diff
@@ -41,19 +48,32 @@ func (c *Credentials) MarshalJSON() ([]byte, error) {
...
+			return errors.New("missing broker username")
+		}
+		if c.Basic.Password == "" {
+			return errors.New("missing broker password")
+		}
+	} else {
+		_, err := tls.X509KeyPair([]byte(c.TLS.Certificate), []byte(c.TLS.Key))
+		if err != nil {
+			return errors.New("invalidate TLS configuration: " + err.Error())
+		}
```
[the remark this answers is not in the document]
DanFrei: Please correct me if i'm wrong but I think that our code applies the aforementioned rules explicitly.
Meaning, it's enough to have TLS present (with valid certificate) to pass the flow.
  However, if we get an empty/nil TLS we check that basic auth is present and valid.
Elswise , we fail the flow .

(Of course this supports also having both basic and TLS present)

pkg/types/service_broker.go
DanFrei: I thing we would still rather keep it as is - as in this case it should be safer to return an empty TLS config to avoid problems going onwards, for instance when there're attempts to convert it to string

pkg/types/service_broker.go
```diff
@@ -32,17 +33,28 @@ const maxNameLength = 255
...
+	if e.Credentials.TLS != nil && e.Credentials.TLS.Certificate != "" && e.Credentials.TLS.Key != "" {
+		cert, err := tls.X509KeyPair([]byte(e.Credentials.TLS.Certificate), []byte(e.Credentials.TLS.Key))
+		if err != nil {
+			return &tls.Config{}, err
+		}
+		tlsConfig.Certificates = []tls.Certificate{cert}
+	}
 
-	LastOperation *Operation `json:"last_operation,omitempty"`
+	return &tlsConfig, nil
```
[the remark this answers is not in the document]
DanFrei: I think we would still rather keep it as is - as in this case it should be safer to return an empty TLS config to avoid problems going onwards, for instance when there're attempts to convert it to string

pkg/types/service_broker.go
DanFrei: There's a reason this section is there:
In case there're non-empty TLS settings, then and only then, we'll apply transformation on the private key ( we can no longer assume that the broker will always haver a user and password as it used to be).
Otherwise we don't want to save it in the database.
if we move this section outside the 'if' - it will not work properly!

api/osb/osb_controller.go
DanFrei: Done

pkg/types/credentials.go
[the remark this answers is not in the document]
DanFrei: As for the component tests for major broker flows, the commit is on the way (probably we'll be delivered on Sunday)

alexwo:
> Add at least one test to test the happy path of the scenario in: `test/osb_test/osb_suite_test.go`.
> There must be at least one component test that tests the tls functionality is working.

Done

alexwo:
> Redesign management of http client/transport
> Add component tests for major broker flows:
> 
> * broker create & update
> * OSB
> * SM as a platform (write APIs)

Done

alexwo:
> Redesign management of http client/transport
> Add component tests for major broker flows:
> 
> * broker create & update - Added
> * OSB - Added
> * SM as a platform (write APIs) - Added

Added component tests to cover all sm -> broker communication scenarios

api/osb/osb_controller.go
```diff
@@ -162,13 +169,25 @@ func (c *Controller) proxy(r *web.Request, logger *logrus.Entry, broker *types.S
...
-func buildProxy(targetBrokerURL *url.URL, logger *logrus.Entry, broker *types.ServiceBroker) *httputil.ReverseProxy {
+func buildProxy(targetBrokerURL *url.URL, logger *logrus.Entry, broker *types.ServiceBroker) (*httputil.ReverseProxy, error) {
 	proxy := httputil.NewSingleHostReverseProxy(targetBrokerURL)
 	director := proxy.Director
 	proxy.Director = func(request *http.Request) {
 		director(request)
 		logger.Infof("Forwarded OSB request to service broker %s at %s", broker.Name, request.URL)
 	}
+
+	brokerClient, err := client.New(broker, nil)
```
dotchev: `brokerClient` is not fully configured, `brokerClient.SendRequest` would crash

alexwo: The brokerClient sendRequest won't be used in this flow, we only need the client here to obtain a copy from the GetTransportWithTLS method.

alexwo: - Adjusted to use a broker-client transport object in the OSB, sm as a platform flows
- Client object will validate the func handler is not nil on New

dotchev:
Wait for the tests before merging 😉

PR #467, opened 2020-03-17
allow setting supportedPlatformIDs in plan metadata

eyalnathan:
# Pull Request Template

## Motivation

Some service brokers require fine grained control over the supported platforms - while SM allows defining the supported platform types, there are requirements to support only specific platform instances.

## Approach

This PR adds the ability to define an additional property in the plan metadata - `supportedPlatformNames`
This parameter can be used instead of the existing `supportedPlatforms` to specify specific platform instances on which the plan will be available. 

## Pull Request status

* [x] Initial implementation
* [x] Refactoring
* [x] Integration tests


storage/interceptors/broker_notifications_interceptor.go
```diff
@@ -164,19 +158,79 @@ func fetchBrokerPlans(ctx context.Context, brokerID string, repository storage.R
 	return objList.(*types.ServicePlans).ServicePlans, nil
 }
 
-func getSupportedPlatformsForPlans(plans []*types.ServicePlan) []string {
-	platformTypes := make(map[string]bool)
+func ResolveSupportedPlatformIDsForPlans(ctx context.Context, plans []*types.ServicePlan, repository storage.Repository) ([]string, error) {
+	var platformTypes map[string]bool
+	platformNames := make(map[string]bool)
```
NickyMateev, line 163: Why create `map[string]bool` instead of `[]string` for both `platformTypes` and `platformNames`? You don't do anything special with these and you immediately after that convert them to slices anyway.

storage/interceptors/broker_notifications_interceptor.go
```diff
@@ -164,19 +158,79 @@ func fetchBrokerPlans(ctx context.Context, brokerID string, repository storage.R
 	return objList.(*types.ServicePlans).ServicePlans, nil
 }
 
-func getSupportedPlatformsForPlans(plans []*types.ServicePlan) []string {
-	platformTypes := make(map[string]bool)
+func ResolveSupportedPlatformIDsForPlans(ctx context.Context, plans []*types.ServicePlan, repository storage.Repository) ([]string, error) {
+	var platformTypes map[string]bool
```
NickyMateev, line 162: Why don't you initialize this like you do with the `platformNames`?

I see that's because you're differentiating whether the struct is not initialized (equal to nil) and when it's simply empty (not equal to nil, but len == 0) and executing some branching logic based on that.

But this makes the code inconsistent and hard to read. Please unify the variable initialization and instead make decisions based on if `len == 0`, rather than if `== nil`.

storage/interceptors/broker_notifications_interceptor.go
```diff
@@ -164,19 +158,79 @@ func fetchBrokerPlans(ctx context.Context, brokerID string, repository storage.R
...
-		for _, t := range types {
-			platformTypes[t] = true
+		planSupportedPlatformNames := plan.SupportedPlatformNames()
+		if planSupportedPlatformNames == nil {
+			// no explicit supported platform names defined - collect the supported platform types
+			if platformTypes == nil {
+				//only initialize this map if any plan not specifying explicit platform names is found
+				platformTypes = make(map[string]bool)
+			}
+
```
NickyMateev, line 172: This case is flawed. Imagine the scenario where a plan has neither supported platform names, nor supported platform types -> result is the `platformType` map will get initialized.

The next plan however has only support platform type `kubernetes` -> result is the platformTypes will now contain on element. As a result, the below code will only fetch kubernetes platforms instead of all platforms (due to the previous plan).

IMO, the loop should terminate at the first occurrence when a plan with no supported types/names is encountered, no further processing should be made other than fetching all platform IDs.

Also as my previous comments mentions we should get rid of these nil checks as they cause unforeseen/hidden bugs like the one that was about to happen here.

storage/interceptors/broker_notifications_interceptor.go
```diff
@@ -164,19 +158,79 @@ func fetchBrokerPlans(ctx context.Context, brokerID string, repository storage.R
 	return objList.(*types.ServicePlans).ServicePlans, nil
 }
 
-func getSupportedPlatformsForPlans(plans []*types.ServicePlan) []string {
-	platformTypes := make(map[string]bool)
+func ResolveSupportedPlatformIDsForPlans(ctx context.Context, plans []*types.ServicePlan, repository storage.Repository) ([]string, error) {
```
NickyMateev, line 161: It's weird for other packages to have to import the `interceptors` package to use this function. Did you try moving this somewhere in the `pkg` package?

test/osb_test/catalog_test.go
```diff
@@ -429,7 +430,12 @@ type prefixedBrokerHandler struct {
 
 func (pbh *prefixedBrokerHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	if strings.HasPrefix(req.URL.Path, pbh.brokerPathPrefix) {
-		common.SetResponse(w, http.StatusOK, common.Object{"services": common.Array{}})
+		servicesMap := make(map[string]interface{})
+		err := json.Unmarshal([]byte(simpleCatalog), &servicesMap)
```
NickyMateev, line 434: Why is this necessary?

test/public_plans_interceptor_test/public_plans_test.go
```diff
@@ -488,4 +489,96 @@ var _ = Describe("Service Manager Public Plans Interceptor", func() {
 			})
 		})
 	})
+
```
NickyMateev, line 492: What about tests where there's a mix - a plan has supported platform types and another has supported platform names? This is a realistic scenario that will happen.

test/public_plans_interceptor_test/public_plans_test.go
```diff
@@ -488,4 +489,96 @@ var _ = Describe("Service Manager Public Plans Interceptor", func() {
...
+	Context("when a plan has specified supported platform IDs", func() {
```
NickyMateev, line 493: _supported platform **names**_

test/osb_test/catalog_test.go
eyalnathan: the current test registers a broker with empty catalog and expects a notification to be created in the DB, while with the new code no notifications are created in this case - there are no plans to notify any platforms about. 
In order to keep the test for broker credentials change while there is a notification in the DB i needed to adapt it so a notification will be created by adding the catalog content.

pkg/types/service_plan.go
```diff
@@ -102,27 +102,58 @@ func (e *ServicePlan) Validate() error {
 		}
 	}
 
+	if e.SupportedPlatformIDs() != nil && e.SupportedPlatformTypes() != nil {
```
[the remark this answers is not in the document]
eyalnathan: fixed, returning empty arrays

pkg/types/service_plan.go
```diff
@@ -102,27 +101,34 @@ func (e *ServicePlan) Validate() error {
...
+// If there are no supported platforms names, nil is returned
+func (e *ServicePlan) SupportedPlatformNames() []string {
+	return e.metadataPropertyAsStringArray("supportedPlatformNames")
+}
 
-	return len(platforms) == 0 || slice.StringsAnyEquals(platforms, platform)
+func (e *ServicePlan) metadataPropertyAsStringArray(propertyKey string) []string {
+	propertyValue := gjson.GetBytes(e.Metadata, propertyKey)
+	if !propertyValue.IsArray() || len(propertyValue.Array()) == 0 {
+		return nil
```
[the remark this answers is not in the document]
eyalnathan: fixed, returning empty array instead of nil

pkg/types/service_plan_utils.go
```diff
@@ -0,0 +1,44 @@
...
+	if platformNames == nil {
+		return e.SupportsPlatformType(platform.Type)
+	} else {
+		return slice.StringsAnyEquals(platformNames, platform.Name)
+	}
+}
+
+// SupportsAllPlatforms determines whether the plan supports all platforms
+func (e *ServicePlan) SupportsAllPlatforms() bool {
+	return len(e.SupportedPlatformNames()) == 0 && len(e.SupportedPlatformTypes()) == 0
```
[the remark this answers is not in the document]
eyalnathan: fixed, all use len(..)==0 now

pkg/types/service_plan_utils.go
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package types
+
+import (
+	"github.com/Peripli/service-manager/pkg/util/slice"
+)
+
+// SupportsPlatform determines whether a specific platform type is among the ones that a plan supports
+func (e *ServicePlan) SupportsPlatformType(platform string) bool {
```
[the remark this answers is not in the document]
eyalnathan: fixed

storage/interceptors/broker_notifications_interceptor.go
```diff
@@ -164,19 +158,79 @@ func fetchBrokerPlans(ctx context.Context, brokerID string, repository storage.R
...
+	platformIDs := make(map[string]bool)
+
+	if len(platformNames) != 0 {
+		// fetch IDs of platform instances with the requested names
+		supportedPlatformNames := make([]string, 0)
+		for name := range platformNames {
+			supportedPlatformNames = append(supportedPlatformNames, name)
+		}
+		var criteria []query.Criterion
+		if len(supportedPlatformNames) != 0 {
```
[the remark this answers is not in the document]
eyalnathan: i refactored this function, this check is no longer there

storage/interceptors/broker_notifications_interceptor.go
eyalnathan: This is a "set" used tto collect them from multiple plans. That was how it was implemented before so i followed the pattern.

storage/interceptors/broker_notifications_interceptor.go
eyalnathan: refactored, please have a look at the new version

storage/interceptors/broker_notifications_interceptor.go
eyalnathan: good catch! fixed (i think)

storage/interceptors/broker_notifications_interceptor.go
eyalnathan: moved to storage/service_plans

test/public_plans_interceptor_test/public_plans_test.go
eyalnathan: fixed

storage/service_plans/resolver.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/Peripli/service-manager/pkg/query"
+	"github.com/Peripli/service-manager/pkg/types"
+	"github.com/Peripli/service-manager/storage"
+)
+
+func ResolveSupportedPlatformIDsForPlans(ctx context.Context, plans []*types.ServicePlan, repository storage.Repository) ([]string, error) {
+	var platformTypes map[string]bool
```
NickyMateev, line 12: I still think this `platformTypes` variable can get initialized inline here instead of on line 25: https://github.com/Peripli/service-manager/blob/a4a3388f2a0d91717b704453171296e2cf8798f7/storage/service_plans/resolver.go#L25

Then you can move the temporary `supportedPlatformTypes` variable below inside the actual branch which uses it and be symmetrical to the `if len(platformNames) != 0 ` check above in terms of flow:

```go
//add a criteria for the supported types
if len(platformTypes) != 0 {
	supportedPlatformTypes := make([]string, 0)
	for platform := range platformTypes {
		supportedPlatformTypes = append(supportedPlatformTypes, platform)
	}

	criteria = []query.Criterion{query.ByField(query.InOperator, "type", supportedPlatformTypes...)}
}
```

eyalnathan: indeed it looks better this way. Thanks!

test/public_plans_interceptor_test/public_plans_test.go
DanFrei: Hi,

These tests are implemented in sm-sap.
Kindly see(within: notifications_test.go):

Context("Define a catalog with mixed metadata definitions: one plan with 'supportedPlatforms' and another with 'supportedPlatformNames'"

Context("Define a catalog illegally having metadata of both supportedPlatforms and supportedPlatformNames within a single plan"

test/common/common.go
```diff
@@ -310,6 +311,15 @@ func RegisterBrokerInSM(brokerJSON Object, SM *SMExpect, headers map[string]stri
 		WithJSON(brokerJSON).Expect().Status(http.StatusCreated).JSON().Object().Raw()
 }
 
+func RegisterBrokerInSMWithWithRawResponse(brokerJSON Object, SM *SMExpect, headers map[string]string) *httpexpect.Response  {
+	request := SM.POST(web.ServiceBrokersURL).
+		WithHeaders(headers).
+		WithJSON(brokerJSON)
+	response := request.Expect()
+
+	return response
```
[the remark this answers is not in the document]
DanFrei: By placing 'request.Expect()' within a 'response' variable and then retrieving - makes it much clearer to the reader that we virtually retrieve a Response (whereas  'return request.Expect()'  doesn't implify it well enough).
On the other hand,  simply returning request.Expect() is shorter + one could see what the function retrieves in the return value.
So I changed it per your request.

test/common/test_context.go
```diff
@@ -633,7 +633,7 @@ func (ctx *TestContext) RegisterBrokerWithRandomCatalogAndTLS(expect *SMExpect)
 
 }
 
-func (ctx *TestContext) RegisterBrokerWithCatalogAndLabelsExpect(catalog SBCatalog, brokerData Object, expect *SMExpect) *BrokerUtils {
+func (ctx *TestContext) RegisterBrokerWithCatalogAndLabelsExpect(catalog SBCatalog, brokerData Object, expect *SMExpect) (error, *BrokerUtils) {
```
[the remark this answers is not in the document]
DanFrei: Thanks!!! fixed!

test/common/common.go
```diff
@@ -310,6 +312,15 @@ func RegisterBrokerInSM(brokerJSON Object, SM *SMExpect, headers map[string]stri
 		WithJSON(brokerJSON).Expect().Status(http.StatusCreated).JSON().Object().Raw()
 }
 
+func RegisterBrokerInSMWithoutErrorSwallowing(brokerJSON Object, SM *SMExpect, headers map[string]string) *httpexpect.Response {
```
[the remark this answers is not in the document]
DanFrei: Done!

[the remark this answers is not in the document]
DanFrei: Agree, fixed.

test/common/test_context.go
```diff
@@ -659,7 +659,18 @@ func (ctx *TestContext) RegisterBrokerWithCatalogAndLabelsExpect(catalog SBCatal
 	}
 
 	MergeObjects(brokerJSON, brokerData)
-	broker := RegisterBrokerInSM(brokerJSON, expect, map[string]string{})
+	response := RegisterBrokerInSMWithoutErrorSwallowing(brokerJSON, expect, map[string]string{})
+	responseStatusCode := response.Raw().StatusCode
+	if http.StatusCreated != responseStatusCode {
```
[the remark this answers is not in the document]
DanFrei: IMHO, this is needed since in registerBrokerWithCatalog function (notifications_suite-test.go), we call GetBrokerAsParams() on the retrieved BrokerUtils object.
Hadn't we had this error handling, we would've needed to check whether BrokerUtils isn't nil and return a null value for 'id' and 'BrokerServer' elswise!
Hence, I think that having this error handling logic is more allegiant and clearer than the aforementioned alternative.

test/broker_test/broker_test.go
```diff
@@ -535,6 +535,13 @@ var _ = test.DescribeTestsFor(test.TestCase{
 								r.Status(http.StatusBadRequest).JSON().Object().Keys().NotContains("services", "credentials")
 							}, "services.0.plans.0.schemas", "{invalid")
 						})
+
+						Context("that has both supportedPlatforms and supportedPlatformNames", func() {
```
NickyMateev, line 539: The positive test is missing (there were two in sm-sap)

eyalnathan: This is already tested in the other tests of the supported platforms feature, do you think we need a dedicated one here as well?

NickyMateev: Yes, I thought it was appropriate to have a dedicated test, also why introduce it internally in the first place then?

eyalnathan: tests added

PR #470, opened 2020-03-20
SMaaP to do async unbind only when binding is retrievable

KirilKabakchiev:


KirilKabakchiev:
Yes, I validated the cleanup and prepared a PR

PR #472, opened 2020-03-23
Change filter order of plugins

desislavaa:
## Motivation

When we want to register one plugin after another, the plugins are decomposed and their filters are ordered with respect to the implemented plugin interfaces.

Let's say we have three plugins: `First`, `Second` and `Third`.
`First` implements `Provisioner`, `InstanceUpdater` and `Deprovisioner`
`Second` implements only `Provisioner`
`Third` implements `Provisioner` and `Deprovisioner`. 

Currently, if we want to register the plugins as `First`, `Second` and then `Third` using the `RegisterPluginsBefore` method, their filters will have the following order:
- `First:Provision`
- `Second:Provision`
- `Third:Provision`
- `Third:Deprovision`
- `First:InstanceUpdater`
- `First:Deprovision`

## Approach

This PR changes the ordering logic as it places all filters that belong to a plugin one after another, and does not allow filters from other plugins between them.

## Pull Request status

* [x] Refactoring
* [x] Unit tests
* [ ] Integration tests

### DIFF
--- a/pkg/web/api.go
+++ b/pkg/web/api.go
@@ -167,8 +167,8 @@ func (api *API) RegisterPluginsBefore(beforePluginName string, plugins ...Plugin
 		}
 
 		pluginSegments := api.decomposePluginOrDie(plugin)
+		beforePluginFirstFilterName := api.getFirstFilterNameOfPlugin(beforePluginName)
 		for _, pluginSegment := range pluginSegments {
-			beforePluginFirstFilterName := api.getFirstFilterNameOfPlugin(beforePluginName)
 			api.registerFilterRelatively(beforePluginFirstFilterName, pluginSegment, func(beforeFilterPosition int) int {
 				return beforeFilterPosition
 			})
--- a/pkg/web/api.go
+++ b/pkg/web/api.go
@@ -168,12 +168,23 @@ func (api *API) RegisterPluginsBefore(beforePluginName string, plugins ...Plugin
 
 		pluginSegments := api.decomposePluginOrDie(plugin)
 		for _, pluginSegment := range pluginSegments {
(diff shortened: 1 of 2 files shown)


pkg/web/api.go
```diff
@@ -168,14 +168,25 @@ func (api *API) RegisterPluginsBefore(beforePluginName string, plugins ...Plugin
 
 		pluginSegments := api.decomposePluginOrDie(plugin)
 		for _, pluginSegment := range pluginSegments {
-			split := strings.Split(pluginSegment.Name(), ":")
-			api.registerFilterRelatively(beforePluginName+":"+split[1], pluginSegment, func(beforeFilterPosition int) int {
+			beforePluginFirstFilterName := api.getFirstFilterNameOfPlugin(beforePluginName)
```
pankrator: This line can be moved outside the `for` on the plugin segments

pkg/web/api_test.go
```diff
@@ -91,23 +91,43 @@ var _ = Describe("API", func() {
...
+					"existing-plugin:Provision",
+					"existing-plugin:UpdateService",
+					"existing-plugin:Deprovision",
+					"existing-plugin:FetchBinding",
+					"existing-plugin:Bind",
 					"existing-plugin:Unbind",
 				}))
 			})
+
+			When("the existing plugin does not implement all plugin interfaces", func() {
```
KirilKabakchiev: as far as i understand we fix the scenario with "transitive relativity" in which the middle plugin is missing a method X that the first and the third plugin have and because of that the third plugin method X will run before the first plugin method X. Does it make sense to add an actual test for that?

PR #473, opened 2020-03-24
Configure default http client timeout

pankrator:
## Motivation

There are a few different aspects of these timeouts:

1. OSB calls via the /v1/osb API of Service Manager:
  - Calls for get catalog are done via CatalogFetcher which uses the default http client which was not configured.
  - Calls for all other requests are done via reverse proxy, which does not know when it should cancel the request, because the context is not cancelled on server.request_timeout. The only time it will time out is when the headers are not yet received. But if there is a "bad broker" which has returned headers, but is streaming the response very slowly, we will wait for the response forever. This can be fixed with something like [http.TimeoutHandler](https://golang.org/pkg/net/http/#TimeoutHandler). There is a branch with implementation: https://github.com/Peripli/service-manager/tree/handler_timeouts.
2. OSB calls via SM Write APIs
  - Calls are done via osb client which is initialized with http client that has configured timeout for the whole request, so they do not suffer the above problem

If you need to reproduce some of the scenarios let me know and I will attach the broker servers and scripts to reproduce.

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [x] Unit tests

### DIFF
--- a/pkg/httpclient/settings.go
+++ b/pkg/httpclient/settings.go
@@ -25,6 +25,7 @@ import (
 )
 
 type Settings struct {
+	Timeout               time.Duration `mapstructure:"timeout" description:"timeout specifies a time limit for the request. The timeout includes connection time, any redirects, and reading the response body"`
 	TLSHandshakeTimeout   time.Duration `mapstructure:"tls_handshake_timeout"`
 	IdleConnTimeout       time.Duration `mapstructure:"idle_conn_timeout"`
 	ResponseHeaderTimeout time.Duration `mapstructure:"response_header_timeout"`
@@ -35,6 +36,7 @@ type Settings struct {
 // DefaultSettings return the default values for httpclient settings
 func DefaultSettings() *Settings {
 	return &Settings{
+		Timeout:               time.Second * 15,
 		TLSHandshakeTimeout:   time.Second * 10,
 		IdleConnTimeout:       time.Second * 10,
 		ResponseHeaderTimeout: time.Second * 10,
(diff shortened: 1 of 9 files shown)


test/common/application.yml
```diff
@@ -3,10 +3,11 @@ server:
   shutdown_timeout: 4000ms
   port: 1234
 httpclient:
-  response_header_timeout: 40000ms
-  tls_handshake_timeout: 4000ms
-  idle_conn_timeout: 4000ms
-  dial_timeout: 4000ms
+  timeout: 1000ms
```
NickyMateev: Aren't these values too low?

test/broker_test/broker_test.go
```diff
@@ -571,6 +572,25 @@ var _ = test.DescribeTestsFor(test.TestCase{
 						Expect(transitiveResourcesActualCount).To(Equal(transitiveResourcesExpectedCount))
 					})
 				})
+
+				Context("when broker is responding slow", func() {
+					It("should timeout", func() {
+						brokerServer.CatalogHandler = func(rw http.ResponseWriter, req *http.Request) {
+							rw.WriteHeader(http.StatusOK)
+							if fl, ok := rw.(http.Flusher); ok {
+								for i := 0; i < 30; i++ {
```
NickyMateev: I see we get very innovative with our testing as time goes on :D

test/common/application.yml
pankrator: mm may be they will be low for Travis tests... I will try to adjust them if they fail.

test/broker_test/broker_test.go
pankrator: Will this one get in unnoticed :D

KirilKabakchiev: noticed ! :)

pkg/httpclient/settings.go
```diff
@@ -64,6 +69,7 @@ func (s *Settings) Validate() error {
 func Configure(settings *Settings) {
 	transport := http.DefaultTransport.(*http.Transport)
 
+	http.DefaultClient.Timeout = settings.Timeout
```
KirilKabakchiev: we use the default transport in some places in the code (without the default client).  Have you checked if what we configured on it is sufficient ?

PR #478, opened 2020-03-30
Respect maximum polling duration from catalog

DimitarPetrov:
## Motivation

Currently service manager polls a service instance/binding for a fixed interval of time specified in the environment. OSB allows each broker to say how long the platform to poll for.

## Approach
By OSB spec, each plan can provide a property maximum_polling_duration which indicates how long an async operation for instance or binding of this plan should be polled at max. 

SM has it's own action timeout which represents how long we wait for an asynchronous action for the current operation to finish.
Polling of a service instance/binding can span multiple actions if it reached the point where the operation became reschedulable. The action timeout of SM could elapse, but since the operation is reschedulable it will be rescheduled by the maintainer and the polling will continue.

With that said, maximum_polling_duration of the plan should span multiple action schedules and should start at the time when the operation became reschedulable for the first time. To achieve that a new column reschedule_timestamp is added which indicates what is the first time that this operation became reschedulable.

When an operation is scheduled and being polled, we get this time and subtract the time elapsed from the max_polling_duration and poll until either action timeout is hit or maximum_polling_duration is hit.


test/service_binding_test/service_binding_test.go
```diff
@@ -675,6 +677,115 @@ var _ = DescribeTestsFor(TestCase{
 								})
 
 								if testCase.async {
+									When("maximum polling duration is reached while polling", func() {
+										var oldCtx *TestContext
```
KirilKabakchiev: this is a buggy approach to do context switching because it swaps out the valid Basic auth credentials - i've tried to fix this bug in the transfer instance PR and the tests there. (Idea is that after doing the context swap and reverting to the old context, its basic credentials wont work

test/service_binding_test/service_binding_test.go
```diff
@@ -675,6 +677,115 @@ var _ = DescribeTestsFor(TestCase{
 								})
 
 								if testCase.async {
```
KirilKabakchiev: why are these in testcase async only?

test/service_binding_test/service_binding_test.go
```diff
@@ -675,6 +677,115 @@ var _ = DescribeTestsFor(TestCase{
...
+									When("maximum polling duration is reached while polling", func() {
```
KirilKabakchiev: similar tests for unbind(delete binding) polling timeout reached ?

test/service_instance_test/service_instance_test.go
```diff
@@ -621,6 +631,77 @@ var _ = DescribeTestsFor(TestCase{
 								})
 
 								if testCase.async {
+									When("maximum polling duration is reached while polling", func() {
+										var oldCtx *TestContext
+
+										BeforeEach(func() {
+											preparePrerequisitesWithMaxPollingDuration(MaximumPollingDuration)
+
+											oldCtx = ctx
```
KirilKabakchiev: fix the context thingy (the comment from bindings applies here)

test/service_instance_test/service_instance_test.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"github.com/Peripli/service-manager/operations"
 	"github.com/Peripli/service-manager/pkg/query"
+	"github.com/tidwall/sjson"
```
KirilKabakchiev: all comment from the bindings test file apply here as well

pkg/types/service_plan.go
```diff
@@ -101,7 +102,9 @@ func (e *ServicePlan) Validate() error {
 			return fmt.Errorf("service plan metadata is invalid JSON")
 		}
 	}
-
+	if time.Duration(e.MaximumPollingDuration)*time.Second > 7*24*time.Hour {
```
KirilKabakchiev: is this value coming from OSB spec or we chose it ?

test/service_binding_test/service_binding_test.go
KirilKabakchiev: do we cover both cases with tetsts currently ?
a case testing a situation in which the action timeout is smaller than polling timeout and verifying that we actually eventually fail the op due to action timeout reached and not due to action timeout. 
And also vise versa action timeout > polling timeout => fail due to polling timeout
If yes there need to be proper BDD context separation indicating that, if not let's add the missing one and also bdd context  separation

test/service_binding_test/service_binding_test.go
```diff
@@ -675,6 +677,115 @@ var _ = DescribeTestsFor(TestCase{
...
+
+											brokerServer.BindingHandlerFunc(http.MethodPut, http.MethodPut+"1", ParameterizedHandler(http.StatusAccepted, Object{"async": true}))
+											brokerServer.BindingLastOpHandlerFunc(http.MethodPut+"1", ParameterizedHandler(http.StatusOK, Object{"state": "in progress"}))
+										})
+
+										AfterEach(func() {
+											ctx = oldCtx
+										})
+
+										When("orphan mitigation unbind synchronously succeeds", func() {
```
KirilKabakchiev: what about async orphan mitigation ? i think there were async orphan mitigation tests somewhere in the thousands of lines

pkg/types/service_plan.go
DimitarPetrov: We choose it

storage/interceptors/smaap_service_binding_interceptor.go
```diff
@@ -537,6 +568,23 @@ func (i *ServiceBindingInterceptor) pollServiceBinding(ctx context.Context, osbC
 	}
 }
 
+func (i *ServiceBindingInterceptor) processMaxPollingDurationElapsed(ctx context.Context, binding *types.ServiceBinding, operation *types.Operation, enableOrphanMitigation bool) error {
+	log.C(ctx).Errorf("Terminating poll last operataion for binding with id %s and name %s due to maximum_polling_duration for it's plan is reached", binding.ID, binding.Name)
```
nyordanoff: Minor: typo in `operataion`

pkg/types/service_plan.go
KirilKabakchiev: does it make sense for it to be configurable? is there a situation in which we want to make it smaller while troubleshooting an issue or as a mitigation of a "buggy" situation? I cant think of an example thats why this is more of a question

PR #480, opened 2020-04-01
Handler timeouts

pankrator:
## Motivation

If a certain request times out, this does not mean that the corresponding context will we cancelled, especially true for `server.WriteTimeout`. The mentioned timeout only closes the response writer, so that the next time someone tries to write, it is not propagated on the socket and just then the request will be closed on the other side.
With this PR we will be cancelling the request context after this write timeout has expired and we will return appropriate status code and body immediately, thus releasing the caller from waiting and also stopping any long running operations on our side.

## Pull Request status

* [x] Initial implementation
* [x] Component tests


pkg/util/api.go
```diff
@@ -146,7 +146,7 @@ func NewJSONResponse(code int, value interface{}) (*web.Response, error) {
 
 func NewJSONResponseWithHeaders(code int, value interface{}, additionalHeaders map[string]string) (*web.Response, error) {
 	headers := http.Header{}
-	headers.Add("Content-Type", "application/json")
+	headers.Set("Content-Type", "application/json")
```
KirilKabakchiev, line 149: any particular reason to Set and not Add ?

pkg/sm/sm.go
```diff
@@ -165,7 +166,9 @@ func New(ctx context.Context, cancel context.CancelFunc, e env.Environment, cfg
 	}
 
 	operationMaintainer := operations.NewMaintainer(ctx, interceptableRepository, postgresLockerCreatorFunc, cfg.Operations, waitGroup)
-	osbClientProvider := osb.NewBrokerClientProvider(cfg.HTTPClient.SkipSSLValidation, int(cfg.HTTPClient.ResponseHeaderTimeout.Seconds()))
+	osbClientTimeout := math.Min(float64(cfg.HTTPClient.Timeout), float64(cfg.Server.RequestTimeout))
```
KirilKabakchiev, line 169: you subtract a second in the timeout handler so it can write before the deadline, so if server timeout is picked here the timeout handler will timeout one second before the request is interrupted. if the request finishes within this second, probably a 2nd write attempt might be done.

pkg/server/server.go
```diff
@@ -97,20 +97,42 @@ func registerControllers(API *web.API, router *mux.Router, config *Settings) {
 		for _, route := range ctrl.Routes() {
 			log.D().Debugf("Registering endpoint: %s %s", route.Endpoint.Method, route.Endpoint.Path)
 			handler := web.Filters(API.Filters).ChainMatching(route)
-			router.Handle(route.Endpoint.Path, api.NewHTTPHandler(handler, config.MaxBodyBytes)).Methods(route.Endpoint.Method)
+			apiHandler := api.NewHTTPHandler(handler, config.MaxBodyBytes)
+			if !route.DisableHTTPTimeouts {
+				router.Handle(route.Endpoint.Path, newContentTypeHandler(http.TimeoutHandler(apiHandler, config.RequestTimeout, `{"error":"Timeout", "description": "operation has timed out"}`))).Methods(route.Endpoint.Method)
```
KirilKabakchiev, line 102: should this timeout be min(config.RequestTimeout, servertimeout-1sec) ? Or we should have config validation that servertimeout is bigger than request timeout

pkg/sm/sm.go
KirilKabakchiev, line 169: lets say server request timeout is picked here and timeout handler just expired and wrote, will this try error happening within a request flow also try writing? what about the other way around? is this ok ?

pkg/util/api.go
pankrator: My thinking here is, because if this is called multiple times, Set will replace the header, but Add will add another value. For example for cookies you can set multiple, but for Content-Type you shouldn't be able to do this

pkg/sm/sm.go
pankrator: It won't fail, because the ResponseWriter in our handler is replaced, by another writer, which if has already timedOut, will not attempt to write to the socket. See the implementation of TimeoutHandler.  I can still try it of course

pkg/server/server.go
pankrator: May be validation of the configuration will be best

pankrator: actually, if you check the creation of the Server struct, I have already put there for WriteTimeout +1, this will make sure that we have time to write the response before completely closing the response.

pkg/web/controller.go
```diff
@@ -29,6 +29,11 @@ type Route struct {
 
 	// Handler is the function that should handle incoming requests for this endpoint
 	Handler HandlerFunc
+
+	// DisableHTTPTimeouts if false (by defaut) will place timeout handler for the endpoint.
```
nyordanoff, line 33: Minor: typo in ` defaut`

PR #481, opened 2020-04-03
Add ListNoLabels to storage

georgifarashev:
# Add ListNoLabels to storage

## Motivation

In some cases fetching entities is required but their labels are not used.

## Approach

Introduce a new db api that fetches entities without labels.

## Pull Request status

* [x] Implementation
* [x] Tests

### DIFF
--- a/storage/interceptable_repository.go
+++ b/storage/interceptable_repository.go
@@ -180,11 +180,26 @@ func (ir *queryScopedInterceptableRepository) GetForUpdate(ctx context.Context, 
 }
 
 func (ir *queryScopedInterceptableRepository) List(ctx context.Context, objectType types.ObjectType, criteria ...query.Criterion) (types.ObjectList, error) {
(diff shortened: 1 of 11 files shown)


storage/interceptable_repository.go
```diff
@@ -180,11 +180,26 @@ func (ir *queryScopedInterceptableRepository) GetForUpdate(ctx context.Context,
 }
 
 func (ir *queryScopedInterceptableRepository) List(ctx context.Context, objectType types.ObjectType, criteria ...query.Criterion) (types.ObjectList, error) {
-	objectList, err := ir.repositoryInTransaction.List(ctx, objectType, criteria...)
+	return ir.list(ctx, objectType, false, criteria...)
```
KirilKabakchiev, line 183: very minor - maybe true would be better here in a sense that you say list with true meaning include labels and false meaning exclude

PR #483, opened 2020-04-06
Allow Broker Platform Credential Rotation during full resync in K8S platforms

NickyMateev:
# Allow Broker Platform Credential Rotation during full resync in K8S platforms

## Motivation

Unlike CF installations, K8S clusters are very volatile environments and it's easier to delete broker resources there. In case a broker is deleted in K8S directly (or CF for that matter), the broker record and its broker platform credentials will still exist in SM. And whenever the K8S agent retriggers full resync it would fail to register the broker back in the platform as it would try to essentially rotate the broker platform credentials - something we do not allow. We only allow full resync to initially set broker platform credentials, but now we will allow this rotation for K8S platforms,

## Approach

Make broker platform credentials controller less restrictive.

## Pull Request status

* [x] Initial implementation
* [x] Integration tests

### DIFF
--- a/test/osb_test/catalog_test.go
+++ b/test/osb_test/catalog_test.go
@@ -444,6 +444,11 @@ var _ = Describe("Catalog", func() {
 					By("rotation of K8S credentials broker platform credentials should work")
 					k8sOSBClient.GET(osbURL + "/v2/catalog").
 						Expect().Status(http.StatusOK).JSON().Object().ContainsKey("services")
+
+					By("old K8S credentials broker platform credentials should not work")
+					k8sOSBClient.SetBasicCredentials(ctx, username, password)
+					k8sOSBClient.GET(osbURL + "/v2/catalog").
+						Expect().Status(http.StatusUnauthorized)
 				})
 			})
 		})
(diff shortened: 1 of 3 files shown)


test/osb_test/catalog_test.go
```diff
@@ -413,11 +413,37 @@ var _ = Describe("Catalog", func() {
 			})
 
 			Context("when broker platform credentials change out of notification processing context when already exist", func() {
-				It("should return 409", func() {
+				FIt("should return 409 for non-kubernetes platforms", func() {
```
pankrator, line 416: Remove the focus on the test

DimitarPetrov, line 416: `FIt` Probably leftover

test/osb_test/catalog_test.go
```diff
@@ -413,11 +413,37 @@ var _ = Describe("Catalog", func() {
...
+					By("initial K8S credentials should work")
+					k8sOSBClient.GET(osbURL + "/v2/catalog").
+						Expect().Status(http.StatusOK).JSON().Object().ContainsKey("services")
+
+					newUsername, newPassword = test.RegisterBrokerPlatformCredentialsExpect(k8sPlatformClient, prefixedBrokerID, http.StatusOK)
+					k8sOSBClient.SetBasicCredentials(ctx, newUsername, newPassword)
+
+					By("rotation of K8S credentials broker platform credentials should work")
+					k8sOSBClient.GET(osbURL + "/v2/catalog").
+						Expect().Status(http.StatusOK).JSON().Object().ContainsKey("services")
```
pankrator, line 446: Will this be enough to verify that the credentials are rotated?

NickyMateev: discussed

PR #487, opened 2020-04-07
Update instance check for visibility only if the plan of the instance is changed

DimitarPetrov:
### Motivation

Currently the check assumed that if the plan is not updated it will not be presented in the request body. However CC sends the plan_id in the body even if it is not changed and remains the same. SM should make a difference between this situations because if not this makes the instance some sort of unusable.

### DIFF
--- a/api/osb/check_instance_visibility_plugin.go
+++ b/api/osb/check_instance_visibility_plugin.go
@@ -58,13 +58,27 @@ func (p *checkVisibilityPlugin) UpdateService(req *web.Request, next web.Handler
 	if err := decodeRequestBody(req, requestPayload); err != nil {
 		return nil, err
 	}
-	if len(requestPayload.PlanID) == 0 { // plan is not changed
+	if len(requestPayload.PlanID) == 0 { // plan is not being updated
 		return next.Handle(req)
 	}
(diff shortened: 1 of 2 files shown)


api/osb/check_instance_visibility_plugin.go
```diff
@@ -58,13 +58,27 @@ func (p *checkVisibilityPlugin) UpdateService(req *web.Request, next web.Handler
...
 	planID, err := findServicePlanIDByCatalogIDs(ctx, p.repository, requestPayload.BrokerID, requestPayload.ServiceID, requestPayload.PlanID)
 	if err != nil {
 		return nil, err
 	}
+
+	byID := query.ByField(query.EqualsOperator, "id", requestPayload.InstanceID)
+	instanceObj, err := p.repository.Get(ctx, types.ServiceInstanceType, byID)
+	if err != nil {
+		if err == util.ErrNotFoundInStorage {
+			return next.Handle(req)
```
pankrator, line 73: May be in case the instance is not found, we should abort the request and return a meaningful message that the instance does not exist anymore?
Not sure if this can happen anyway.

DimitarPetrov: This is happening for the instances that exist, but not in SM because are not yet migrated.

This can be removed in the future when SM is source of truth, but even then it will be fairly complex, because CC and SM can get out of sync for various reasons. However now it is prerequisite for normal usage of instances not yet migrated and migration itself.

PR #489, opened 2020-04-14
Insert labels in batches to avoid pq param limit

dotchev:
pq has a limit of 65535 parameters, see https://github.com/lib/pq/blob/master/conn.go#L1282
now we insert labels in batches to avoid this limit

### DIFF
--- a/storage/postgres/storage.go
+++ b/storage/postgres/storage.go
@@ -217,9 +217,30 @@ func (ps *Storage) createLabels(ctx context.Context, labels []PostgresLabel) err
 		strings.Join(dbTags, ", :"),
 	)
 
-	log.C(ctx).Debugf("Executing query %s", sqlQuery)
-	_, err := ps.pgDB.NamedExecContext(ctx, sqlQuery, labels)
-	return checkIntegrityViolation(ctx, err)
+	// break into batches so the PostgreSQL limit of 65535 parameters is not exceeded
+	const maxParams = 65535
+	maxRows := maxParams / len(dbTags)
+	for len(labels) > 0 {
+		rows := min(len(labels), maxRows)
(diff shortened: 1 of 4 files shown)


storage/postgres/storage.go
```diff
@@ -217,9 +217,30 @@ func (ps *Storage) createLabels(ctx context.Context, labels []PostgresLabel) err
...
+	for len(labels) > 0 {
+		rows := min(len(labels), maxRows)
+		log.C(ctx).Debugf("Executing query %s", sqlQuery)
+		result, err := ps.pgDB.NamedExecContext(ctx, sqlQuery, labels[:rows])
+		if err != nil {
+			return checkIntegrityViolation(ctx, err)
+		}
+		rowsAffected, err := result.RowsAffected()
+		if err == nil {
+			log.C(ctx).Debugf("%d rows affected", rowsAffected)
```
pankrator, line 232: rows inserted will be better

storage/postgres/storage.go
```diff
@@ -217,9 +217,30 @@ func (ps *Storage) createLabels(ctx context.Context, labels []PostgresLabel) err
...
+	const maxParams = 65535
+	maxRows := maxParams / len(dbTags)
+	for len(labels) > 0 {
+		rows := min(len(labels), maxRows)
+		log.C(ctx).Debugf("Executing query %s", sqlQuery)
+		result, err := ps.pgDB.NamedExecContext(ctx, sqlQuery, labels[:rows])
+		if err != nil {
+			return checkIntegrityViolation(ctx, err)
+		}
+		rowsAffected, err := result.RowsAffected()
```
pankrator, line 230: Can this error be not nil? What will happen if there is err?

PR #492, opened 2020-04-15
Fix flappy storage test

dotchev:
The test should not depend on the order of label values


dotchev:
Already done with #490

PR #496, opened 2020-04-23
Fix duplicate json keys in body

DimitarPetrov:


pkg/util/api.go
```diff
@@ -75,17 +77,72 @@ func RequestBodyToBytes(request *http.Request) ([]byte, error) {
...
 
+func validJson(data []byte) error {
+	if !json.Valid(data) {
+		return fmt.Errorf("invalid json")
+	}
+	return checkDuplicateKeys(json.NewDecoder(bytes.NewReader(data)), nil)
+}
+
+func checkDuplicateKeys(d *json.Decoder, path []string) error {
+	t, err := d.Token()
```
KirilKabakchiev, line 101: you can also unmarshal and marshal i think - it would be probably slower but easier to maintain - not sure whether it makes sense

DimitarPetrov: Unmarshaling and Marshaling gets the last provided value for objects (if the value in which you unmarshal in is struct) and merge the values for duplicate keys if unmarshal in map which makes it impossible to know which is the right value, we can't know what json is provided to rely on this. It can be manipulated in many ways, this is why most of other language JSON parsers does not allow duplicate keys and I decided to go in this direction.

test/service_binding_test/service_binding_test.go
```diff
@@ -311,6 +311,47 @@ var _ = DescribeTestsFor(TestCase{
...
+							Context("when request body contains multiple label objects", func() {
+								It("returns 400", func() {
+									ctx.SMWithOAuthForTenant.POST(web.ServiceBindingsURL).
+										WithQuery("async", testCase.async).
+										WithHeader("Content-Type", "application/json").
+										WithBytes([]byte(fmt.Sprintf(`{
+										"name": "test-binding-name",
+										"service_instance_id": "%s",
+										"maintenance_info": {},
+										"labels": {},
```
dpanayotov, line 342: add one with `string` and an object; also a test for the context

pkg/util/api.go
```diff
@@ -75,17 +77,72 @@ func RequestBodyToBytes(request *http.Request) ([]byte, error) {
...
+			}
+			i++
+		}
+		// Consume trailing ]
+		if _, err := d.Token(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
dotchev, line 144: As this function contains quite some logic, it would be good to have a unit test for it
This will make it easier to test different corner cases

test/broker_test/broker_test.go
```diff
@@ -261,6 +263,52 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+								},
+								"labels": {},
+								"labels": {
+									"%s":["test-tenant"]
+								}	
+							}`, brokerWithLabelsServer.URL(), brokerWithLabelsServer.Username, brokerWithLabelsServer.Password, TenantLabelKey))).
+								Expect().
+								Status(http.StatusBadRequest).
+								JSON().Object().
+								Keys().Contains("error", "description")
```
dotchev, line 305: some more specific error check
how can we be sure the error is due to the duplicate labels property?

DimitarPetrov:
@dotchev 
> The same pair of tests are added for service brokers, service bindings and service instances.
Why these resources?

These are resources that can be tenant scoped and this is the most critical label which must not be manipulated, so I decided it is reasonable to have a test for each one of them. Also this is a protected label for them and we must have the confidence that it cannot be changed from outside.

> Wouldn't it be more generic if we add this as a global filter on all POST, PUT, PATCH requests?
Then it would not matter if we call RequestBodyToBytes.

Here it is implemented when we transform the Request to our implementation `web.Request` which looked like the appropriate place for me. Also generic json validation and content-type checks were here before this change, so this logic just enrich the check and makes it more restrictive.

I also tought about the global filter, but here looks more global, because it is how our framework works and the checks for json validity and duplicate keys will be separated apart which will complicate validation.

I am OK with both approaches.

dotchev:
ok, just add an unit test

PR #504, opened 2020-05-22
Sa termination

drorhayun:
# Pull Request Template

## Prerequisites

## Motivation

## Approach

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [x] Unit tests
* [x] Integration tests

## Third-party code

## Work in Progress label


pkg/types/cascade/interface.go
```diff
@@ -0,0 +1,61 @@
...
+func GetCascadeObject(ctx context.Context, object types.Object) (Cascade, bool) {
+	switch object.GetType() {
+	case types.TenantType:
+		return &TenantCascade{object.(*types.Tenant)}, true
+	case types.PlatformType:
+		return &PlatformCascade{object.(*types.Platform)}, true
+	case types.ServiceBrokerType:
+		return &ServiceBrokerCascade{object.(*types.ServiceBroker)}, true
+	case types.ServiceInstanceType:
+		containerIdKey := ctx.Value(ContainerKey{})
```
eyalnathan, line 53: containerIDKey (capital D)

eyalnathan: or containerLabelKey?

pkg/web/routes.go
```diff
@@ -53,4 +53,6 @@ const (
 
 	// ProfileURL is the Configuration API base URL path
 	ProfileURL = "/" + apiVersion + "/profile"
+
+	TenantURL = "/" + apiVersion + "/tenant"
```
eyalnathan, line 57: shouldn't this be /tenants? all other resource endpoints are using plural

storage/postgres/migrations/20200514181500_operations_cascade.up.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+
+ALTER TABLE operations ADD COLUMN IF NOT EXISTS cascade_root_id varchar(100);
```
eyalnathan, line 3: do we want to create index on the new columns? the "cascade_root_it" is used for queries in many flows

pavelmaliy: no sure, we're not looking for specific root_id only in cleanup
maybe it'll be better to index parent_id

PR #518, opened 2020-06-08
Resue context criteria for name uniqueness checks

eyalnathan:
## Motivation

When checking for service instance name uniqueness during creation or update in SMaaP, the scope for uniqueness is defined to be the tenant. Some use cases may need to limit the scope even further, to provide "namespaces" within the tenant. Such scoping is achieved by using custom labels (as in the case of the tenant scope), but the name uniqueness checks do not "respect" those labels. 

## Approach

This change changes the logic of the unique_instance_name_interceptor to re-use any labelQuery criteria that were added to the context, this way custom labelQueries added to support custom scopes will be included, making the scope for uniqueness check correct in such cases. 

## Pull Request status

* [x] Initial implementation
* [x] Integration tests - covered by existing test cases


dotchev:
Any tests?

eyalnathan:
> Any tests?

The tenant scope is covered by existing tests for both create and update instance flows, custom use cases shall be tested where they are implemented.

PR #526, opened 2020-06-16
Return last operation with list API calls

DanFrei:
## Motivation
Allow clients to obtain the last operation by examining the resource list endpoints in order to construct a meaningful list of results.

## Approach
1. A new named query is introduced to allow performant lookups for the last resource operations
2. Base controller implementation was updated to allow the return of last operations in list APIs
3. The Attach last operations and last operation implementation are unified
4. Added support 4 "in statements" with Named queries 

## Pull Request status
* [x] Initial implementation
* [ ] Refactoring
* [] Unit tests
* [x ] Integration tests


storage/interceptable_repository.go
```diff
@@ -121,6 +121,10 @@ func (ir *queryScopedInterceptableRepository) QueryForList(ctx context.Context,
 	return ir.repositoryInTransaction.QueryForList(ctx, objectType, queryName, queryParams)
 }
 
+func (ir *queryScopedInterceptableRepository) QueryForListWithInStatement(ctx context.Context, objectType types.ObjectType, queryName NamedQuery, queryParams []interface{}) (types.ObjectList, error) {
```
alexwo, line 124: remove not used anymore

storage/named_query.go
```diff
@@ -39,6 +41,21 @@ var namedQueries = map[NamedQuery]string{
...
+	FROM operations ops 
+    inner join
+		 (
+			 select max(op.paging_sequence) last_operation_sequence
+			 from operations op
+			 group by resource_id
+		 ) lastOperationPerResource
+		 on lastOperationPerResource.last_operation_sequence = ops.paging_sequence
+	WHERE resource_id in (:id_list)`,
+	QueryByTypeAndIDIn: `
```
eyalnathan, line 55: did you mean to push this query?

api/base_controller.go
```diff
@@ -509,22 +516,76 @@ func cleanObject(object types.Object) {
...
 	if err != nil {
-		return util.HandleStorageError(err, types.OperationType.String())
+		return err
 	}
-	if list.Len() == 0 {
-		log.C(ctx).Debugf("No last operation found for entity with id %s of type %s", objectID, object.GetType().String())
-		return nil
+
+	if lastOperation, ok := ops[objectID]; ok {
+		lastOperation.TransitiveResources = nil
```
eyalnathan, line 585: this is not needed in the list case? why is it done here?

application.yml
```diff
@@ -24,8 +24,7 @@ storage:
   skip_ssl_validation: false
   max_idle_connections: 5
 api:
-  token_issuer_url: http://localhost:8080/uaa
-  client_id: cf
+  token_issuer_url: http://localhost:8080/
```
eyalnathan, line 27: why is this changed?

test/query_test/query_test.go
```diff
@@ -60,6 +177,16 @@ var _ = Describe("Service Manager Query", func() {
 			id2 = createNotification(repository, now.Add(-30*time.Minute))
 		})
 
+		It("finds objects by ID in", func() {
```
eyalnathan, line 180: did you mean to push this?

test/service_instance_and_binding_test/service_binding_test/service_binding_test.go
```diff
@@ -224,6 +224,28 @@ var _ = DescribeTestsFor(TestCase{
 			})
 
 			Describe("GET", func() {
+
+				Context("when attach_last_operations is truthy", func() {
```
eyalnathan, line 228: shouldn't this be part of the generic tests for all types? it doesn't have any binding specific logic

application.yml
alexwo: Should not change, reverted it.

test/service_instance_and_binding_test/service_binding_test/service_binding_test.go
DanFrei: Fixed

storage/named_query.go
```diff
@@ -5,6 +5,8 @@ type NamedQuery int
 const (
 	QueryByMissingLabel NamedQuery = iota
 	QueryByExistingLabel
+	QueryForLastOperationsPerResource
+	QueryByTypeAndIDIn
```
eyalnathan, line 9: this should also be removed

test/list.go
```diff
@@ -486,6 +488,75 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase, responseMode Resp
...
+				Context("when the last operation does not exist", func() {
+					It("should return the resource without it", func() {
+						resp := ctx.SMWithOAuth.GET(t.API).
+							WithQuery("attach_last_operations", "true").
+							Expect().Status(http.StatusOK).
+							JSON()
+
+						resource := resp.Path("$.items[0]")
+						lastOp := resource.Object().Value("last_operation")
+						resourceId := lastOp.Object().Value("resource_id").String().Raw()
```
eyalnathan, line 522: why do you need the last operation for the resource ID? it is in $.items[0].id

test/list.go
```diff
@@ -486,6 +488,75 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase, responseMode Resp
...
+						criteria := query.ByField(query.EqualsOperator, "resource_id", resourceId)
+						err := ctx.SMRepository.Delete(context.Background(), types.OperationType, criteria)
```
eyalnathan, line 524: the setup (like deleting the last ops) is usually done in BeforeEach, and only the test itself in the It, this way it is much easier to understand what is being tested.

test/list.go
```diff
@@ -486,6 +488,75 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase, responseMode Resp
...
+							WithQuery("attach_last_operations", "true").
+							Expect().Status(http.StatusOK).
+							JSON()
+						for _, resource := range resp.Path("$.items[*]").Array().Iter() {
+							currentResourceId := resource.Object().Value("id").String().Raw()
+							currentResourceState := resource.Object().Value("ready").Boolean().Raw()
+							lastOp := resource.Object().Value("last_operation")
+
+							Expect(lastOp.Object().Value("resource_id").String().Raw()).To(Equal(currentResourceId))
+							Expect(lastOp.Object().Value("ready").Boolean().Raw()).To(Equal(currentResourceState))
```
eyalnathan, line 504: what is the point of comparing the "ready" field?

eyalnathan: and how do we know that this is really the last operation on that resource and not just one of them (not the last)?

test/query_test/query_test.go
```diff
@@ -50,6 +51,81 @@ var _ = Describe("Service Manager Query", func() {
...
+
+			list, err := repository.QueryForList(context.Background(), types.OperationType, storage.QueryForLastOperationsPerResource, queryParams)
+			Expect(err).ShouldNot(HaveOccurred())
+			lastOperation := list.ItemAt(0).(*types.Operation)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(list.Len()).To(BeEquivalentTo(1))
+			Expect(lastOperation.State).To(Equal(types.SUCCEEDED))
+		})
+
+		It("should return the correct last operation after a new one is added to a resource", func() {
```
eyalnathan, line 78:
```
When("new operation is created for the resource", func(){
  BeforeEach - do the setup (add operation)
  It("should return the new operation as last operation", func(){
     verify the expected result
  })
})
```

test/query_test/query_test.go
```diff
@@ -50,6 +51,81 @@ var _ = Describe("Service Manager Query", func() {
...
+			queryParams := map[string]interface{}{
+				"id_list": []string{serviceInstance1.ID},
+			}
+
+			list, err := repository.QueryForList(context.Background(), types.OperationType, storage.QueryForLastOperationsPerResource, queryParams)
+			Expect(err).ShouldNot(HaveOccurred())
+			lastOperation := list.ItemAt(0).(*types.Operation)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(list.Len()).To(BeEquivalentTo(1))
+			Expect(lastOperation.State).To(Equal(types.SUCCEEDED))
```
eyalnathan, line 75: is this enough to be sure that this is the last one? don't you want to compare IDs?

test/query_test/query_test.go
```diff
@@ -50,6 +51,81 @@ var _ = Describe("Service Manager Query", func() {
...
+				Base: types.Base{
+					ID:        "my_test_op_latest",
+					CreatedAt: time.Now(),
+					UpdatedAt: time.Now(),
+					Labels:    make(map[string][]string),
+					Ready:     true,
+				},
+				Description:       "my_test_op_latest",
+				Type:              types.CREATE,
+				State:             types.IN_PROGRESS,
```
eyalnathan, line 93: why IN_PROGRESS? i am worried some maintainer job might try to process it, maybe make it FAILED/SUCCEEDED

test/service_instance_and_binding_test/service_binding_test/service_binding_test.go
```diff
@@ -224,6 +224,7 @@ var _ = DescribeTestsFor(TestCase{
 			})
 
 			Describe("GET", func() {
+
```
eyalnathan, line 227: remove this

test/list.go
```diff
@@ -486,6 +488,75 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase, responseMode Resp
...
+								resource.Object().ContainsKey("last_operation")
+							}
+						}
+
+						Expect(foundResource).To(Equal(true))
+					})
+				})
+			}
+
+			Context("when attach_last_operations is falsy", func() {
```
eyalnathan, line 547: this should be in the if.SupportsAsync, otherwise it doesn't test much -  there are no operations anyway, right?

storage/named_query.go
DanFrei: Done

api/base_controller.go
```diff
@@ -509,22 +516,76 @@ func cleanObject(object types.Object) {
...
+			LastOp.TransitiveResources = nil
+			resource.SetLastOperation(LastOp)
+		}
+	}
+
+	return nil
+}
+
+func getLastOperations(ctx context.Context, resourceIDs []string, repository storage.Repository) (map[string]*types.Operation, error) {
+
```
eyalnathan, line 547: remove?

storage/named_query.go
```diff
@@ -39,6 +40,17 @@ var namedQueries = map[NamedQuery]string{
 	(SELECT ID FROM {{.LABELS_TABLE}} 
 				WHERE key=:key
 				AND {{.ENTITY_TABLE}}.{{.PRIMARY_KEY}} = {{.LABELS_TABLE}}.{{.REF_COLUMN}})`,
+	QueryForLastOperationsPerResource: `
```
eyalnathan, line 43: can we make this one look more like the neighbours? CAPITALS, {{.ENTITY_TABLE}} etc.

test/common/common.go
```diff
@@ -220,6 +220,20 @@ func MapContains(actual Object, expected Object) {
 	}
 }
 
+func RemoveResourceByCriterion(testCtx *TestContext, criterion query.Criterion, resourceType types.ObjectType) error {
+	if err := testCtx.SMRepository.InTransaction(context.TODO(), func(ctx context.Context, storage storage.Repository) error {
```
eyalnathan, line 224: Do you need to open the tx explicitly? won't it be created automatically in Delete?

eyalnathan: Sorry, looking in the IDE i would like to rephrase - why not use testCtx.SMRepository.Delete?

alexwo: yes this won't work as there will be an operation missing in the context

test/list.go
```diff
@@ -486,6 +489,130 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase, responseMode Resp
 				})
 			}
 
+			if t.SupportsAsyncOperations {
+				Context("when attach_last_operations is truthy", func() {
+					var resourceWithOneOperation, resourceWithFewOperations common.Object
+					var lastOperationForResourceId string
+					var resourceWithOneOperationId, resourceWithFewOperationsId string
+
+					createTestOperation := func(resourceID string, opType types.OperationCategory) (types.Object, string) {
```
eyalnathan, line 498: why the complex return value? seems like only the operation ID is used

test/list.go
```diff
@@ -486,6 +489,130 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase, responseMode Resp
...
+						resp := ctx.SMWithOAuth.GET(t.API).
+							WithQuery("attach_last_operations", "true").
+							Expect().Status(http.StatusOK).
+							JSON()
+
+						for _, resource := range resp.Path("$.items[*]").Array().Iter() {
+							itemResourceId := resource.Object().Value("id").String().Raw()
+							if itemResourceId == resourceWithOneOperationId || itemResourceId == resourceWithFewOperationsId {
+								lastOp := resource.Object().Value("last_operation")
+								resource.Object().ContainsKey("last_operation")
```
eyalnathan, line 550: switch order of the lines?

storage/named_query.go
alexwo: Do you see a use case for having ENTITY_TABLE variable in this statement?

test/list.go
```diff
@@ -486,6 +489,130 @@ func DescribeListTestsFor(ctx *common.TestContext, t TestCase, responseMode Resp
...
+
+					It("retrieves the resources list when each resource contains the last operation it is associated to", func() {
+						resp := ctx.SMWithOAuth.GET(t.API).
+							WithQuery("attach_last_operations", "true").
+							Expect().Status(http.StatusOK).
+							JSON()
+
+						for _, resource := range resp.Path("$.items[*]").Array().Iter() {
+							itemResourceId := resource.Object().Value("id").String().Raw()
+							if itemResourceId == resourceWithOneOperationId || itemResourceId == resourceWithFewOperationsId {
```
eyalnathan, line 548: are any other resources expected? if it is always just those 2 the if can be removed

test/list.go
alexwo: yes just copied this from a different test, can be simplified

test/list.go
alexwo: yes.. this test is going to be executed by different tests that prepare content that is not cleaned. actually we can't assume what resources will be stored in the database.

test/query_test/query_test.go
eyalnathan: i am still not sure that this is the right operation.. please don't resolve comments without making any change or leaving a reply

test/query_test/query_test.go
```diff
@@ -50,6 +51,85 @@ var _ = Describe("Service Manager Query", func() {
...
+				Expect(err).ShouldNot(HaveOccurred())
+			})
+			It("should return the new operation as last operation", func() {
+				queryParams := map[string]interface{}{
+					"id_list": []string{serviceInstance1.ID},
+				}
+				list, err := repository.QueryForList(context.Background(), types.OperationType, storage.QueryForLastOperationsPerResource, queryParams)
+				Expect(err).ShouldNot(HaveOccurred())
+				lastOperation := list.ItemAt(0).(*types.Operation)
+				Expect(err).ShouldNot(HaveOccurred())
```
eyalnathan, line 109: not needed

api/base_controller.go
alexwo: getLastOperations is used in attachLastOperations &  attachLastOperation, that's the common code that use's the named query for both methods.

test/query_test/query_test.go
```diff
@@ -50,6 +51,85 @@ var _ = Describe("Service Manager Query", func() {
...
+				Expect(list.Len()).To(BeEquivalentTo(1))
+				Expect(lastOperation.State).To(Equal(types.SUCCEEDED))
+				Expect(list.Len()).To(BeEquivalentTo(1))
+				Expect(lastOperation.ID).To(Equal("my_test_op_latest"))
+			})
+
+		})
+
+		It("should return the last operation for every instances in the query", func() {
```
eyalnathan, line 118: I would add another operation for a different resource, and make sure that i got only the operations for the resources that were requested, and not the other one. I think at the moment this is not tested.

test/query_test/query_test.go
alexwo: here we test that in case a resource has 2 operations (old / new), the query picks up the latest expected one.
should not we test it?

storage/named_query.go
eyalnathan: operations? but i get your point..

api/base_controller.go
eyalnathan: sorry, meant the empty line :)

test/query_test/query_test.go
eyalnathan: Expect(err).ShouldNot(HaveOccurred()) is already done 2 lines above, and there's no new assignment for err, so no point in checking again

test/query_test/query_test.go
alexwo: > i am still not sure that this is the right operation.. please don't resolve comments without making any change or leaving a reply

tests was extended with both:
Expect(lastOperation.State).To(Equal(types.SUCCEEDED))
Expect(lastOperation.Type).To(Equal(types.CREATE))

as we create the instance, we know that it should have exactly one created SUCCEEDED operation

eyalnathan: the other resource also has a single SUCCEEDED & created operation. you query for a specific resource, it makes sense to verify that you got the operation for that resource and not another. just add `Expect(lastOperation.ResourceID).To(Equal(serviceInstance1.ID))`.

alexwo: Ahh got it .. i thought you wanted to make sure that it returns that correct "last operation" of the instance in the case of multiple ops for the same resource. I will add an expect to confirm the instance id.. got it 👍

PR #528, opened 2020-06-23
Prevent deletion of operations being "last operation" for their corresponding instances

DanFrei:
# Pull Request Template

## Motivation

In order to be able to conduct an analysis on the various resources maintained in SM, one should have access to the latest operation of a given resource.
By current implementation, the maintainer cleanup tasks, periodically delete operations (come a certain condition/expiration time) thus some resources may remain without their corresponding last operation.
In this case, such resources cannot be properly monitored.

## Approach

Current cleanup tasks will be enhanced to avoid removal of operations being "last" for their corresponding resources.
This can be achieved by the following approach:
Extending SMs criteria-based "ORM" to support NOT EXISTS/EXSITS operation that will check each record with a sub-query retrieving the last operations.

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [ ] Unit tests
* [ ] Integration tests

## Work in Progress label


operations/maintainer.go
```diff
@@ -187,6 +187,11 @@ func (om *Maintainer) cleanupExternalOperations() {
 // cleanupFinishedCascadeOperations cleans up all successful/failed internal cascade operations which are older than some specified time
 func (om *Maintainer) CleanupFinishedCascadeOperations() {
 	currentTime := time.Now()
+	lastOpsList, err := om.repository.QueryForList(om.smCtx, types.OperationType, storage.QueryForAllLastOperations, map[string]interface{}{});
+	if err != nil {
+		log.C(om.smCtx).Debugf("Failed to cleanup operations: %s", err)
```
alexwo, line 192: Error message is incorrect

operations/maintainer.go
```diff
@@ -195,6 +200,7 @@ func (om *Maintainer) CleanupFinishedCascadeOperations() {
 		query.ByField(query.InOperator, "state", string(types.SUCCEEDED), string(types.FAILED)),
 		// check if operation hasn't been updated for the operation's maximum allowed time to live in DB//
 		query.ByField(query.LessThanOperator, "updated_at", util.ToRFCNanoFormat(currentTime.Add(-om.settings.Lifespan))),
+		query.ByField(query.NotInOperator, "id", getObjectListIDs(lastOpsList)...),
```
eyalnathan, line 209: will this work on real systems with content? assuming all instances have last operations, this could be ~100k IDs, not sure postgres will like that..

operations/maintainer.go
```diff
@@ -515,3 +543,16 @@ func (om *Maintainer) markStuckOperationsFailed() {
 
 	log.C(om.smCtx).Debug("Finished marking stuck operations as failed")
 }
+
+func getObjectListIDs(lastOpsList types.ObjectList) []string {
```
eyalnathan, line 547: this is not specific to last ops, and is already implemented in base_controller, maybe it would be better to extract to some common util? if not, is the nil check needed? and why notInIds?

operations/maintainer.go
DanFrei: That's an excellent point! thanks Eyal!!!!
We did some research for a more efficient alternative and found that using NOT EXISTS should be far better performance wise.
see here:
https://explainextended.com/2009/09/16/not-in-vs-not-exists-vs-left-join-is-null-postgresql/
https://stackoverflow.com/questions/16996002/postgresql-query-is-slow-when-using-not-in
https://dba.stackexchange.com/questions/196020/postgresql-optimize-query-with-where-not-in-clause

pkg/query/operators.go
```diff
@@ -33,6 +33,10 @@ const (
 	InOperator inOperator = "in"
 	// NotInOperator takes two operands and tests if the left is not contained in the right
 	NotInOperator notInOperator = "notin"
+	//
+	NotExistsOperator notExistsOperator = "notexists"
+	//
```
alexwo, line 38: comment on what it does

pkg/query/operators.go
```diff
@@ -33,6 +33,11 @@ const (
 	InOperator inOperator = "in"
 	// NotInOperator takes two operands and tests if the left is not contained in the right
 	NotInOperator notInOperator = "notin"
+	// NotExistsOperator receives a sub-query as single left-operand and checks the sub-query for rows existence, and if the NotExistsOperator receives a sub-query as single left-operand and checks the sub-query for rows existence,
+	// and if there are no rows then it are no rows then it will return TRUE, otherwise FALSE.
+	NotExistsOperator notExistsOperator = "notexists"
```
eyalnathan, line 38: maybe these operators can be private? they are only used in the package, and do not make much sense without the wrapping ByIDExist

pkg/query/selection.go
```diff
@@ -109,13 +111,23 @@ type Criterion struct {
 	RightOp []string
 	// Type is the type of the query
 	Type CriterionType
+	// TemplateParametersMap contains the template parameters of Criterion supporting sub-queries
+	TemplateParametersMap map[string]interface{}
```
eyalnathan, line 115: I would create a type for that, e.g. `type TemplateParameters map[string]interface{}`, even if not I suggest removing the "Map" from the name (and the interface function).

pkg/query/selection.go
```diff
@@ -109,13 +111,23 @@ type Criterion struct {
...
 }
 
 // ByField constructs a new criterion for field querying
 func ByField(operator Operator, leftOp string, rightOp ...string) Criterion {
 	return NewCriterion(leftOp, operator, rightOp, FieldQuery)
 }
 
+func ByIdNotExist(subQuery string) Criterion {
```
eyalnathan, line 123: could be nice to get the template params here already, and not have to set it afterwards

pkg/query/selection.go
```diff
@@ -175,13 +187,26 @@ func (c Criterion) Validate() error {
...
 	for _, op := range c.RightOp {
-		if strings.ContainsRune(op, '\n') {
+		if strings.ContainsRune(op, '\n') && c.Type != ExistQuery {
 			return &util.UnsupportedQueryError{Message: fmt.Sprintf("%s with key \"%s\" has value \"%s\" contaning forbidden new line character", c.Type, c.LeftOp, op)}
 		}
 	}
 	return nil
 }
 
+func (c *Criterion) GetTemplateParams() map[string]interface{} {
```
eyalnathan, line 197: do you need a getter? it is public anyways

pkg/query/selection.go
```diff
@@ -175,13 +187,26 @@ func (c Criterion) Validate() error {
...
+	return c.TemplateParametersMap
+}
+
+func (c *Criterion) AddTemplateParams(templateParametersMap map[string]interface{}) {
```
eyalnathan, line 201: same here, is this needed?

storage/postgres/storage.go
```diff
@@ -465,6 +465,10 @@ func (ps *Storage) UpdateLabels(ctx context.Context, objectType types.ObjectType
 	return ps.updateLabels(ctx, objectType, objectID, labelChanges)
 }
 
+func (ps *Storage) GetEntitiesByTableNameMap() map[string]string {
```
eyalnathan, line 468: this is a bit low level, both in terminology and implementation - you are returning an internal object of the underlying scheme, what happens if someone makes modifications to it? I think we should have something like "GetEntities" that will return []Entity (new type that will have the required fields - entity name, table name etc.). the impl. can use this map as input to create the entity list.

pkg/query/operators.go
eyalnathan: now i see they are used also from other packages, so it won't be that easy..

DanFrei: Yes I totally agree.
Initially I wanted to make them private but that would've complicate the code so I think it would be better this way.

pkg/query/selection.go
DanFrei: I've removed "Map" from the name.

pkg/query/selection.go
DanFrei: Good point! thanks!
It's done.

pkg/query/selection.go
DanFrei: True! getter removed.

pkg/query/selection.go
DanFrei: I've changed it to be Private (as now it's being used as part of ByIdNotExistWithTemplateParameters and ByIdExistWithTemplateParameters)

storage/postgres/storage.go
DanFrei: You're right! good point.
I've fixed it according to your suggestion.
Thanks!!!

pkg/query/operators.go
DanFrei: After discussing over this issue, it has been decided to rename "NotExistsOperator" and "ExistsOperator" to NotExistsSubquery and ExistsSubquery in order to distinguish the aforementioned from the rest of the operators as these aren't "regular" operators.

storage/interfaces.go
```diff
@@ -46,6 +46,11 @@ type Label interface {
 	GetValue() string
 }
 
+type ResourceEntity struct {
```
eyalnathan, line 49: rename to EntityMetadata?

pkg/query/selection.go
```diff
@@ -175,13 +201,22 @@ func (c Criterion) Validate() error {
...
 	for _, op := range c.RightOp {
-		if strings.ContainsRune(op, '\n') {
+		if strings.ContainsRune(op, '\n') && c.Type != ExistQuery {
 			return &util.UnsupportedQueryError{Message: fmt.Sprintf("%s with key \"%s\" has value \"%s\" contaning forbidden new line character", c.Type, c.LeftOp, op)}
 		}
 	}
 	return nil
 }
 
+func (c *Criterion) addTemplateParams(templateParameters TemplateParameters) {
```
eyalnathan, line 211: does anyone need this "add" functionality? why not just use assignment of the parameters field?

pkg/query/selection.go
```diff
@@ -109,13 +113,35 @@ type Criterion struct {
...
+	criterion := ByIdNotExist(subQuery)
+	criterion.addTemplateParams(templateParameters)
+	return criterion
+}
+
+func ByIdNotExist(subQuery string) Criterion {
+	return NewCriterion("", NotExistsSubquery, []string{subQuery}, ExistQuery)
+}
+
+func ByIdExistWithTemplateParameters(subQuery string, templateParameters TemplateParameters) Criterion {
```
eyalnathan, line 135: usually the function with less parameters call the one with more parameters using nil for the missing params, and not the other way around, so you only have one function with actual logic.

DanFrei: Ok, got it.

pkg/query/selection.go
DanFrei: Ok, I've removed the "add" functionality and simply set with the given parameters field.

operations/maintainer.go
```diff
@@ -252,6 +260,28 @@ func (om *Maintainer) cleanupInternalFailedOperations() {
...
+		// check if operation hasn't been updated for the operation's maximum allowed time to live in DB
+		query.ByField(query.LessThanOperator, "updated_at", util.ToRFCNanoFormat(currentTime.Add(-om.settings.Lifespan))),
+	}
+	// Build the 'WHERE' clause of the deletion statement with 'byIdNotExist' criterion for each resource table.
+	for _, entity := range om.repository.GetEntities() {
+		if entity.TableName == "operations" {
+			continue
+		}
+		templateParameters := query.TemplateParameters{"RESOURCE_TABLE": entity.TableName}
+		byIdNotExistCriterion := query.ByIDNotExistWithTemplateParameters(storage.GetSubQuery(storage.QueryForNonResourcelessOperations), templateParameters)
```
eyalnathan, line 275: another byId :)

storage/postgres/where_clause_tree.go
```diff
@@ -100,7 +102,10 @@ func criterionSQL(c query.Criterion, dbTags []tagType, tableAlias string) (strin
 	ttype := findTagType(dbTags, c.LeftOp)
 	dbCast := determineCastByType(ttype)
 	var clause string
-	if tableAlias != "" {
+	if c.Type == query.ExistQuery {
+		rightResolvedValue, _ := tsprintf(rightOpQueryValue.(string), c.TemplateParameters)
```
eyalnathan, line 106: the error should probably not be ignored

storage/sub_query.go
```diff
@@ -0,0 +1,28 @@
+package storage
+
+type SubQuery int
+
+const (
+	QueryForAllLastOperationsPerResource SubQuery = iota
+	QueryForNonResourcelessOperations
+)
+
+var subQueries = map[SubQuery]string{
```
eyalnathan, line 10: maybe some documentation (comments) on how to work with sub queries? how to write the SQL, how it should be used etc.

test/operations_test/operations_test.go
```diff
@@ -257,12 +257,78 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+						testResource2 := &types.Platform{
+							Base: types.Base{ID: "test-resource2"},
+						}
+						lastInternalSuccessfulOperation := &types.Operation{
+							Base:       types.Base{ID: "test-last-op-internal-successful"},
+							ResourceID: testResource1.ID,
+							Type:       types.CREATE,
+							State:      "succeeded",
+						}
+						lastInternalFailedOperation := &types.Operation{
```
eyalnathan, line 298: why failed? and if failed, why is the status "succeeded"? and what does "internal" mean here? does it matter?

test/operations_test/operations_test.go
```diff
@@ -257,12 +257,78 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+						ctx.SMRepository.Create(context.Background(), testResource1)
+						ctx.SMRepository.Create(context.Background(), testResource2)
+						ctx.SMRepository.Create(context.Background(), lastInternalSuccessfulOperation)
+						ctx.SMRepository.Create(context.Background(), lastInternalFailedOperation)
+					})
+					It("Should not be deleted", func() {
+						ensureOperationExist(ctx.SMRepository, "test-last-op-internal-successful")
+						ensureOperationExist(ctx.SMRepository, "test-last-op-internal-failed")
+
+						Eventually(func() bool {
```
eyalnathan, line 313: Eventually is good when the expected state is not the current state. Calling it like this doesn't really test anything - you ensure that this is the state before the "Eventually", so it is immediately fulfilled. If the last ops will be deleted the test will miss it.

test/query_test/query_test.go
```diff
@@ -51,167 +49,308 @@ var _ = Describe("Service Manager Query", func() {
...
+
+					_, err := repository.Create(context.Background(), resource)
+					_, err = repository.Create(context.Background(), opForInstance1)
+					Expect(err).ToNot(HaveOccurred())
+					repository.Create(context.Background(), resourcelessOperation)
+				})
+
+				It("should retrieve only the operation that is not associated to a resource", func() {
+					templateParameters := query.TemplateParameters{"RESOURCE_TABLE": "platforms"}
+					criterion := query.ByIDNotExistWithTemplateParameters(storage.GetSubQuery(storage.QueryForNonResourcelessOperations), templateParameters)
```
eyalnathan, line 281: aren't we in a context called "By ID Exist with template params"? this is testing ByIDNotExist

operations/maintainer.go
DanFrei: Fixed!

storage/postgres/where_clause_tree.go
DanFrei: Since now the tsprintf is done outside the Criterion, this err is no longer relevant here :)
Thanks for this, now the solution is more simple 👍

storage/sub_query.go
DanFrei: Done!

pkg/query/selection.go
```diff
@@ -36,6 +36,8 @@ const (
 	Separator string = "and"
 )
 
+type TemplateParameters map[string]interface{}
```
eyalnathan, line 39: remove this, use map in the one place it is used

storage/sub_query.go
```diff
@@ -0,0 +1,45 @@
...
+	QueryForAllLastOperationsPerResource SubQuery = iota
+	QueryForNonResourcelessOperations
+)
+
+// The aforementioned sub-queries are dedicated to be used with ByIDExists/ByIDNotExists Criterion to allow additional querying/filtering on top of
+// combined with main query generated from the criteria.
+// As in sql's EXIST/ NOT EXIST,	in order for main query to work with the sub-query,the  sub-queries will require a where clause that compares the id from
+// the parent query with and id retrieved from the sub-query.
+//
+// Example: Get all internal operations which aren't orphans (have corresponding resources) using ByIdExist criterion:
```
eyalnathan, line 15: simplify the example - no need for additional criterias. in addition, make sure the example works, and add the way to use it (e.g. repo.Count/Delete/.. with the relevant object type)

pkg/query/selection.go
DanFrei: Done

storage/sub_query.go
```diff
@@ -0,0 +1,45 @@
...
+// The aforementioned sub-queries are dedicated to be used with ByIDExists/ByIDNotExists Criterion to allow additional querying/filtering
+// combined with a main query generated from the criteria.
+// As in sql's EXIST/ NOT EXIST, in order for the main query to work with the sub-query, the sub-queries will require a where clause that
+// compares the id from the parent query with and id retrieved from the sub-query.
+//
+// Example: Get all internal operations which aren't orphans (have corresponding resources) using ByIDExist criterion:
+//
+// queryForAllNonOrphanOperations := `
+//    SELECT id
+//    FROM service_instances.id
```
eyalnathan, line 19: FROM service_instances , remove `.id`

DanFrei: done :)

PR #530, opened 2020-07-07
Save instance in context

kerenlahav:
## Motivation

1. Save instance that was fetched from DB to context to avoid querying the DB again to get the same instance
2. Manage instance deprovision and polling by an external plugin

## Approach

1. The instance will be saved and available in the context after fetched from DB
2. An external plugin want to manage the deprovision process, it saves the delete operation in the context, which signals the osb_store_plugin that someone else is managing this operation and the response should be returned to the client without further actions

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [ ] Unit tests
* [x] Integration tests


api/osb/osb_store_plugin.go
```diff
@@ -433,6 +433,12 @@ func (sp *storePlugin) Deprovision(request *web.Request, next web.Handler) (*web
 	if err != nil {
 		return nil, err
 	}
```
pavelmaliy, line 435: test for it?

api/osb/check_platform_id_plugin.go
```diff
@@ -90,7 +91,17 @@ func (p *checkPlatformIDPlugin) assertPlatformID(req *web.Request, next web.Hand
 		}
 		return nil, util.HandleStorageError(err, string(types.ServiceInstanceType))
 	}
-	instance := object.(*types.ServiceInstance)
+
+	instance, ok := object.(*types.ServiceInstance)
```
eyalnathan, line 95: can this happen? you just loaded the instance from DB so it has to be a service instance, and if the instance was not found in DB the flow does not get here

api/osb/osb_store_plugin.go
```diff
@@ -433,6 +433,12 @@ func (sp *storePlugin) Deprovision(request *web.Request, next web.Handler) (*web
...
+	ctx := request.Context()
+	_, operationFound := opcontext.Get(ctx)
+	if operationFound {
+		log.C(ctx).Debug("operation found in context, continue..")
```
eyalnathan, line 439: not clear why continue, can you enhance the message (as the one below about polling)?

instances/instcontext/context.go
```diff
@@ -0,0 +1,22 @@
+package instcontext
```
eyalnathan, line 1: i like the other way of managing context keys better, have a look at https://github.com/Peripli/service-manager/blob/master/pkg/web/context_keys.go

test/osb_and_plugin_test/osb_test/deprovision_test.go
```diff
@@ -388,4 +388,23 @@ var _ = Describe("Deprovision", func() {
...
+		BeforeEach(func() {
+			brokerServer.ServiceInstanceHandler = parameterizedHandler(http.StatusOK, `{}`)
+			ctx.SMWithBasic.PUT(smBrokerURL+"/v2/service_instances/"+SID).
+				WithHeader(brokerAPIVersionHeaderKey, brokerAPIVersionHeaderValue).
+				WithJSON(provisionRequestBodyMap()()).Expect().StatusRange(httpexpect.Status2xx)
+		})
+
+		It("osb store plugin should not create delete operation", func() {
+			shouldSaveOperationInContext = true
+			brokerServer.ServiceInstanceHandler = parameterizedHandler(http.StatusAccepted, `{"operation":"abc123"}`)
```
eyalnathan, line 402: this should be in BeforeEach, the It part should only do the test, not the preparations

api/osb/osb_store_plugin.go
```diff
@@ -610,6 +615,12 @@ func (sp *storePlugin) PollInstance(request *web.Request, next web.Handler) (*we
 	if err != nil {
 		return nil, err
 	}
+	ctx := request.Context()
```
eyalnathan, line 618: is this tested? I only see a test for deprovisioning flow

PR #532, opened 2020-07-12
Lifecycle changes

alexwo:
## Motivation

## Approach

Create changes:
- Preserve ASYNC instances or bindings that fail during the creation
- Don't keep instances or bindings that require sync orphaned mitigation ( use the operation context for orphaned mitigation instead)

Delete changes:
- Delete service instances or bindings once orphaned mitigation is completed successfully
- Keep instances or bindings in case orphaned mitigation fails

Operation changes:
- Don't return the operation context on rest calls
- +1 JSON context column is added

## Pull Request status

* [x] Initial implementation
* [X ] Integration tests

### DIFF
--- a/storage/interceptors/smaap_service_binding_interceptor.go
+++ b/storage/interceptors/smaap_service_binding_interceptor.go
@@ -156,6 +156,7 @@ func (i *ServiceBindingInterceptor) AroundTxCreate(f storage.InterceptCreateArou
 				return nil, fmt.Errorf("failed to marshal OSB context %+v: %s", bindRequest.Context, err)
 			}
 			binding.Context = contextBytes
+			operation.Context.ServiceInstanceID = binding.ServiceInstanceID
 
 			log.C(ctx).Infof("Sending bind request %s to broker with name %s", logBindRequest(bindRequest), broker.Name)
 			bindResponse, err = osbClient.Bind(bindRequest)
@@ -170,7 +171,6 @@ func (i *ServiceBindingInterceptor) AroundTxCreate(f storage.InterceptCreateArou
 					operation.DeletionScheduled = time.Now()
 					operation.Reschedule = false
 					operation.RescheduleTimestamp = time.Time{}
-					operation.Context.ServiceInstanceID = binding.ServiceInstanceID
 					if _, err := i.repository.Update(ctx, operation, types.LabelChanges{}); err != nil {
 						return nil, fmt.Errorf("failed to update operation with id %s to schedule orphan mitigation after broker error %s: %s", operation.ID, brokerError, err)
 					}
@@ -205,7 +205,6 @@ func (i *ServiceBindingInterceptor) AroundTxCreate(f storage.InterceptCreateArou
 				if bindResponse.OperationKey != nil {
 					operation.ExternalID = string(*bindResponse.OperationKey)
 				}
-				operation.Context.ServiceInstanceID = binding.ServiceInstanceID
 				if _, err := i.repository.Update(ctx, operation, types.LabelChanges{}); err != nil {
 					return nil, fmt.Errorf("failed to update operation with id %s to mark that next execution should be a reschedule: %s", instance.ID, err)
 				}
--- a/storage/interceptors/smaap_service_binding_interceptor.go
+++ b/storage/interceptors/smaap_service_binding_interceptor.go
@@ -250,22 +250,27 @@ func (i *ServiceBindingInterceptor) AroundTxDelete(f storage.InterceptDeleteArou
 
 		if bindings.Len() != 0 {
 			binding := bindings.ItemAt(0).(*types.ServiceBinding)
-
 			operation, found := opcontext.Get(ctx)
 			if !found {
 				return fmt.Errorf("operation missing from context")
 			}
 
 			deleteBindingError := i.deleteSingleBinding(ctx, binding, operation)
+
 			if !shouldKeepResource(operation, deleteBindingError) {
+
 				if operation.Context == nil {
+					// if triggered by a service broker deletion, the context object can be empty.
 					operation.Context = &types.OperationContext{}
-					operation.Context.ServiceInstanceID = binding.ServiceInstanceID
-					err := enrichOperationContext(ctx, operation, i.repository)
-					if err != nil {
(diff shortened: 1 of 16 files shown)


storage/interceptors/smaap_service_binding_interceptor.go
```diff
@@ -238,27 +241,54 @@ func (i *ServiceBindingInterceptor) AroundTxDelete(f storage.InterceptDeleteArou
...
 		return nil
 	}
 }
 
+func shouldRemoveResource(operation *types.Operation) bool {
+	if operation.State == types.FAILED {
+		if operation.Type == types.CREATE && operation.Context.Async {
+			return false
+		}
+
```
pavelmaliy, line 284: you already asked operation.State == types.FAILED in the outer if

storage/interceptors/smaap_service_binding_interceptor.go
```diff
@@ -238,27 +241,54 @@ func (i *ServiceBindingInterceptor) AroundTxDelete(f storage.InterceptDeleteArou
...
+		if !found {
+			return fmt.Errorf("operation missing from context")
+		}
+
 		if bindings.Len() != 0 {
 			binding := bindings.ItemAt(0).(*types.ServiceBinding)
-
-			operation, found := opcontext.Get(ctx)
-			if !found {
-				return fmt.Errorf("operation missing from context")
```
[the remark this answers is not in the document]
alexwo: we always should have the operation in the context in this flow. if it's not there something wrong probably occurred.
do you see a case for proceeding to delete single binding without it?

storage/interceptors/smaap_service_binding_interceptor.go
```diff
@@ -166,20 +166,22 @@ func (i *ServiceBindingInterceptor) AroundTxCreate(f storage.InterceptCreateArou
 					StatusCode:  http.StatusBadGateway,
 				}
 				if shouldStartOrphanMitigation(err) {
-					// store the instance so that later on we can do orphan mitigation
-					_, err := f(ctx, obj)
-					if err != nil {
-						return nil, fmt.Errorf("broker error %s caused orphan mitigation which required storing the resource which failed with: %s", brokerError, err)
-					}
-
-					// mark the operation as deletion scheduled meaning orphan mitigation is required
```
pavelmaliy, line 175: store service instanceID always you are trying to fetch the binding from repository anyway

storage/interceptors/smaap_service_binding_interceptor.go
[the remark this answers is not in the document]
pavelmaliy: this if is unneeded because !operation.context.Async will never happen here

storage/interceptors/smaap_service_binding_interceptor.go
alexwo: hmm it seem to be out dated?
https://github.com/Peripli/service-manager/blob/b1f4fc5b925343a3fcf3bd60e0f8d938130eb1a8/storage/interceptors/smaap_service_binding_interceptor.go#L179

storage/interceptors/smaap_service_binding_interceptor.go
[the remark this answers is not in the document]
alexwo: not sure do you mean this : "if !found {return fmt.Errorf("operation missing from context")"  that's the one that's highlighted? or something else..?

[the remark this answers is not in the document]
pavelmaliy: if operation.Type == types.CREATE && !operation.Context.Async {
can't happen

storage/interceptors/smaap_service_binding_interceptor.go
```diff
@@ -238,27 +241,54 @@ func (i *ServiceBindingInterceptor) AroundTxDelete(f storage.InterceptDeleteArou
...
-				return fmt.Errorf("operation missing from context")
+			if operation.Type == types.CREATE && !operation.Context.Async {
+				if err := f(ctx, deletionCriteria...); err != nil {
+					return err
+				}
 			}
-
 			if err := i.deleteSingleBinding(ctx, binding, operation); err != nil {
 				return err
 			}
```
pavelmaliy, line 258: why don't just  prepare binding object or from DB or from context without this logic?

storage/interceptors/smaap_service_binding_interceptor.go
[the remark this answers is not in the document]
alexwo: actually it still can happen, if the instance requires polling and we don't know in advance if its going to pass or fail. we preserve the instance as done here:
https://github.com/Peripli/service-manager/blob/b1f4fc5b925343a3fcf3bd60e0f8d938130eb1a8/storage/interceptors/smaap_service_binding_interceptor.go#L217

and this results in :
operation.Type == types.CREATE && !operation.Context.Async

storage/interceptors/smaap_service_binding_interceptor.go
```diff
@@ -238,27 +241,54 @@ func (i *ServiceBindingInterceptor) AroundTxDelete(f storage.InterceptDeleteArou
...
+		} else if operation.InOrphanMitigationState() && operation.Context.ServiceInstanceID != "" {
+			// In case we don't have instance & use the operation context to perform orphan mitigation
+			binding := types.ServiceBinding{}
+			binding.ServiceInstanceID = operation.Context.ServiceInstanceID
+			binding.ID = operation.ResourceID
+			if err := i.deleteSingleBinding(ctx, &binding, operation); err != nil {
+				return err
```
pavelmaliy, line 265: you can't return here err
otherwise if operation failed it will return as error and you won't reach the rest of the code

storage/interceptors/smaap_service_binding_interceptor.go
```diff
@@ -238,27 +241,54 @@ func (i *ServiceBindingInterceptor) AroundTxDelete(f storage.InterceptDeleteArou
...
+			}
 		}
 
-		if err := f(ctx, deletionCriteria...); err != nil {
-			return err
```
pavelmaliy, line 255: let;s discuss further about this if not sure you need it

PR #539, opened 2020-07-22
Set db connection read and write timeout

nirbenrey:
# Pull Request Template

## Motivation

In case the database goes down, or the network dies, the open connections will hang.

## Approach

We set a timeout for reading and writing on the database connection.
Since this functionality is not supported by lib/pq we use https://github.com/Kount/pq-timeouts
library as suggested in https://github.com/lib/pq/issues/450

## Pull Request status

* [] Initial implementation
* [] validation
* [x] completed

## Third-party code

https://github.com/Kount/pq-timeouts

### DIFF
--- a/storage/postgres/storage.go
+++ b/storage/postgres/storage.go
@@ -39,7 +39,7 @@ import (
 )
 
 const (
-	postgresDriverName  = "postgres"
+	postgresDriverName  = "pq-timeouts"
 	foreignKeyViolation = "foreign_key_violation"
 )
 
@@ -72,11 +72,11 @@ func (ps *Storage) Open(settings *storage.Settings) error {
 	ps.mutex.Lock()
 	defer ps.mutex.Unlock()
 	if ps.db == nil {
-		sslModeParam := ""
+		url := fmt.Sprintf("%s?read_timeout=%d&write_timeout=%d", settings.URI, settings.ReadTimeout, settings.WriteTimeout)
 		if settings.SkipSSLValidation {
-			sslModeParam = "?sslmode=disable"
+			url += "&sslmode=disable"
 		}
-		db, err := ps.ConnectFunc(postgresDriverName, settings.URI+sslModeParam)
+		db, err := ps.ConnectFunc(postgresDriverName, url)
 		if err != nil {
 			return fmt.Errorf("could not connect to PostgreSQL: %s", err)
 		}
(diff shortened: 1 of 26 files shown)


storage/postgres/storage.go
```diff
@@ -72,11 +74,24 @@ func (ps *Storage) Open(settings *storage.Settings) error {
...
+		if err != nil {
+			return fmt.Errorf("could not parse PostgreSQL URI: %s", err)
+		}
+
+		parsedQuery, err := url.ParseQuery(parsedUrl.RawQuery)
+		if err != nil {
+			return fmt.Errorf("could not parse PostgreSQL URL query: %s", err)
+		}
+
+		parsedQuery.Add("read_timeout", strconv.Itoa(settings.ReadTimeout))
```
eyalnathan, line 87: what happens if the URL already contain those? i am not sure what would the URL look like in e.g. private cloud, in IBM for example it had the `sslmode=required` already part of the URL in the binding credentials.

nirbenrey: It will be added twice to the URL for example "postgres://postgres:postgres@localhost:5432/postgres?read_timeout=100&read_timeout=900000&sslmode=disable&write_timeout=900000".

In origin/master implementation for sslmode, you can configure the storage.uri with sslmode=required query param and  configure also storage.skip_ssl_validation=true resulting in "broken" URL "postgres://postgres:postgres@localhost:5432/postgres?sslmode=required&sslmode=disable".

So, giving priority to storage.uri query params read_timeout and write_timeout over the specific ones won't break anything simply because no one is using it yet. But giving priority to the sslmode query param can cause a regression.

Any thoughts?

nirbenrey: Resolved by using parsedQuery.Set(...) when setting read_timeout and write_out

storage/postgres/storage.go
```diff
@@ -72,11 +74,24 @@ func (ps *Storage) Open(settings *storage.Settings) error {
...
+			return fmt.Errorf("could not parse PostgreSQL URI: %s", err)
+		}
+
+		parsedQuery, err := url.ParseQuery(parsedUrl.RawQuery)
+		if err != nil {
+			return fmt.Errorf("could not parse PostgreSQL URL query: %s", err)
+		}
+
+		parsedQuery.Set("read_timeout", strconv.Itoa(settings.ReadTimeout))
+		parsedQuery.Set("write_timeout", strconv.Itoa(settings.WriteTimeout))
```
eyalnathan, line 88: maybe a small test to make sure that this behaves as you expect?

PR #540, opened 2020-07-26
Lifecycle addon v2

alexwo:
## Motivation

For instance and binding management APIs, we would like to return the actual broker response, in cases where the client does not specify the async mode.

## Approach

In case a client does not specify an expected response mode via the async=true/false query param.
Service Manager will attempt to fulfill a synchronize request to provision or de-provision a resource.

1. If the service broker response requires polling the response to the client will be async.
2. If the service broker response does not require polling, the client will get the broker response instantly 
3. In case of any errors returned by the service broker before polling has started, those will be instantly sent back to the client

Support is added to POST, PUT methods

## Pull Request status

* [x] Initial implementation
* [x] Integration tests

### DIFF
--- a/api/base_controller.go
+++ b/api/base_controller.go
@@ -196,23 +196,29 @@ func (c *BaseController) CreateObject(r *web.Request) (*web.Response, error) {
 		Context:       c.prepareOperationContextByRequest(r),
 	}
 
-	if operation.Context.Async {
-		log.C(ctx).Debugf("Request will be executed asynchronously due to client request async=true")
+	var createdObj types.Object
+	if operation.Context.IsAsyncDefinedByClient {
+		log.C(ctx).Debugf("Request will be executed by broker response")
+		createdObj, err = c.scheduler.ScheduleStorageAction(ctx, operation, action)
+		if err != nil {
+			return nil, err
+		}
+		if createdObj == nil {
+			return util.NewLocationResponse(operation.GetID(), operation.ResourceID, c.resourceBaseURL)
+		}
+	} else if operation.Context.Async {
+		log.C(ctx).Debugf("Request will be executed asynchronously")
 		return c.executeAsync(ctx, operation, action, result.GetID())
-	}
-
-	createdObj, err := c.scheduler.ScheduleSyncStorageAction(ctx, operation, action)
-	if err != nil {
-		return nil, util.HandleStorageError(err, c.objectType.String())
+	} else {
+		log.C(ctx).Debugf("Request will be executed synchronously")
+		createdObj, err = c.scheduler.ScheduleSyncStorageAction(ctx, operation, action)
+		if err != nil {
+			return nil, util.HandleStorageError(err, c.objectType.String())
+		}
 	}
 
 	if err := attachLastOperation(ctx, createdObj.GetID(), createdObj, c.repository); err != nil {
 		return nil, err
-	}
-
-	if createdObj.GetLastOperation().IsAsyncResponse() {
-		log.C(ctx).Debugf("Request will be executed asynchronously due to broker async response")
-		return c.executeAsync(ctx, operation, action, result.GetID())
 	}
 
 	cleanObject(createdObj.GetLastOperation())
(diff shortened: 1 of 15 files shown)


api/base_controller.go
```diff
@@ -192,23 +193,17 @@ func (c *BaseController) CreateObject(r *web.Request) (*web.Response, error) {
 		ResourceType:  c.objectType,
 		PlatformID:    types.SMPlatform,
 		CorrelationID: log.CorrelationIDFromContext(ctx),
-		Context:       &types.OperationContext{Async: c.shouldExecuteAsync(r)},
+		Context:       c.getOperationContextByRequest(r),
 	}
 
-	if c.shouldExecuteAsync(r) {
-		log.C(ctx).Debugf("Request will be executed asynchronously")
+	if operation.IsAsyncResponse() {
```
drorhayun, line 199: in this case you still don't have the response from the broker
it means that e.Context.BrokerResponse.Async will be always false
so you can just check the operation.Context.Async flag

api/base_controller.go
```diff
@@ -218,9 +213,29 @@ func (c *BaseController) CreateObject(r *web.Request) (*web.Response, error) {
 		return nil, err
 	}
 
+	if createdObj.GetLastOperation().IsAsyncResponse() {
+		if err := c.checkAsyncSupport(); err != nil {
```
drorhayun, line 217: same here, you check it in c.executeAsync function

api/base_controller.go
```diff
@@ -282,27 +297,34 @@ func (c *BaseController) DeleteSingleObject(r *web.Request) (*web.Response, erro
...
 		CorrelationID: log.CorrelationIDFromContext(ctx),
-		Context:       &types.OperationContext{Async: c.shouldExecuteAsync(r)},
+		Context:       c.getOperationContextByRequest(r),
 	}
 
-	if c.shouldExecuteAsync(r) {
-		log.C(ctx).Debugf("Request will be executed asynchronously")
+	if operation.IsAsyncResponse() {
+		log.C(ctx).Debugf("Request will be executed asynchronously due to client request async=true")
 		if err := c.checkAsyncSupport(); err != nil {
```
drorhayun, line 305: same here, you check it in c.executeAsync function

api/base_controller.go
```diff
@@ -282,27 +297,34 @@ func (c *BaseController) DeleteSingleObject(r *web.Request) (*web.Response, erro
...
 		return nil, util.HandleStorageError(err, c.objectType.String())
 	}
 
+	operation, err = getLastOperation(ctx, objectID, c.repository)
+	if err != nil {
+		return nil, err
+	}
+
+	if operation.IsAsyncResponse() {
+		if err := c.checkAsyncSupport(); err != nil {
```
drorhayun, line 321: same here, you check it in c.executeAsync function

api/base_controller.go
```diff
@@ -508,10 +526,20 @@ func (c *BaseController) PatchObject(r *web.Request) (*web.Response, error) {
 		return nil, util.HandleStorageError(err, c.objectType.String())
 	}
 
-	cleanObject(object)
 	if err := attachLastOperation(ctx, object.GetID(), object, c.repository); err != nil {
 		return nil, err
 	}
+
+	if object.GetLastOperation().IsAsyncResponse() {
+		if err := c.checkAsyncSupport(); err != nil {
```
drorhayun, line 534: same here, you check it in c.executeAsync function

api/base_controller.go
```diff
@@ -282,27 +297,34 @@ func (c *BaseController) DeleteSingleObject(r *web.Request) (*web.Response, erro
...
-		return util.NewLocationResponse(operation.GetID(), objectID, c.resourceBaseURL)
+		return c.executeAsync(ctx, operation, action, objectID)
 	}
 
-	log.C(ctx).Debugf("Request will be executed synchronously")
 	if _, err := c.scheduler.ScheduleSyncStorageAction(ctx, operation, action); err != nil {
 		return nil, util.HandleStorageError(err, c.objectType.String())
 	}
 
+	operation, err = getLastOperation(ctx, objectID, c.repository)
```
drorhayun, line 315: why you need to fetch the last operation?
is not the operation that you passed to ScheduleSyncStorageAction function?

api/base_controller.go
alexwo: This is needed 2 reflect a case where the client did not send the **async=true/false** 
Broker's actual response is checked later on and is stored in the BrokerResponse.async context-param.

api/base_controller.go
alexwo: yes its relevant for all areas, it was there before but actually seems to be redundant. we will fix it everywhere.

pkg/types/operartion_utils.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+func (e *Operation) IsAsyncResponse() bool {
+
+	if e.Context.IsAsyncDefinedByClient {
+		return e.Context.BrokerResponse.Async
```
drorhayun, line 6: you don't need this property anymore (BrokerResponse.Async)
now you have indication if the user pass async or not
you can update the async to true (in the interceptor) in case of async response and !Context.IsAsyncDefinedByClient

and you can just return e.Context.Async

api/base_controller.go
```diff
@@ -652,13 +662,22 @@ func (c *BaseController) parsePageToken(ctx context.Context, token string) (stri
...
 
-func (c *BaseController) shouldExecuteAsync(r *web.Request) bool {
+func (c *BaseController) prepareOperationContextByRequest(r *web.Request) *types.OperationContext {
+	operationContext := &types.OperationContext{BrokerResponse: types.BrokerResponse{}}
 	async := r.URL.Query().Get(web.QueryParamAsync)
+
 	if async == "" {
-		return c.isAsyncDefault
+		operationContext.Async = false
+		operationContext.IsAsyncDefinedByClient = true
```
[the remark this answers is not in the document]
alexwo: IsAsyncNotDefinedByClient  👍

pkg/types/operation.go
```diff
@@ -89,9 +89,15 @@ type Operation struct {
 }
 
 type OperationContext struct {
-	Async             bool   `json:"async"`
-	ServicePlanID     string `json:"service_plan_id"`
-	ServiceInstanceID string `json:"service_instance_id"`
+	Async                  bool   `json:"async"`
+	IsAsyncDefinedByClient bool   `json:"by_broker_response,omitempty"`
```
drorhayun, line 93: change the json property to is_async...

alexwo: it will require a migration we already have these changes persisted..

PR #543, opened 2020-07-30
update broker credentials

kerenlahav:
## Motivation

When a broker is being updated and the update is not successful in the platform itself, for example in CF, the credentials for the platform-broker are already updated in Service Manager.

If multiple updates of the same type happen, then the broker will become unusable because the platform will no longer have the credentials to call Service Manager for this broker.

## Approach

keep old credentials until the new ones are verified

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [ ] Unit tests
* [ ] Integration tests


storage/postgres/migrations/20200722413800_alter_broker_platform_credentials_active.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
```
eyalnathan, line 1: usually these come with a corresponding `.down.sql` script rollback the migration if needed, can you add it?

kerenlahav: done

PR #552, opened 2020-09-13
allowing query separators to be a substring of a label key

evyaffe:
# Pull Request Template

## Motivation

When querying the DB using labelQuery for landscape_label the request fail because 'and' is considered a separator and is contained in landscape as a substring

## Approach

The change is in the validation - we check for 'and' as a standalone word

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [ ] Unit tests
* [ ] Integration tests


pkg/query/selection_test.go
```diff
@@ -332,8 +332,10 @@ new line`))
 		Entry("Non-numeric operand is not allowed by comparison operators",
 			ByField(LessThanOperator, "left", "not numeric"),
 			"not numeric or datetime"),
-		Entry("Separator word 'and' is not allowed to appear in left operand",
-			ByField(EqualsOperator, "band", "music"),
+		Entry("Separator word 'and' is allowed to appear in left operand as a substring",
+			ByField(EqualsOperator, "band", "music")),
+		Entry("Separator word 'and' is not allowed to appear in left operand as a standalone word",
+			ByField(EqualsOperator, "and", "this"),
```
eyalnathan, line 338: maybe a few more like this? "and b", "a and b", "a and"

test/broker_test/broker_test.go
```diff
@@ -687,10 +687,19 @@ var _ = test.DescribeTestsFor(test.TestCase{
 								WithJSON(postBrokerRequestWithLabels).
 								Expect().Status(http.StatusCreated).JSON().Object().Keys().Contains("id", "labels")
 						})
+
+						When("When creating labeled broker with a key containing a forbidden separator as substring", func() {
```
eyalnathan, line 691: Maybe "containing a separator not as a standalone word"? Feels strange saying "if you do something forbidden it should work well"

eyalnathan (approved):

PR #556, opened 2020-09-17
keep sub-account scoped platforms visibilities 

alexwo:
#  Motivation

The broker public plans interceptor (resyncPlanVisibilitiesWithSupportedPlatforms) does not distinguish between public plan visibilities & tenant scoped platform visibilities.
This leads to incorrect visibilities removal if a non-public plan visibility exists for a sub-account platform and the broker is updated.

## Approach
- Identify if a plan is visible on a sub-account scoped platform and keep the visibility in this case 
- Return platforms map instead of just the platform ids strings, in order to check if the platform is tenant scoped or not

## Pull Request status

* [x] Refactoring
* [x] Unit tests

### DIFF
--- a/storage/interceptors/broker_notifications_interceptor.go
+++ b/storage/interceptors/broker_notifications_interceptor.go
@@ -30,12 +30,12 @@ func NewBrokerNotificationsInterceptor() *NotificationsInterceptor {
 				}
 			}
 
-			supportedPlatforms, err := service_plans.ResolveSupportedPlatformsForPlans(ctx, plans, repository)
+			supportedPlatforms, err := service_plans.ResolveSupportedPlatformIDsForPlans(ctx, plans, repository)
 			if err != nil {
 				return nil, err
 			}
 
-			return removeSMPlatform(getAsPlatformIDs(supportedPlatforms)), nil
+			return removeSMPlatform(supportedPlatforms), nil
 		},
 		AdditionalDetailsFunc: func(ctx context.Context, objects types.ObjectList, repository storage.Repository) (objectDetails, error) {
 			details := make(objectDetails, objects.Len())
@@ -157,14 +157,4 @@ func fetchBrokerPlans(ctx context.Context, brokerID string, repository storage.R
 	}
 
 	return objList.(*types.ServicePlans).ServicePlans, nil
-}
-
-func getAsPlatformIDs(platforms map[string]*types.Platform) []string {
-	platformIDs := make([]string, 0)
-
-	for id := range platforms {
-		platformIDs = append(platformIDs, id)
-	}
-
-	return platformIDs
 }
--- a/storage/interceptors/broker_notifications_interceptor.go
+++ b/storage/interceptors/broker_notifications_interceptor.go
@@ -35,7 +35,7 @@ func NewBrokerNotificationsInterceptor() *NotificationsInterceptor {
 				return nil, err
 			}
 
-			return removeSMPlatform(supportedPlatforms), nil
+			return removeSMPlatform(getAsPlatformIds(supportedPlatforms)), nil
 		},
 		AdditionalDetailsFunc: func(ctx context.Context, objects types.ObjectList, repository storage.Repository) (objectDetails, error) {
 			details := make(objectDetails, objects.Len())
@@ -157,4 +157,14 @@ func fetchBrokerPlans(ctx context.Context, brokerID string, repository storage.R
 	}
 
 	return objList.(*types.ServicePlans).ServicePlans, nil
+}
+
+func getAsPlatformIds(platforms map[string]*types.Platform) []string {
+	platformIDs := make([]string, 0)
+
+	for id := range platforms {
+		platformIDs = append(platformIDs, id)
+	}
+
+	return platformIDs
 }
--- a/storage/interceptors/broker_notifications_interceptor.go
+++ b/storage/interceptors/broker_notifications_interceptor.go
@@ -30,12 +30,12 @@ func NewBrokerNotificationsInterceptor() *NotificationsInterceptor {
 				}
 			}
 
-			supportedPlatforms, err := service_plans.ResolveSupportedPlatformIDsForPlans(ctx, plans, repository)
+			supportedPlatforms, err := service_plans.ResolveSupportedPlatformsForPlans(ctx, plans, repository)
 			if err != nil {
 				return nil, err
 			}
 
-			return removeSMPlatform(getAsPlatformIds(supportedPlatforms)), nil
+			return removeSMPlatform(getAsPlatformIDs(supportedPlatforms)), nil
 		},
 		AdditionalDetailsFunc: func(ctx context.Context, objects types.ObjectList, repository storage.Repository) (objectDetails, error) {
 			details := make(objectDetails, objects.Len())
@@ -159,7 +159,7 @@ func fetchBrokerPlans(ctx context.Context, brokerID string, repository storage.R
 	return objList.(*types.ServicePlans).ServicePlans, nil
 }
 
-func getAsPlatformIds(platforms map[string]*types.Platform) []string {
+func getAsPlatformIDs(platforms map[string]*types.Platform) []string {
 	platformIDs := make([]string, 0)
 
 	for id := range platforms {
--- a/storage/interceptors/broker_public_plans_interceptor.go
+++ b/storage/interceptors/broker_public_plans_interceptor.go
@@ -37,15 +37,15 @@ type publicPlanProcessor func(broker *types.ServiceBroker, catalogService *types
 type supportedPlatformsProcessor func(ctx context.Context, plan *types.ServicePlan, repository storage.Repository) (map[string]*types.Platform, error)
 
 type PublicPlanCreateInterceptorProvider struct {
-	IsCatalogPlanPublicFunc     publicPlanProcessor
-	SupportedPlatformsProcessor supportedPlatformsProcessor
-	TenantKey                   string
+	IsCatalogPlanPublicFunc publicPlanProcessor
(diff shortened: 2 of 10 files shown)


i500866:
re-trigger tests

sigalmaya (commented):
isn't it better to define a type for supportedPlatfroms? like with IsCatalogPlanPublicFunc

storage/interceptors/broker_public_plans_interceptor.go
```diff
@@ -53,7 +56,8 @@ func (p *PublicPlanCreateInterceptorProvider) Name() string {
 
 type PublicPlanUpdateInterceptorProvider struct {
 	IsCatalogPlanPublicFunc publicPlanProcessor
-	SupportedPlatforms      func(ctx context.Context, plan *types.ServicePlan, repository storage.Repository) ([]string, error)
+	SupportedPlatforms      func(ctx context.Context, plan *types.ServicePlan, repository storage.Repository) (map[string]*types.Platform, error)
```
eyalnathan, line 59: don't you want to change the other occurrences to `SupportedPlatformsProcessor` as well?

storage/interceptors/broker_public_plans_interceptor.go
```diff
@@ -34,16 +34,19 @@ const (
 )
 
 type publicPlanProcessor func(broker *types.ServiceBroker, catalogService *types.ServiceOffering, catalogPlan *types.ServicePlan) (bool, error)
+type SupportedPlatformsProcessor func(ctx context.Context, plan *types.ServicePlan, repository storage.Repository) (map[string]*types.Platform, error)
```
eyalnathan, line 37: shouldn't this be private?

storage/service_plans/resolver.go
```diff
@@ -7,17 +7,13 @@ import (
 	"github.com/Peripli/service-manager/storage"
 )
 
-func ResolveSupportedPlatformIDsForPlans(ctx context.Context, plans []*types.ServicePlan, repository storage.Repository) ([]string, error) {
+func ResolveSupportedPlatformIDsForPlans(ctx context.Context, plans []*types.ServicePlan, repository storage.Repository) (map[string]*types.Platform, error) {
```
eyalnathan, line 10: isn't this exactly what the other function in the file is doing? you can just call the other one when you need it, and not change the other flows using this one

test/common/test_context.go
```diff
@@ -754,6 +754,19 @@ func (ctx *TestContext) RegisterPlatform() *types.Platform {
 	return ctx.RegisterPlatformWithType("test-type")
 }
 
+func (ctx *TestContext) RegisterTenantPlatform() *types.Platform {
```
eyalnathan, line 757: did you see `common.GenerateRandomPlatform()`? not sure we need this as well

test/public_plans_interceptor_test/public_plans_test.go
```diff
@@ -116,20 +118,43 @@ var _ = Describe("Service Manager Public Plans Interceptor", func() {
...
 				},
-				SupportedPlatforms: func(ctx context.Context, plan *types.ServicePlan, repository storage.Repository) ([]string, error) {
+				SupportedPlatforms: func(ctx context.Context, plan *types.ServicePlan, repository storage.Repository) (map[string]*types.Platform, error) {
 					return service_plans.ResolveSupportedPlatformIDsForPlans(ctx, []*types.ServicePlan{plan}, repository)
 				},
+				TenantKey: "tenant",
 			}).OnTxBefore(interceptors.BrokerCreateCatalogInterceptorName).Register()
-
+			_, err := smb.EnableMultitenancy("tenant", func(request *web.Request) (string, error) {
+				extractTenantFromToken := multitenancy.ExtractTenantFromTokenWrapperFunc("zid")
```
eyalnathan, line 127: looks like this is the 4th occurrence of this callback function in the tests, maybe it is time to extract and re-use?

storage/interceptors/broker_notifications_interceptor.go
```diff
@@ -158,3 +158,13 @@ func fetchBrokerPlans(ctx context.Context, brokerID string, repository storage.R
 
 	return objList.(*types.ServicePlans).ServicePlans, nil
 }
+
+func getAsPlatformIds(platforms map[string]*types.Platform) []string {
```
eyalnathan, line 162: if you are keeping this please change Ids -> IDs

test/common/test_context.go
alexwo: we can use RegisterPlatformInSM(GenerateRandomPlatform(), ctx.SMWithOAuthForTenant, map[string]string{}) directly, but this seems to be more readble this way?

eyalnathan: you can still use the other one as the impl. of this one, no need for the logic again.

storage/interceptors/broker_public_plans_interceptor.go
```diff
@@ -82,13 +88,14 @@ func (p *publicPlanCreateInterceptor) OnTxCreate(f storage.InterceptCreateOnTxFu
...
 		}
-		return newObject, resync(ctx, obj.(*types.ServiceBroker), txStorage, p.isCatalogPlanPublicFunc, p.supportedPlatforms)
+		return newObject, resync(ctx, obj.(*types.ServiceBroker), txStorage, p.isCatalogPlanPublicFunc, p.supportedPlatformsFunc, p.tenantKey)
 	}
 }
 
 type publicPlanUpdateInterceptor struct {
 	isCatalogPlanPublicFunc publicPlanProcessor
-	supportedPlatforms      func(ctx context.Context, plan *types.ServicePlan, repository storage.Repository) ([]string, error)
+	supportedPlatformsFunc  func(ctx context.Context, plan *types.ServicePlan, repository storage.Repository) (map[string]*types.Platform, error)
```
eyalnathan, line 97: use `supportedPlatformsProcessor`?

storage/interceptors/broker_public_plans_interceptor.go
```diff
@@ -101,11 +108,11 @@ func (p *publicPlanUpdateInterceptor) OnTxUpdate(f storage.InterceptUpdateOnTxFu
 		if err != nil {
 			return nil, err
 		}
-		return result, resync(ctx, result.(*types.ServiceBroker), txStorage, p.isCatalogPlanPublicFunc, p.supportedPlatforms)
+		return result, resync(ctx, result.(*types.ServiceBroker), txStorage, p.isCatalogPlanPublicFunc, p.supportedPlatformsFunc, p.tenantKey)
 	}
 }
 
-func resync(ctx context.Context, broker *types.ServiceBroker, txStorage storage.Repository, isCatalogPlanPublicFunc publicPlanProcessor, supportedPlatforms func(ctx context.Context, plan *types.ServicePlan, repository storage.Repository) ([]string, error)) error {
+func resync(ctx context.Context, broker *types.ServiceBroker, txStorage storage.Repository, isCatalogPlanPublicFunc publicPlanProcessor, supportedPlatforms func(ctx context.Context, plan *types.ServicePlan, repository storage.Repository) (map[string]*types.Platform, error), tenantKey string) error {
```
eyalnathan, line 115: use `supportedPlatformsProcessor`?

storage/interceptors/broker_public_plans_interceptor.go
```diff
@@ -214,13 +225,17 @@ func resyncPlanVisibilitiesWithSupportedPlatforms(ctx context.Context, txStorage
 }
 
 // platformsAnyMatchesVisibility checks whether any of the platform IDs matches the provided visibility
-func platformsAnyMatchesVisibility(platformIDs []string, visibility *types.Visibility) (int, bool) {
-	for i, platformID := range platformIDs {
-		if visibility.PlatformID == platformID {
-			return i, true
+func platformsAnyMatchesVisibility(supportedPlatforms map[string]*types.Platform, visibility *types.Visibility, tenantKey string) (string, bool, bool) {
```
eyalnathan, line 228: this function is not so clear, wouldn't it be more readable to have this one return the matching platform object, and then pass it to another function that checks if it is tenant scoped?

storage/interceptors/broker_public_plans_interceptor.go
```diff
@@ -176,19 +186,20 @@ func resyncPublicPlanVisibilities(ctx context.Context, txStorage storage.Reposit
...
-func resyncPlanVisibilitiesWithSupportedPlatforms(ctx context.Context, txStorage storage.Repository, planVisibilities types.ObjectList, isPlanPublic bool, planID string, broker *types.ServiceBroker, supportedPlatformIDs []string) error {
+func resyncPlanVisibilitiesWithSupportedPlatforms(ctx context.Context, txStorage storage.Repository, planVisibilities types.ObjectList, isPlanPublic bool, planID string, broker *types.ServiceBroker, supportedPlatforms map[string]*types.Platform, tenantKey string) error {
 	for i := 0; i < planVisibilities.Len(); i++ {
 		visibility := planVisibilities.ItemAt(i).(*types.Visibility)
 
 		shouldDeleteVisibility := true
 
-		idx, matches := platformsAnyMatchesVisibility(supportedPlatformIDs, visibility)
-		if isPlanPublic { // trying to match the current visibility to one of the supported platforms that should have visibilities
+		id, matches, isSubAccountScoped := platformsAnyMatchesVisibility(supportedPlatforms, visibility, tenantKey)
```
eyalnathan, line 195: isSubaccountScoped -> isTenantScoped

storage/interceptors/broker_public_plans_interceptor.go
alexwo: should not it be similiar to isCatalogPlanPublicFunc ..?

eyalnathan (dismissed):

eyalnathan (approved):

PR #557, opened 2020-09-29
Instance binding parameters

alexwo:
# Pull Request Template

## Motivation
Allow to get service instance or binding params incase instance or binding retrievable are supported by a service broker 

## Approach
Introduces 2 new APIs for getting service instance and binding parameters 
 
## Pull Request status

* [x] Initial implementation
* [x ] Integration tests


api/osb/utils.go
```diff
@@ -0,0 +1,60 @@
...
+			return nil, &util.HTTPError{
+				ErrorType:   "ServiceBrokerErr",
+				Description: fmt.Sprintf("error fetching $s with id %s from broker with name %s: timed out", resourceType, resourceId, broker.Name),
+				StatusCode:  http.StatusGatewayTimeout,
+			}
+		}
+		return nil, fmt.Errorf("error getting content from body of response with status %s: %s", response.Status, err)
+	}
+
+	log.C(ctx).Debugf("Successfully fetched $s with id %s from broker with name %s and URL %s", resourceType, resourceId, broker.Name, broker.BrokerURL)
```
alexwo, line 56: $s?

api/osb/utils.go
```diff
@@ -0,0 +1,60 @@
...
+		log.C(ctx).WithError(err).Errorf("Error while forwarding request to service broker %s", broker.Name)
+		return nil, &util.HTTPError{
+			ErrorType:   "ServiceBrokerErr",
+			Description: fmt.Sprintf("could not reach service broker %s at %s", broker.Name, broker.BrokerURL),
+			StatusCode:  http.StatusBadGateway,
+		}
+	}
+
+	if response.StatusCode != http.StatusOK {
+		log.C(ctx).WithError(err).Errorf("error fetching %s with id %s from broker with name %s: %s", resourceType, resourceId, broker.Name, util.HandleResponseError(response))
```
alexwo, line 35: incase this method is used to get the catalog it won't have the resourceID?

sigalmaya: No it wont have

alexwo: In this case, perhaps we can just omit the resourceID and replace it with resource URL (resourceID will be part of it)
we can print something as error fetching %s at URL: %s from broker with name %s: %s",resourceType,URL,broker.name,response..)

storage/utils.go
```diff
@@ -0,0 +1,31 @@
...
+
+func GetServiceByServiceInstance(repository Repository,ctx context.Context, serviceInstanceId  string) (*types.ServiceOffering, error){
+	byID := query.ByField(query.EqualsOperator, "id", serviceInstanceId)
+	criteria := query.CriteriaForContext(ctx)
+	obj, err := repository.Get(ctx, types.ServiceInstanceType, append(criteria, byID)...)
+	if err != nil {
+		return nil, util.HandleStorageError(err, types.ServiceInstanceType.String())
+	}
+	serviceInstance := obj.(*types.ServiceInstance)
+	planObject, err := repository.Get(context.TODO(), types.ServicePlanType, query.ByField(query.EqualsOperator, "id", serviceInstance.ServicePlanID))
```
alexwo, line 20: we should not use context.TODO() .. but the actual context from the request

storage/utils.go
```diff
@@ -0,0 +1,31 @@
...
+	if err != nil {
+		return nil, util.HandleStorageError(err, types.ServicePlanType.String())
+	}
+	plan := planObject.(*types.ServicePlan)
+	serviceObject, err := repository.Get(context.TODO(), types.ServiceOfferingType, query.ByField(query.EqualsOperator, "id", plan.ServiceOfferingID))
```
alexwo, line 25: we should use the request context in repository.GET calls instead of context.TODO()

storage/utils.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"context"
+	"github.com/Peripli/service-manager/pkg/query"
+	"github.com/Peripli/service-manager/pkg/types"
+	"github.com/Peripli/service-manager/pkg/util"
+
+)
+
+
+func GetServiceByServiceInstance(repository Repository,ctx context.Context, serviceInstanceId  string) (*types.ServiceOffering, error){
```
alexwo, line 12: perhaps we should rename this method to GetServiceOfferingByServiceInstanceId

api/service_instance_controller.go
```diff
@@ -83,3 +101,48 @@ func (c *ServiceInstanceController) Routes() []web.Route {
 		},
 	}
 }
+
+func (c *ServiceInstanceController) GetParameters(r *web.Request) (*web.Response, error) {
```
alexwo, line 105: perhaps we should block async=true API calls to this method as it works only in the sync mode & return async is not supported error.

PR #563, opened 2020-10-12
don't fetch labels on update labels operation

pavelmaliy:
improve performance of update labels db operation,
no need to fetch the labels in update flow
and add CountLabelValues to storage API

### DIFF
--- a/storage/interceptable_repository.go
+++ b/storage/interceptable_repository.go
@@ -216,6 +216,10 @@ func (ir *queryScopedInterceptableRepository) Count(ctx context.Context, objectT
 	return ir.repositoryInTransaction.Count(ctx, objectType, criteria...)
 }
 
+func (ir *queryScopedInterceptableRepository) CountLabelValues(ctx context.Context, objectType types.ObjectType, criteria ...query.Criterion) (int, error) {
+	return ir.repositoryInTransaction.CountLabelValues(ctx, objectType, criteria...)
+}
+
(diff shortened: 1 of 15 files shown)


storage/interceptable_repository.go
```diff
@@ -350,11 +355,14 @@ func (ir *queryScopedInterceptableRepository) Update(ctx context.Context, obj ty
...
-	obj, err := ir.repositoryInTransaction.Get(ctx, objectType, byID)
+	result, err := ir.repositoryInTransaction.ListNoLabels(ctx, objectType, byID)
 	if err != nil {
 		return err
 	}
-
+	if result == nil || result.Len() == 0 {
+		return util.ErrNotFoundInStorage
+	}
+	obj := result.ItemAt(0)
```
eyalnathan, line 365: maybe also check that you have a single result and fail if you got multiple? i know you are querying by ID, but someone might mess the query itself

storage/interfaces.go
```diff
@@ -192,6 +192,9 @@ type Repository interface {
 	// Count retrieves number of objects of particular type in SM DB
 	Count(ctx context.Context, objectType types.ObjectType, criteria ...query.Criterion) (int, error)
 
+	// Count labels of retrieved objects of particular type in SM DB
```
eyalnathan, line 195: // Count label **values** of ...

eyalnathan (approved):

PR #567, opened 2020-10-13
add tenant label to offerings and plans

evyaffe:
Tenant label is required for offerings and plans in addition to the current setup in which broker labels only has it.

### DIFF
--- a/pkg/types/base.go
+++ b/pkg/types/base.go
@@ -61,14 +61,6 @@ func (b *Base) GetLabels() Labels {
 	return b.Labels
 }
 
-func (b *Base) ExtractTenantLabelValue(tenantLabelKey string) []string {
-	tenant, ok := b.Labels[tenantLabelKey]
-	if !ok {
-		return nil
-	}
-	return tenant
-}
-
 func (b *Base) GetPagingSequence() int64 {
 	return b.PagingSequence
 }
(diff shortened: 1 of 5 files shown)


pkg/types/base.go
```diff
@@ -61,6 +61,14 @@ func (b *Base) GetLabels() Labels {
 	return b.Labels
 }
 
+func (b *Base) ExtractTenantLabelValue(tenantLabelKey string) []string {
```
eyalnathan, line 64: not used anymore, right?

storage/interceptors/broker_create_catalog_interceptor.go
```diff
@@ -141,3 +148,13 @@ func brokerCatalogAroundTx(ctx context.Context, broker *types.ServiceBroker, fet
 
 	return nil
 }
+
+func addTenantToLabels(labels types.Labels, tenantKey string, tenantValue []string) types.Labels {
```
eyalnathan, line 152: this is not adding, it is setting. wouldn't setTenantLabel be a better name?

eyalnathan: or even setLabelValue, you pass in the key as well

pkg/types/base.go
evyaffe: removed

storage/interceptors/broker_create_catalog_interceptor.go
evyaffe: changed

eyalnathan:
@evyaffe can we close this one? we can keep the branch around if we think we might need it at some point

PR #568, opened 2020-10-19
WIP: Support cascade delete

dmitry-bakaleinik-sap:
# Pull Request Template

## Prerequisites

## Motivation

## Approach

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [ ] Unit tests
* [ ] Integration tests

## Third-party code

## Work in Progress label


api/base_controller.go
```diff
@@ -260,6 +268,30 @@ func (c *BaseController) DeleteSingleObject(r *web.Request) (*web.Response, erro
 	if err != nil {
 		return nil, fmt.Errorf("could not generate GUID for %s: %s", c.objectType, err)
 	}
+	var cascadeRootId = ""
+	if opCtx.Cascade {
+		if c.supportsCascadeDelete {
+			// Scan if requested resource really exists
+			resources, err := c.repository.List(ctx, c.objectType, criteria...)
```
eyalnathan, line 275: why do we need to check this here? is it different than non-cascading flow? who's the one to return 404 when it is not cascade?

api/filters/platform_termination_filter.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (*platformTerminationFilter) Name() string {
+	return PlatformTerminationFilterName
+}
+
+func (f *platformTerminationFilter) Run(req *web.Request, next web.Handler) (*web.Response, error) {
+	platformID := req.PathParams[web.PathParamResourceID]
+
+	switch req.Request.Method {
```
eyalnathan, line 50: why switch-case for a single option?

api/filters/platform_termination_filter.go
```diff
@@ -0,0 +1,82 @@
...
+	case http.MethodDelete:
+		if platformID != "" {
```
eyalnathan, line 52: do we support DELETE /v1/platforms?fieldQuery=<whatever>? if so, this filter will not be applied in that flow

api/base_controller.go
dmitry-bakaleinik-sap: If not to do a check here, then in case when resource is not found we are getting "/v1/operation not found" error instead of. As the message is confusing, we have a check in advance to provide more accurate error message.

api/osb/platform_termination_plugin.go
```diff
@@ -0,0 +1,84 @@
...
+func (p *platformTerminationPlugin) UpdateService(req *web.Request, next web.Handler) (*web.Response, error) {
+	return p.validateNotPendingTermination(req, next)
+}
+
+// Bind intercepts bind requests and check if the instance is in the platform from where the request comes
+func (p *platformTerminationPlugin) Bind(req *web.Request, next web.Handler) (*web.Response, error) {
+	return p.validateNotPendingTermination(req, next)
+}
+
+func (p *platformTerminationPlugin) Run(req *web.Request, next web.Handler) (*web.Response, error) {
```
eyalnathan, line 42: why do we have Run here? it is not defines in any of the plugin interfaces

api/base_controller.go
```diff
@@ -270,13 +302,27 @@ func (c *BaseController) DeleteSingleObject(r *web.Request) (*web.Response, erro
...
+		ResourceID:    resourceID,
 		ResourceType:  c.objectType,
 		PlatformID:    types.SMPlatform,
 		CorrelationID: log.CorrelationIDFromContext(ctx),
-		Context:       c.prepareOperationContextByRequest(r),
+		Context:       opCtx,
+		CascadeRootID: cascadeRootId,
+	}
+	if c.supportsCascadeDelete && opCtx.Cascade {
+		concurrentOp, err := c.scheduler.FindConcurrentOperation(ctx, operation)
```
eyalnathan, line 313: wouldn't it make more sense to first check if there's a concurrent op for the same resource (type + id) and only if not found construct a new one?

operations/scheduler.go
```diff
@@ -694,6 +697,32 @@ func (s *Scheduler) executeOperationPreconditions(ctx context.Context, operation
...
+	sameResourceOperations, err := s.repository.List(ctx, types.OperationType, criteria...)
+	if err != nil {
+		return nil, err
+	}
+	if sameResourceOperations.Len() > 0 {
+		same := sameResourceOperations.ItemAt(0).(*types.Operation)
+		if same.InOrphanMitigationState() {
+			return same, nil
+		}
+		switch same.State {
```
eyalnathan, line 716: (low prio) why switch? if all options result in the same return value, why not use if with or? also applies for the InOrphanMitigationState check

storage/interceptors/smaap_virtual_resource_cascade_operation_create_interceptor.go
```diff
@@ -0,0 +1,92 @@
...
+	return func(ctx context.Context, storage storage.Repository, obj types.Object) (types.Object, error) {
+		operation := obj.(*types.Operation)
+		// currently we have only one virtual object: tenant
+		isVirtual := types.IsVirtualType(operation.ResourceType)
+		if !isVirtual || operation.CascadeRootID == "" || operation.Type != types.DELETE {
+			return f(ctx, storage, operation)
+		}
+
+		// init operation properties
+		operation.PlatformID = types.SMPlatform
```
eyalnathan, line 58: maybe init `operations` here with the relevant fields instead of creating it before you know if you need to use it?

operations/scheduler.go
```diff
@@ -694,6 +697,32 @@ func (s *Scheduler) executeOperationPreconditions(ctx context.Context, operation
 	return nil
 }
 
+func (s *Scheduler) FindConcurrentOperation(ctx context.Context, operation *types.Operation) (*types.Operation, error) {
```
eyalnathan, line 700: i noticed there's a `doesExistCascadeOperationForResource` function that does pretty much the same, consider reuse

api/base_controller.go
dmitry-bakaleinik-sap: refactored to check by resource id before creating operation

api/osb/platform_termination_plugin.go
dmitry-bakaleinik-sap: It's now removed, the OSB tests didn't showed up any regression.

operations/scheduler.go
dmitry-bakaleinik-sap: Refactored to a common function shared across this code and interceptors which require the same logic

operations/scheduler.go
dmitry-bakaleinik-sap: It was refactored to filter using query itself instead having compute logic

storage/interceptors/smaap_virtual_resource_cascade_operation_create_interceptor.go
dmitry-bakaleinik-sap: Done, the operation check was moved upper

api/filters/platform_termination_filter.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (*platformTerminationFilter) Name() string {
+	return PlatformTerminationFilterName
+}
+
+func (f *platformTerminationFilter) Run(req *web.Request, next web.Handler) (*web.Response, error) {
+	platformID := req.PathParams[web.PathParamResourceID]
+	cascadeParam := req.URL.Query().Get(web.QueryParamCascade)
+	if req.Request.Method == http.MethodDelete && platformID != "" && cascadeParam == "true" {
```
eyalnathan, line 50: my previous comment is still relevant - if i call `DELETE /v1/platform?fieldQuery=name eq 'some-name'` the filter will not run as `platformID` will be empty.

test/operations_test/cascade_test/parallel_delete_test.go
```diff
@@ -0,0 +1,62 @@
...
+			triggerCascadeOperation(context.Background(), types.TenantType, tenantID, false)
+
+			platformOperation, err := ctx.SMRepository.Get(
+				context.Background(),
+				types.OperationType,
+				query.ByField(query.EqualsOperator, "resource_id", platformID),
+				query.ByField(query.EqualsOperator, "type", string(types.DELETE)))
+
+			Expect(err).NotTo(HaveOccurred())
+			platformOperationId := platformOperation.GetID()
```
eyalnathan, line 32: test setup should be in BeforeEach

test/operations_test/cascade_test/parallel_delete_test.go
```diff
@@ -0,0 +1,62 @@
...
+			Expect(platformDeleteResponse.Header("Location").Raw()).To(HaveSuffix(platformOperationId))
+
+		})
+
+		It("Deleting tenant during platform deletion should pass", func() {
+
+			ctx.SMWithOAuth.DELETE(web.PlatformsURL+"/"+platformID).
+				WithQuery("cascade", "true").
+				Expect().
+				Status(http.StatusAccepted)
```
eyalnathan, line 48: test setup should be in BeforeEach

test/osb_and_plugin_test/plugin_test/platform_termination_plugin_test.go
```diff
@@ -0,0 +1,126 @@
...
+		osbOperations := []struct {
+			name           string
+			method         string
+			path           string
+			queries        []string
+			expectedStatus int
+		}{
+			{"fetchCatalog", "GET", "/v2/catalog", []string{""}, http.StatusOK},
+			{"provision", "PUT", "/v2/service_instances/1234", []string{""}, http.StatusUnprocessableEntity},
+			{"provision", "PUT", "/v2/service_instances/1234", []string{"accepts_incomplete=true"}, http.StatusUnprocessableEntity},
```
eyalnathan, line 97: duplicate name - maybe name this one "provisionAsync" so we can tell the difference in the test output?

test/platform_test/platform_test.go
```diff
@@ -523,23 +530,231 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+							Expect(err).ToNot(HaveOccurred())
+
+							platformFromStorage.(*types.Platform).Active = true
+							if updatedPlatform, err = storage.Update(ctx, platformFromStorage, types.LabelChanges{}); err != nil {
+								return err
+							}
+							Expect(updatedPlatform.(*types.Platform).Active).To(Equal(true))
+							return nil
+						})
+						Expect(err).ToNot(HaveOccurred())
```
eyalnathan, line 556: test setup should be in BeforeEach

test/platform_test/platform_test.go
```diff
@@ -523,23 +530,231 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
```
eyalnathan, line 580: test setup should be in BeforeEach

eyalnathan, line 580: maybe extract to function and reuse in both Its?

test/sm_test/sm_test.go
```diff
@@ -191,6 +192,17 @@ var _ = Describe("SM", func() {
...
+				var updatedPlatform types.Object
+				byID := query.ByField(query.EqualsOperator, "name", types.SMPlatform)
+				platformFromStorage, err := ctx.SMRepository.Get(context.Background(), types.PlatformType, byID)
+				Expect(err).ToNot(HaveOccurred())
+
+				platformFromStorage.(*types.Platform).Active = false
+				if updatedPlatform, err = ctx.SMRepository.Update(context.Background(), platformFromStorage, types.LabelChanges{}); err != nil {
+					Expect(err).ToNot(HaveOccurred())
+				}
+				Expect(updatedPlatform.(*types.Platform).Active).To(Equal(false))
```
eyalnathan, line 204: why do we need to set SM platform to inactive?

api/service_instance_controller.go
```diff
@@ -42,7 +42,7 @@ func NewServiceInstanceController(ctx context.Context, options *Options) *Servic
 	return &ServiceInstanceController{
 		BaseController: NewAsyncController(ctx, options, web.ServiceInstancesURL, types.ServiceInstanceType, true, func() types.Object {
 			return &types.ServiceInstance{}
-		}),
+		}, true),
```
eyalnathan, line 45: do we block creation of new bindings while instance is being cascade-deleted, like we do for platforms? I didn't see any code doing that or test covering it

api/filters/platform_termination_filter.go
dmitry-bakaleinik-sap: During design review, we agreed that **delete objects** process should stay as it. Your requirement, will demand an introduction of some extra handling for this flow.
Potentially, if you call for delete platforms (with some platform that still having sub-resources), the request will fail (as still sub-resources are in place).

Is it still relevant to introduce a new logic for delete objects flow?

api/service_instance_controller.go
dmitry-bakaleinik-sap: 1) Cascade delete functionality automatically creates sub operations for broker/instances deletion when platform deletion is requested.
2) Once delete operation exists for broker/instance, it is already protected by delete handling logic. For example: https://github.com/Peripli/service-manager/blob/700d4b92ac9159ce165c6faea87a1fd5009f75c8/storage/interceptors/smaap_service_binding_interceptor.go#L140
3) We have a set of requests (including binding, unbinding and etc...) with check for expected response code. We are expecting to get unprocessable entity when calling for instance binding during platform deletion - https://github.com/Peripli/service-manager/blob/50d53f181f744b5ed6c3bfc08c9616d9d7f4a77a/test/osb_and_plugin_test/plugin_test/platform_termination_plugin_test.go#L101.

test/operations_test/cascade_test/parallel_delete_test.go
dmitry-bakaleinik-sap: done

test/operations_test/cascade_test/parallel_delete_test.go
dmitry-bakaleinik-sap: done

test/osb_and_plugin_test/plugin_test/platform_termination_plugin_test.go
dmitry-bakaleinik-sap: doe

test/platform_test/platform_test.go
dmitry-bakaleinik-sap: done

test/platform_test/platform_test.go
dmitry-bakaleinik-sap: done

dmitry-bakaleinik-sap: done

test/sm_test/sm_test.go
dmitry-bakaleinik-sap: Removed (once we had a filter that handled delete objects flow, and that filter prevented tests in this case to work)

eyalnathan (approved):

PR #601, opened 2020-12-14
Healthcheck

sigalmaya:
This pull request adds a new health check that monitored platforms that are labeled as "monitored"


api/healthcheck/monitored_platform_indicator_test.go
```diff
@@ -0,0 +1,136 @@
+package healthcheck
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/Peripli/service-manager/pkg/health"
+	"github.com/Peripli/service-manager/pkg/types"
+	storagefakes2 "github.com/Peripli/service-manager/storage/storagefakes"
```
eyalnathan, line 9: why storagefakes2?

api/healthcheck/helper.go
```diff
@@ -0,0 +1,29 @@
...
+	fatalInactivePlatforms := 0
+	for _, platform := range platforms {
+		healthObj := health.New().WithDetail("type", platform.Type)
+		if platform.Active {
+			healthObj = healthObj.WithStatus(health.StatusUp)
+		} else {
+			inactivePlatforms++
+			healthObj.WithStatus(health.StatusDown).WithDetail("since", platform.LastActive)
+			if fatal != nil && fatal(platform) {
+				fatalInactivePlatforms++
```
eyalnathan, line 20: don't we want an indicator in the details in case it is fatal?

sigalmaya: I didnt get quite the question. You have details always, what fatal function gives you is an option to decide based on some object which platform will consider as not healthy, for example, the old platform health indicator had a function plaform.type = cf , so only inactive platform of type cf could change the status of the healthcheck.

eyalnathan (approved):

PR #604, opened 2020-12-15
Check for instance visibility without returning the label list to sm

alexwo:
## Motivation

Improve the check_instance_visiblity plugin and filter by checking for instance visibility without returning a full label list that matches a plan & platform visibility to sm.

## Approach

Refracture the code to perform the visibility check on the DB layer (by using 2 named queries)

## Pull Request status

* [x ] Refactoring

### DIFF
--- a/api/osb/check_instance_visibility_plugin.go
+++ b/api/osb/check_instance_visibility_plugin.go
@@ -6,7 +6,6 @@ import (
 	"github.com/Peripli/service-manager/pkg/query"
 	"github.com/Peripli/service-manager/pkg/types"
 	"github.com/Peripli/service-manager/pkg/util"
-	"github.com/Peripli/service-manager/pkg/util/slice"
 	"github.com/Peripli/service-manager/pkg/web"
 	"github.com/Peripli/service-manager/storage"
 	"github.com/tidwall/gjson"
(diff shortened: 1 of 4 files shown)


storage/named_query.go
```diff
@@ -56,6 +58,23 @@ var namedQueries = map[NamedQuery]string{
 	SELECT v.* FROM visibilities v
 	LEFT OUTER JOIN visibility_labels vl on v.id = vl.visibility_id
 	WHERE (vl.id IS NULL and v.platform_id in (:platform_ids)) OR v.platform_id IS NULL`,
+	QueryForVisibility: `
+	SELECT v.* FROM visibilities v
+	INNER JOIN visibility_labels vl on v.id = vl.visibility_id
+	WHERE key=:key and val = :val and platform_id = :platform_id and service_plan_id = :service_plan_id`,
+
```
pavelmaliy, line 65: why to do generic template if you only support visibility table?

api/osb/check_instance_visibility_plugin.go
```diff
@@ -114,35 +106,40 @@ func (p *checkVisibilityPlugin) checkVisibility(req *web.Request, next web.Handl
...
-				}
-				orgGUIDs, ok := v.Labels["organization_guid"]
-				if !ok {
-					return next.Handle(req)
-				}
-				if slice.StringsAnyEquals(orgGUIDs, payloadOrgGUID) {
-					return next.Handle(req)
-				}
-			}
+
```
pavelmaliy, line 109: why do you need named query here query.ByField + query.ByLabel won't do the job?

storage/named_query.go
pavelmaliy, line 65: why to do generic template if you only support visibility table?

storage/named_query.go
```diff
@@ -61,6 +62,12 @@ var namedQueries = map[NamedQuery]string{
 	SELECT v.* FROM visibilities v
 	LEFT OUTER JOIN visibility_labels vl on v.id = vl.visibility_id
 	WHERE (vl.id IS NULL and v.service_plan_id in (:service_plan_ids))`,
```
pavelmaliy, line 64: alternative query:
select * from visibilities v
left outer join visibility_labels l on v.id = l.visibility_id
where v.platform_id is null or l.id is null or (l.key=:key and l.val=:val)

dmitry-bakaleinik-sap: We see a big performance degradation when left outer join is used.
For example if we have 50k visibilities and 5mil visibility labels such join will create a big temporary set in which filtering using where is done after.

Our approach is to filter using where condition only. In our query some visibilities will be taken by platform criteria and for only for produced subset of visibilities a subquery will be performed.

pavelmaliy: much more cumbersome but if you say big performance degradation I can't argue with that :)

eyalnathan (approved):

PR #606, opened 2021-01-10
get supported agents versions api

sigalmaya:
- Adding API that returns the supported agents' versions by the service manager. 
- The supported versions API reads the configurations agent settings.  The configuration could be injected as env variables.


api/notifications/notifications.go
```diff
@@ -32,10 +32,20 @@ func (c *Controller) handleWS(req *web.Request) (*web.Response, error) {
 		return nil, errors.New("user details not found in request context")
 	}
 
+
 	platform, err := extractPlatformFromContext(user)
 	if err != nil {
 		return nil, err
 	}
+	version:= req.Header.Get("X-Version")
```
eyalnathan, line 40: can we have a more specific header here? e.g. `X-Peripli-Agent-Version`

pkg/types/platform.go
```diff
@@ -53,6 +53,7 @@ type Platform struct {
 	Description       string       `json:"description"`
 	Credentials       *Credentials `json:"credentials,omitempty"`
 	OldCredentials    *Credentials `json:"old_credentials,omitempty"`
+	Version           string       `json:"-"`
```
eyalnathan, line 56: don't we want it returned in the platforms API? same question for "active" (i know it was there before)

storage/postgres/platform.go
```diff
@@ -42,6 +42,7 @@ type Platform struct {
 	CredentialsActive bool           `db:"credentials_active"`
 	LastActive        time.Time      `db:"last_active"`
 	Technical         bool           `db:"technical"`
+	Version           string 	     `db:"version"`
```
eyalnathan, line 45: indentation (or does it only look like this in the browser?)

eyalnathan: there a few more indentation "issues", can you `go fmt` everything?

sigalmaya: only in the browser

pkg/types/platform.go
sigalmaya: Are active and version field allowed to be changed via API ?

api/notifications/notifications.go
```diff
@@ -36,6 +37,15 @@ func (c *Controller) handleWS(req *web.Request) (*web.Response, error) {
 	if err != nil {
 		return nil, err
 	}
+	version := req.Header.Get(AgentVersionHeader)
+	if version != "" {
+		platform.Version = version
+		_, err = c.repository.Update(ctx, platform, nil)
```
eyalnathan, line 43:
```suggestion
		if platform.Version != version {
		  platform.Version = version
		  _, err = c.repository.Update(ctx, platform, nil)
		}
```

api/notifications/notifications.go
```diff
@@ -36,6 +37,15 @@ func (c *Controller) handleWS(req *web.Request) (*web.Response, error) {
...
+		if err != nil {
+			logger.Errorf("Could not update platform version for platform %s: %v", platform.ID, err)
+			return nil, err
+		}
```
eyalnathan, line 47:
```suggestion
	if platform.Version != version {	
		platform.Version = version
		_, err = c.repository.Update(ctx, platform, nil)
		if err != nil {
			logger.Errorf("Could not update platform version for platform %s: %v", platform.ID, err)
			return nil, err
		}
	}
```

pkg/env/env.go
```diff
@@ -133,10 +134,17 @@ func (v *ViperEnv) Unmarshal(value interface{}) error {
 			return err
 		}
 	}
-
 	return v.Viper.Unmarshal(value)
+
 }
 
+func (v *ViperEnv) parseJson() {
+	versions := v.Viper.GetStringMapStringSlice("agents.versions")
```
eyalnathan, line 142: is this the right way to do it? seems weird to have code for a specific config property in here.

pkg/agents/agents_version.go
```diff
@@ -0,0 +1,12 @@
+package agents
+
+type Settings struct {
+	SupportedVersions map[string][]string `mapstructure:"supported_versions"`
+	VersionsEnv       string              `mapstructure:"versions"`
```
eyalnathan, line 5: do we really need both?

eyalnathan (approved):

PR #641, opened 2021-02-22
Return an appropriate error in case the request is not supported by t…

sigalmaya:
Return an appropriate error in case the request is not supported by the service manager. 
This pull request returns an error when user requests to filter service plans/ service offerings by the environment type.

### DIFF
--- a/test/disabled_query_params_test/disabled_query_parameters_test.go
+++ b/test/disabled_query_params_test/disabled_query_parameters_test.go
@@ -1,0 +1,75 @@
+package disabled_query_params_test
+
+import (
+	"github.com/Peripli/service-manager/pkg/web"
+	"github.com/Peripli/service-manager/test/common"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/pflag"
+	"net/http"
+	"testing"
+)
+
+func TestDisabledQuery(t *testing.T) {
+	RegisterFailHandler(Fail)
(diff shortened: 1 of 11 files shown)


test/disabled_query_params_test/disabled_query_parameters_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+		BeforeEach(func() {
+			ctxBuilder = common.NewTestContextBuilderWithSecurity().WithEnvPreExtensions(func(set *pflag.FlagSet) {
+				Expect(set.Set("api.disabled_query_parameters", "")).ToNot(HaveOccurred())
+			})
+			ctx = ctxBuilder.Build()
+		})
+
+		Context("the query param is provided", func() {
+			It("Returns return an error", func() {
```
eyalnathan, line 38: shouldn't this be It("does not return an error")?

test/disabled_query_params_test/disabled_query_parameters_test.go
```diff
@@ -0,0 +1,75 @@
...
+				ctx.SMWithOAuth.GET(web.ServicePlansURL).WithQuery("environment", "cf").
+					Expect().
+					Status(http.StatusOK)
+			})
+		})
+
+	})
+
+	Describe("query param is disabled", func() {
```
eyalnathan, line 47: can you add another test case here that checks it with a different value set in the env instead of the default "environment" value?

eyalnathan (approved):

PR #654, opened 2021-04-13
contains operator for fieldQuery and labelQuery

sigalmaya:
This pull request features support for a new operator "contains" operator. "contains operator is applicable for fieldQuery and labelQuery for get and delete APIs.
for example to retrieve all platforms that their name contains "service", use:
v1/platforms?fieldQuery="name contains 'service'"


pkg/query/operators.go
```diff
@@ -25,6 +25,8 @@ const (
 	GreaterThanOperator gtOperator = "gt"
 	// GreaterThanOrEqualOperator takes two operands and tests if the left is greater than or equal the right
 	GreaterThanOrEqualOperator geOperator = "ge"
+	// Contains takes two operands and tests if the left contains the right
+	ContainsOPerator containsOperator = "contains"
```
alexwo, line 29: should be ContainsOperator instead ContainsOPerator?

sigalmaya: its O not null, it looks like null but is not

eyalnathan:
```suggestion
	ContainsOperator containsOperator = "contains"
```

sigalmaya: ahh sorry you mean P

dmitry-bakaleinik-sap (approved):

PR #658, opened 2021-05-18
provision reference instance via selector

EpsilonTal:
**Motivation**
Sharing a service instance between different scopes allows apps in different scopes to share databases, messaging queues, and other types of services. This enables development teams to share a resource, while eliminates the use of service bindings as keys and user-provided services or other means to maintain credentials of instances that were provisioned in another scope.
  
**User Story**

As a developer, I would like to be able to refer to a shared instance without the need to provide the ID of the original instance, so that I could apply a robust automated script that can be executed in different landscapes and setups.


api/common/sharing/utils.go
```diff
@@ -10,54 +12,108 @@ import (
...
-		return "", util.HandleStorageError(util.ErrNotFoundInStorage, types.ServicePlanType.String())
+	filteredInstancesMap := map[string]*types.ServiceInstance{}
+	referencedInstanceID := parameters[instance_sharing.ReferencedInstanceIDKey].String()
+	if len(referencedInstanceID) > 1 {
+		filteredInstancesMap, err = filterInstancesByID(ctx, repository, referencedInstanceID, tenantIdentifier, getTenantId())
+	} else {
+		filteredInstancesMap, err = filterInstancesBySelectors(ctx, repository, sharedInstancesMap, parameters, smaap)
 	}
-	// verify the reference plan and the target instance are of the same service offering:
-	sameOffering, err := isSameServiceOffering(ctx, repository, referencePlan.ServiceOfferingID, referencedInstance.ServicePlanID)
```
EpsilonTal, line 36: sameOffering moved to `filterInstancesByID` function

api/common/sharing/utils.go
```diff
@@ -10,56 +12,105 @@ import (
...
+		}
+	} else {
+		filteredInstancesMap := map[string]*types.ServiceInstance{}
+		filteredInstancesMap, err = filterInstancesBySelectors(ctx, repository, sharedInstancesMap, parameters, smaap)
+		err = validateSelectorResults(filteredInstancesMap, parameters)
+		if err != nil {
+			return "", err
+		}
+
+		referencedInstance, err = retrieveInstanceFromMap(sharedInstancesMap)
```
pavelmaliy, line 48: what this function does???
your looping over map and return the last result

api/common/sharing/utils.go
```diff
@@ -12,44 +12,236 @@ import (
...
+
+		selectorVal, exists = parameters[instance_sharing.ReferenceInstanceNameSelector]
+		if exists {
+			selectors["name"] = instance.Name == selectorVal.String()
+		}
+
+		selectorVal, exists = parameters[instance_sharing.ReferencePlanNameSelector]
+		if exists {
+			selectors["service_plan_id"], err = foundSharedInstanceWithPlan(ctx, repository, instance, selectorVal.String(), smaap)
+			if err != nil {
```
pavelmaliy, line 167: you can do it with one query and not query per instance
first bring all plans of shared instances and just look in the result

EpsilonTal: in that case, we better use a different named query with inner join on service plans

that brings us back to the question of single named query or many...

api/common/sharing/utils.go
```diff
@@ -10,46 +10,224 @@ import (
...
+	}
+	return nil
+}
 
+func getReferencedInstancesByID(ctx context.Context, repository storage.Repository, referencedInstanceID string, tenantIdentifier string, tenantValue string) (*types.ServiceInstance, error) {
 	referencedInstanceObj, err := repository.Get(ctx, types.ServiceInstanceType,
-		query.ByLabel(query.EqualsOperator, tenantIdentifier, getTenantId()),
-		query.ByField(query.EqualsOperator, "id", referencedInstanceID.String()),
+		query.ByLabel(query.EqualsOperator, tenantIdentifier, tenantValue),
+		query.ByField(query.EqualsOperator, "id", referencedInstanceID),
```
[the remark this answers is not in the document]
EpsilonTal: I think we do need it, we have it in the context but it is not good for us.
We have checked it before for the instance sharing feature.

Also, this line covers the flow of the container.

We can't use the criteria from context for the container flow because it also has the `container_id` key.
So when using the criteria from context for instances in a container, you actually filter by 2 keys.
We want to allow instance sharing for containers so need just the `tenantIdentifier`.

api/common/sharing/utils.go
```diff
@@ -10,46 +10,232 @@ import (
...
+	}
+	parametersRaw := gjson.GetBytes(req.Body, "parameters").Raw
+	if parametersRaw != "" {
+		return util.HandleInstanceSharingError(util.ErrChangingParametersOfReferenceInstance, instance.ID)
 	}
+	return nil
+}
 
+func filterInstancesByID(ctx context.Context, repository storage.Repository, referencedInstanceID string, tenantIdentifier string, tenantValue string) (*types.ServiceInstance, error) {
 	referencedInstanceObj, err := repository.Get(ctx, types.ServiceInstanceType,
```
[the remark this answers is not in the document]
EpsilonTal: done, probably old code you are seeing.

api/common/sharing/utils.go
```diff
@@ -10,46 +10,224 @@ import (
...
+	sharedInstances, err := getSharedInstances(ctx, repository, tenantIdentifier, tenantID)
+	if err != nil {
+		return nil, util.HandleStorageError(err, types.ServiceInstanceType.String())
+	}
+	return convertObjectListToInstancesMap(sharedInstances), nil
+}
+
+func validateParameters(parameters map[string]gjson.Result) error {
+	if len(parameters) == 0 {
+		return util.HandleInstanceSharingError(util.ErrMissingOrInvalidReferenceParameter, instance_sharing.ReferencedInstanceIDKey)
```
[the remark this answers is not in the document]
EpsilonTal: old code.

i removed it from the `validateParameters` function.
we do have 
`if results.Len() > 1 && len(parameters) == 0 {
		return util.HandleInstanceSharingError(util.ErrMissingOrInvalidReferenceParameter, instance_sharing.ReferencedInstanceIDKey)
	}`

in the `validateSingleResult` function

api/common/sharing/utils.go
```diff
@@ -10,56 +12,105 @@ import (
...
+	var referencedInstance *types.ServiceInstance
+	referencedInstanceID := parameters[instance_sharing.ReferencedInstanceIDKey].String()
+	if len(referencedInstanceID) > 1 {
+		referencedInstance, err = filterInstancesByID(ctx, repository, referencedInstanceID, tenantIdentifier, getTenantId())
+		if err != nil {
+			return "", err
+		}
+	} else {
+		filteredInstancesMap := map[string]*types.ServiceInstance{}
+		filteredInstancesMap, err = filterInstancesBySelectors(ctx, repository, sharedInstancesMap, parameters, smaap)
```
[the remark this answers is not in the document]
EpsilonTal: done

api/common/sharing/utils.go
```diff
@@ -10,56 +12,105 @@ import (
...
 
-	if !exists {
-		return "", util.HandleInstanceSharingError(util.ErrMissingOrInvalidReferenceParameter, instance_sharing.ReferencedInstanceIDKey)
+	err = validateParameters(parameters)
+	if err != nil {
+		return "", err
+	}
+
+	sharedInstancesMap := map[string]*types.ServiceInstance{}
+	sharedInstancesMap, err = getSharedInstancesByTenant(ctx, repository, tenantIdentifier, getTenantId())
```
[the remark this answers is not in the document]
EpsilonTal: done

api/common/sharing/utils.go
EpsilonTal: as discussed: moved from a map to a list of instances.
removed the converters functions

api/common/sharing/utils.go
```diff
@@ -10,56 +12,105 @@ import (
...
+	if err != nil {
+		return "", err
 	}
 
+	var referencedInstance *types.ServiceInstance
+	referencedInstanceID := parameters[instance_sharing.ReferencedInstanceIDKey].String()
+	if len(referencedInstanceID) > 1 {
+		referencedInstance, err = filterInstancesByID(ctx, repository, referencedInstanceID, tenantIdentifier, getTenantId())
```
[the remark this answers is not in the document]
EpsilonTal: as discussed, changed to condition of `exists && referencedInstanceID.String() != "*"`

api/common/sharing/utils.go
EpsilonTal: It brings us back to the question of a single `named_query` or many.
I think when we have a `plan_name_selector` we should run another `named_query` with inner join on service plans.
Instead of comparing two lists.

What do you think?

api/common/sharing/utils.go
```diff
@@ -12,44 +12,236 @@ import (
...
+}
+
+func foundSharedInstanceWithLabels(instance *types.ServiceInstance, labels []byte) (bool, error) {
+	var selectorLabels types.Labels
+	if err := util.BytesToObject(labels, &selectorLabels); err != nil {
+		return false, err
+	}
+	/*
+		todo: remove after review
+		should accept this label selector:
```
EpsilonTal, line 214: test for example:
`creates reference instance by label selector`

pkg/util/errors.go
```diff
@@ -281,6 +284,24 @@ func HandleInstanceSharingError(err error, entityName string) error {
...
+	case ErrInvalidReferenceSelectors:
+		return &HTTPError{
+			ErrorType:   "BadRequest",
+			Description: "Failed to create the reference. Your query selectors are invalid.",
+			StatusCode:  http.StatusBadRequest,
+		}
+	case ErrNoResultsForReferenceSelector:
+		return &HTTPError{
+			ErrorType:   "NotFound",
+			Description: "Failed to create the reference. Could not find a shared instance with the selectors provided.",
```
i522006, line 302: Couldn't find a shared instance with the provided selectors.

EpsilonTal: Done

api/common/sharing/utils.go
pavelmaliy: no, just bring corresponding plans after you brought the shared instances
and look in there instead of query per instance
if you have 5 instances and user asked by plan you'll end with 5 queries

EpsilonTal: done, the query moved to the start of the selector function.
so if you pass the plan selector, we get the plan with its `offeringID`, if found a plan we compare it.

single request instead of multiple for each instance.

api/common/sharing/utils.go
```diff
@@ -12,44 +12,234 @@ import (
...
+			log.C(ctx).Errorf("%s", util.ErrMultipleReferenceSelectorResults)
+			return types.ServiceInstances{}, util.HandleInstanceSharingError(util.ErrMultipleReferenceSelectorResults, "")
+		}
+	}
+	if filteredInstances.Len() == 0 {
+		log.C(ctx).Errorf("%s", util.ErrNoResultsForReferenceSelector)
+		return types.ServiceInstances{}, util.HandleInstanceSharingError(util.ErrNoResultsForReferenceSelector, "")
+	}
+	return filteredInstances, nil
+}
```
pavelmaliy, line 187: return *types.ServiceInstance no need for array

api/common/sharing/utils.go
```diff
@@ -12,44 +12,234 @@ import (
...
+
+			instanceLabels: {
+				"key": ["x", "y"],
+				"region": ["us", "eu", "jp"]
+			}
+	*/
+
+	for selectorLabelKey, selectorLabelArray := range selectorLabels {
+		instanceLabelVal, exists := instanceLabels[selectorLabelKey]
+		if exists && instanceLabels != nil {
```
pavelmaliy, line 229: change instanceLabelVal to instanceLabelArray it's confusing in contains(instanceLabelVal, selectorLabelVal)

api/common/sharing/utils.go
```diff
@@ -12,44 +12,234 @@ import (
...
+			match := true
+			for _, selectorLabelVal := range selectorLabelArray {
+				match = match && contains(instanceLabelVal, selectorLabelVal)
+				if !match {
```
pavelmaliy, line 233: why you need all this games with booleans?
if !contains(instaLabelArray, selectorLabelVal) {
    return false, nil
}

api/common/sharing/utils.go
EpsilonTal: fixed

api/common/sharing/utils.go
EpsilonTal: fixed

api/common/sharing/utils.go
EpsilonTal: done

api/common/sharing/utils.go
```diff
@@ -62,14 +240,11 @@ func isSameServiceOffering(ctx context.Context, repository storage.Repository, o
 	return offeringID == servicePlan.ServiceOfferingID, nil
 }
```
pavelmaliy, line 242: cnt, err := repository.Count(byID, byOferringID)
return cnt > 0

api/common/sharing/utils.go
```diff
@@ -12,44 +12,223 @@ import (
...
+		if selectorPlan != nil && instance.ServicePlanID != selectorPlan.ID {
+			continue
+		}
+
+		selectorVal, exists = parameters[instance_sharing.ReferenceLabelSelector]
+		if exists {
+			var selectorLabels types.Labels
+			if err := util.BytesToObject([]byte(selectorVal.Raw), &selectorLabels); err != nil {
+				log.C(ctx).Errorf("%s", err)
+				return nil, err
```
pavelmaliy, line 171: if you have nothing to add don't log the error it will be logged later
you have more than one

api/common/sharing/utils.go
```diff
@@ -12,44 +12,223 @@ import (
...
+			}
+			if len(selectorLabels) > 0 {
+				match, err := matchLabels(instance.Labels, selectorLabels)
+				if err != nil {
+					log.C(ctx).Errorf("%s", err)
+					return nil, err
```
pavelmaliy, line 177: same

api/common/sharing/utils.go
EpsilonTal: i will lower the amount of logs.

storage/named_query.go
```diff
@@ -83,6 +84,23 @@ var namedQueries = map[NamedQuery]string{
...
+	from service_instances
+	inner join service_plans on service_plans.id = service_instances.service_plan_id
+	inner join service_instance_labels on service_instances.id = service_instance_labels.service_instance_id
+	where service_instances.shared=true and
+		  service_plans.service_offering_id = :offering_id and
+		  exists (
+			SELECT * FROM service_instance_labels 
+			WHERE 	service_instance_labels.key = :tenant_identifier and 
+					service_instance_labels.val = :tenant_id
+		  )`,
```
pavelmaliy, line 103: are you sure this query works?
you need to add in exists and service_instance_labels.service_instance_id = service_instances.id

pavelmaliy: I think you need much more tests

EpsilonTal: I just found it on an ownership test, fixing it.

EpsilonTal: added tests of service offering and ownership validation for each selector:

when shared instance owned by a different tenant
- should fail to provision with plan name selector
- should fail to provision with instance name selector
- should fail to provision with instance id=(*) selector
- should fail to provision with a label selector

when a shared instance is of a different service offering
- should fail to provision with plan name selector
- should fail to provision with instance name selector
- should fail to provision with a label selector

pavelmaliy: it won't check your fix
you need to check multiple labels selector

EpsilonTal: we do have 
`when the tenant has many shared instances` 
-> `it creates a reference to an instance by label selector`
-> `it should fail to create a reference with labels selector if one of the labels does not match`

EpsilonTal: also added now:
-> `it creates a reference to an instance by label selector if at least one label key matches a single selector value`
-> `it creates a reference to an instance by label selector if at least one label matches`

pavelmaliy: yes you right the query bug is for tenancy check, the exists will return always  true

EpsilonTal: also added now:
-> `it creates a reference to an instance by label selector if the instance has at least one selector label value`
-> `it creates a reference to an instance by label selector if the instance has at least one selector label (with its all values)`

pavelmaliy: The description of test unclear what does it mean?
"if the instance has at least one selector label value"
and this:
"if the instance has at least one selector label (with its all values)"
Describe("creating reference")
When("selector by labels")
It(should succeed blah blah)
If I don't understand no one will understand

pavelmaliy: "at least" is confusing when you need all labels to match

EpsilonTal: Do you think we need tests for both OSB and SMAAP flows?
I have some duplications that I can delete, I think we can keep only the SMAAP tests since it's a common util file.

EpsilonTal: I will rename the tests

pavelmaliy: you need both for basic flows (byID, byLabels, byPlan ....)
and you can create dedicated test file to run all cases on the common logic to test ExtractReferenceID

EpsilonTal: OK great, this is the current status

EpsilonTal: we have this for the OSB flow: 
image

EpsilonTal: btw, we don't have a dedicated test file for the selectors.
we run it from `service_instance_tests.go`
under the `instance sharing` section

api/common/sharing/utils.go
```diff
@@ -12,64 +12,236 @@ import (
...
 	}
 
-	return referencedInstanceID.String(), nil
+	for i := 0; i < instances.Len(); i++ {
+		// "*" for any shared instance of the tenant, if parameter not passed, set true for any shared instance
+		selectorVal, exists := parameters[instance_sharing.ReferencedInstanceIDKey]
+		if exists && len(selectorVal.String()) > 0 && selectorVal.String() != "*" {
+			continue
+		}
+
```
pavelmaliy, line 153: this case can be removed
if you had id you won't be in filterInstancesBySelectors func
and you block id with other selectors

api/common/sharing/utils.go
```diff
@@ -12,64 +12,229 @@ import (
...
+		return util.HandleInstanceSharingError(util.ErrMissingOrInvalidReferenceParameter, instance_sharing.ReferencedInstanceIDKey)
+	}
+	ID, IDExists := parameters[instance_sharing.ReferencedInstanceIDKey]
+	name, nameExists := parameters[instance_sharing.ReferenceInstanceNameSelector]
+	plan, planExists := parameters[instance_sharing.ReferencePlanNameSelector]
+	labels, labelsExists := parameters[instance_sharing.ReferenceLabelSelector]
+	var selectorLabels types.Labels
+	if err := util.BytesToObject([]byte(labels.Raw), &selectorLabels); labelsExists && err != nil {
+		return err
 	}
```
pavelmaliy, line 113: util.BytesToObject works??
don't you need to init selectorLabels?
check if labelsExists before unmarshaling

api/common/sharing/utils.go
```diff
@@ -12,64 +12,229 @@ import (
...
+
+		if selectorPlan != nil && instance.ServicePlanID != selectorPlan.ID {
+			continue
+		}
+
+		selectorVal, exists = parameters[instance_sharing.ReferenceLabelSelector]
+		if exists {
+			var selectorLabels types.Labels
+			if err := util.BytesToObject([]byte(selectorVal.Raw), &selectorLabels); err != nil {
+				return nil, err
```
pavelmaliy, line 162: util.BytesToObject works without initializing selectorLabels?
if not, it's really strange that the tests are green

EpsilonTal: It works, but I initialized it anyway.

api/common/sharing/utils.go
EpsilonTal: it works but initialized it anyway.
moved the `labelsExists` condition from the error handling to the unmarshal step.

api/common/sharing/utils.go
```diff
@@ -12,64 +12,231 @@ import (
...
+	}
+	selectorPlan := planObject.(*types.ServicePlan)
+	return selectorPlan, nil
+}
+
+func matchLabels(instanceLabels types.Labels, selectorLabels types.Labels) (bool, error) {
+	for selectorLabelKey, selectorLabelArray := range selectorLabels {
+		instanceLabelArray, exists := instanceLabels[selectorLabelKey]
+		if exists && instanceLabels != nil {
+			for _, selectorLabelVal := range selectorLabelArray {
```
pavelmaliy, line 212: you mean
if exists && len(instanceLabelArray) >0
right?
you can just check 
if exists &&  !contains(instanceLabelArray, selectorLabelVal) {
    return false, nil
}

}

api/common/sharing/utils.go
```diff
@@ -12,64 +12,231 @@ import (
...
+
+		selectorVal, exists = parameters[instance_sharing.ReferenceLabelSelector]
+		if exists {
+			selectorLabels := types.Labels{}
+			if err := util.BytesToObject([]byte(selectorVal.Raw), &selectorLabels); err != nil {
+				return nil, err
+			}
+			if len(selectorLabels) > 0 {
+				match, err := matchLabels(instance.Labels, selectorLabels)
+				if err != nil {
```
pavelmaliy, line 168: if  instance.Labels is nil you will get an error

api/common/sharing/utils.go
pavelmaliy: you mean
if exists && len(instanceLabelArray) >0
right?

api/common/sharing/utils.go
pavelmaliy, line 212: you mean
if exists && len(instanceLabelArray) >0
right?

pavelmaliy (approved):

pkg/util/errors.go
```diff
@@ -282,6 +285,24 @@ func HandleInstanceSharingError(err error, entityName string) error {
 			Description: fmt.Sprintf("Failed to create the reference. The instance %s, for which you want to create the reference, does not match the service offering.", entityName),
```
i522006, line 285: "Failed to create the reference. The Instance %s, for which you want to create the reference, belongs to a different service."

pkg/util/errors.go
```diff
@@ -282,6 +285,24 @@ func HandleInstanceSharingError(err error, entityName string) error {
...
 			StatusCode:  http.StatusNotFound,
 		}
+	case ErrMultipleReferenceSelectorResults:
+		return &HTTPError{
+			ErrorType:   "BadRequest",
+			Description: "Failed to create the reference. There are multiple shared instances that match your criteria. Specify explicitly the instance id you would like to refer to.",
```
i522006, line 291: "Failed to create the reference. There are multiple shared instances that match your criteria. Specify the ID of the instance to which you want to refer."

pkg/util/errors.go
EpsilonTal: do you mean "The instance"? 
you wrote "The Instance"

idobergerfroind (approved):

PR #661, opened 2021-06-01
Reference schema

sigalmaya:


schemas/reference_plan.json
```diff
@@ -0,0 +1,34 @@
...
+    "supportedPlatforms": [],
+    "translations": {
+      "en-US": {
+        "displayName": "reference-instance",
+        "description": "Allows to create a reference to a shared service instance from any environment in a subaccount and manage service bindings to that service instance."
+      }
+    }
+  },
+  "schemas": {
+    "service_instance": {a
```
EpsilonTal, line 12: @sigalmaya  - i think `a` is here by mistake (:

storage/interceptors/broker_create_catalog_interceptor.go
```diff
@@ -18,10 +18,12 @@ package interceptors
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"github.com/Peripli/service-manager/pkg/instance_sharing"
 	"github.com/Peripli/service-manager/pkg/types"
 	"github.com/Peripli/service-manager/pkg/util"
+	schemas2 "github.com/Peripli/service-manager/schemas"
```
EpsilonTal, line 26: @sigalmaya  - do we need to name the imported package `schemas2`?

pkg/sm/sm.go
```diff
@@ -211,7 +211,6 @@ func New(ctx context.Context, cancel context.CancelFunc, e env.Environment, cfg
 	smb.RegisterPluginsBefore(osb.OSBStorePluginName, osb.NewCheckVisibilityPlugin(interceptableRepository))
 	smb.RegisterPlugins(osb.NewCheckPlatformIDPlugin(interceptableRepository))
 	smb.RegisterPlugins(osb.NewPlatformTerminationPlugin(interceptableRepository))
-	smb.RegisterPlugins(osb.NewInstanceSharingPlugin(transactionalRepository, cfg.Multitenancy.LabelKey))
```
EpsilonTal, line 214: @sigalmaya - did you move this line to another place?

EpsilonTal, line 214: @sigalmaya - did you move this line to another place?

EpsilonTal (changes requested):

schemas/schemas.go
```diff
@@ -0,0 +1,36 @@
+package schemas
+
+const ReferencePlan = `{
```
EpsilonTal, line 3: Where are the other properties of the plan being defined?

`referencePlan.ID = UUID.String()`
`referencePlan.CatalogID = UUID.String()`
`referencePlan.CatalogName = instance_sharing.ReferencePlanName`
`referencePlan.Name = instance_sharing.ReferencePlanName`
`referencePlan.Description = instance_sharing.ReferencePlanDescription`
`referencePlan.ServiceOfferingID = serviceOfferingId`
`referencePlan.Bindable = newTrue()`
`referencePlan.Ready = true`
`referencePlan.CreatedAt = time.Now()`
`referencePlan.UpdatedAt = time.Now()`

EpsilonTal: Saw now, on `generateReferencePlanObject`

why do we split them? 
won't it be better to have everything in the same place?

storage/interceptors/broker_create_catalog_interceptor.go
```diff
@@ -202,6 +205,13 @@ func generateReferencePlanObject(serviceOfferingId string) *types.ServicePlan {
 	referencePlan.Ready = true
 	referencePlan.CreatedAt = time.Now()
 	referencePlan.UpdatedAt = time.Now()
+	schema, metadata, err := schemas.SchemaLoader(schemas.ReferencePlan)
```
EpsilonTal, line 208: same as above,
can we move the entire setting of the new reference plan to the JSON file?

EpsilonTal: so `schemas.SchemaLoader` will return a `servicePlan` object instead of `schema, metadata`

EpsilonTal (changes requested):

schemas/predefined_plans.go
```diff
@@ -0,0 +1,70 @@
...
+	"encoding/json"
+	"fmt"
+	"github.com/Peripli/service-manager/pkg/types"
+	"github.com/gofrs/uuid"
+	"time"
+)
+
+const ReferencePlan = `{
+ "name": "reference-instance",
+  "catalog_name": "reference-instance",
```
EpsilonTal, line 13: I see many times the same names and strings (for the description for example)
I think it will be better to define them once under a variable so if we have changes we will be aligned for each.

That way we avoid mismatches.

Like we used to have:
`referencePlan.CatalogName = instance_sharing.ReferencePlanName`
`referencePlan.Name = instance_sharing.ReferencePlanName`
`referencePlan.Description = instance_sharing.ReferencePlanDescription`

EpsilonTal (changes requested):

schemas/predefined_plans.go
```diff
@@ -0,0 +1,70 @@
...
+  "bindable": true,
+  "ready": true,
+ "metadata": {
+    "supportedPlatforms": [
+      
+    ],
+    "translations": {
+      "en-US": {
+        "displayName": "reference-instance",
+        "description": "Allows to create a reference to a shared service instance from any environment in a subaccount and manage service bindings to that service instance."
```
EpsilonTal, line 24: the same for the description

EpsilonTal (changes requested):

schemas/predefined_plans.go
EpsilonTal: name, catalog_name, displayName etc...
description, translations.lang.description etc...

EpsilonTal: what do you think? 
we might consider using the JSON as template and format it..

EpsilonTal: consider maybe use a `BuildReferenceSchema` function instead of a constant so you will be able to format the schema with the additional constants of the name and description.

EpsilonTal: `func BuildReferencePlanSchema() string {
	return fmt.Sprintf(`{
  "name": "%[1]s",
  "catalog_name": "%[1]s",
  "description": "%[2]s",
  "bindable": true,
  "ready": true,
  "metadata": {
    "supportedPlatforms": [],
    "translations": {
      "en-US": {
        "displayName": "%[1]s",
        "description": "%[2]s"
      }
    }
  },
  "schemas": {
    "service_instance": {
      "create": {
        "parameters": {
          "$schema": "http://json-schema.org/draft-04/schema#",
          "type": "object",
          "additionalProperties": false,
          "_show_form_view": true,
          "properties": {
            "%[3]s": {
              "title": "Referenced Instance ID",
              "description": %[2]s",
              "_title": "TITLE_XTIT",
              "_description": "DESCRIPTION_XMSG",
              "type": "string",
              "minLength": 1,
              "maxLength": 100
            }
          }
        }
      }
    }
  }
}`, instance_sharing.ReferencePlanName, instance_sharing.ReferencePlanDescription, instance_sharing.ReferencedInstanceIDKey)
}`

EpsilonTal: that way we also use the `instance_sharing.ReferencedInstanceIDKey`

pkg/instance_sharing/types.go
```diff
@@ -4,11 +4,21 @@ const (
...
 
-	ReferencePlanNameSelector     = "plan_name_selector"
-	ReferenceInstanceNameSelector = "instance_name_selector"
-	ReferenceLabelSelector        = "instance_labels_selector"
+	ReferencedInstanceIDKey         = "referenced_instance_id"
+	ReferencedInstanceIDTitle       = "Reference Instance ID"
+	ReferencedInstanceIDDescription = "Specify the ID of the instance to which you want to refer."
+
+	ReferenceInstanceNameSelectorKey         = "instance_name_selector"
+	ReferenceInstanceNameSelectorTitle       = "Find by instance name"
```
i522006, line 14: These should all be title cases.
Find by Instance Name

pkg/instance_sharing/types.go
```diff
@@ -4,11 +4,21 @@ const (
...
+	ReferenceInstanceNameSelectorDescription = "You can use various search criteria to find shared instances to which you want to create a reference."
+
+	ReferencePlanNameSelectorKey         = "plan_name_selector"
+	ReferencePlanNameSelectorTitle       = "Find by plan name"
```
i522006, line 18: Find by Plan Name

pkg/instance_sharing/types.go
```diff
@@ -4,11 +4,21 @@ const (
...
-	ReferencePlanNameSelector     = "plan_name_selector"
-	ReferenceInstanceNameSelector = "instance_name_selector"
-	ReferenceLabelSelector        = "instance_labels_selector"
+	ReferencedInstanceIDKey         = "referenced_instance_id"
+	ReferencedInstanceIDTitle       = "Reference Instance ID"
+	ReferencedInstanceIDDescription = "Specify the ID of the instance to which you want to refer."
+
+	ReferenceInstanceNameSelectorKey         = "instance_name_selector"
+	ReferenceInstanceNameSelectorTitle       = "Find by instance name"
+	ReferenceInstanceNameSelectorDescription = "You can use various search criteria to find shared instances to which you want to create a reference."
```
i522006, line 15: This isn't what we agreed on.
It should be:
Create a reference to a shared instance that has the name you specified.

pkg/instance_sharing/types.go
```diff
@@ -4,11 +4,21 @@ const (
...
+
+	ReferencePlanNameSelectorKey         = "plan_name_selector"
+	ReferencePlanNameSelectorTitle       = "Find by plan name"
+	ReferencePlanNameSelectorDescription = "You can use various search criteria to find shared instances to which you want to create a reference."
```
i522006, line 19: Create a reference to a shared instance that was created with a specific plan. \nFor example, plan=large returns a shared instance that was created with the 'large' plan, and then you create a reference to that instance.

pkg/instance_sharing/types.go
```diff
@@ -4,11 +4,21 @@ const (
...
+
+	ReferenceLabelSelectorKey         = "instance_labels_selector"
+	ReferenceLabelSelectorTitle       = "Find by instance label"
```
i522006, line 22: Find by Instance Label

pkg/instance_sharing/types.go
```diff
@@ -4,11 +4,21 @@ const (
...
+	ReferenceLabelSelectorDescription = "You can use labels query to find shared instances to which you want to create a reference. For example: \"origin\": [\"eu\"] returns an instance whose origin is in the EU"
```
i522006, line 23: Create a reference to a shared instance that fulfills certain labeling criteria. For example \"origin\": [\"eu\"] returns a shared instance that has an 'origin' label whose value is 'EU'.

pkg/instance_sharing/types.go
i522006: Find by Instance Labels

pkg/instance_sharing/types.go
EpsilonTal: You can use the instance name to find a shared instance to which you want to create a reference.

pkg/instance_sharing/types.go
EpsilonTal: You can use the plan name to find the shared instance to which you want to create a reference.

pkg/instance_sharing/types.go
EpsilonTal: You can use the labels query to find the shared instance to which you want to create a reference. For example: \"origin\": [\"eu\"] returns an instance whose origin is in the EU.

pkg/instance_sharing/types.go
```diff
@@ -4,11 +4,21 @@ const (
...
-	ReferencedInstanceIDKey = "referenced_instance_id"
-
 	SupportsInstanceSharingKey = "supportsInstanceSharing"
 
-	ReferencePlanNameSelector     = "plan_name_selector"
-	ReferenceInstanceNameSelector = "instance_name_selector"
-	ReferenceLabelSelector        = "instance_labels_selector"
+	ReferencedInstanceIDKey         = "referenced_instance_id"
+	ReferencedInstanceIDTitle       = "Reference Instance ID"
+	ReferencedInstanceIDDescription = "Specify the ID of the instance to which you want to refer."
```
EpsilonTal, line 11: Specify the ID of the instance to which you want to create a reference.

schemas/predefined_plans.go
```diff
@@ -0,0 +1,98 @@
...
+              "description": "%[11]s",
+              "type": "string",
+              "minLength": 0,
+              "maxLength": 100
+            },
+            "%[12]s": {
+              "title": "%[13]s",
+              "description": "%[14]s",
+              "type": "object"
+            }
```
EpsilonTal, line 57: Waiting for uriel.even.sapir@sap.com review, to verify we can use an object input without properties.

schemas/predefined_plans.go
```diff
@@ -0,0 +1,98 @@
...
+  },
+  "schemas": {
+    "service_instance": {
+      "create": {
+        "parameters": {
+          "$schema": "http://json-schema.org/draft-04/schema#",
+          "type": "object",
+          "additionalProperties": false,
+          "_show_form_view": true,
+		  "_controlsOrder": ["%[3]s", "%[6]s", "%[9]s", "%[12]s"]
```
urieleve, line 30: comma is required at end of line

schemas/predefined_plans.go
```diff
@@ -0,0 +1,98 @@
...
+              "title": "%[10]s",
+              "description": "%[11]s",
+              "type": "string",
+              "minLength": 0,
+              "maxLength": 100
+            },
+            "%[12]s": {
+              "title": "%[13]s",
+              "description": "%[14]s",
+              "type": "object"
```
urieleve, line 56: ,"properties": {}
an Object property must have this fields

urieleve (changes requested):

schemas/predefined_plans.go
```diff
@@ -0,0 +1,98 @@
...
+  "metadata": {
+    "supportedPlatforms": []
+  },
+  "schemas": {
+    "service_instance": {
+      "create": {
+        "parameters": {
+          "$schema": "http://json-schema.org/draft-04/schema#",
+          "type": "object",
+          "additionalProperties": false,
```
urieleve, line 28: If we want the users  to add the additional label properties using the JSON display, then we should set additionalProperties=true, otherwise the validation in the cockpit will block this

schemas/predefined_plans.go
EpsilonTal: Thanks! fixed.

schemas/predefined_plans.go
EpsilonTal: we decided to remove the labels query for the moment.

schemas/predefined_plans.go
EpsilonTal: we will keep it false for now as we don't support any other keys.

EpsilonTal (approved):

i522006 (approved):

PR #686, opened 2021-07-13
Reference selectors with label query

EpsilonTal:
# [Instance Sharing] Claim shared service instance by labels

## Motivation

Sharing a service instance between different scopes allows apps in different scopes to share databases, messaging queues, and other types of services. This enables development teams to share resource, while eliminates the use service bindings as keys and user-provided services or other means to maintain credentials of instances which were provisioned in another scope.

## User story

As a developer, I would like to be able to refer to a shared instance without the need to provide the ID of the original instance so that I could apply a robust automated script that can be executed in different landscapes and setups

## Approach

Implemented like the service plan selector
* Retrieve in advance the service instances that match the criteria of the labels.
* Filter the labeled instances comparing to the full list of shared instances.

## Pull Request status

* [x] Initial implementation
* [x] Refactoring
* [x] Unit tests
* [x] Integration tests

### DIFF
--- a/api/common/sharing/utils.go
+++ b/api/common/sharing/utils.go
@@ -40,7 +40,7 @@ func ExtractReferencedInstanceID(req *web.Request, repository storage.Repository
 		selectors := parameters[instance_sharing.SelectorsKey].Map()
 		referencedInstance, err = filterInstancesBySelectors(ctx, repository, referencePlan.ServiceOfferingID, tenantIdentifier, getTenantId(), selectors, smaap)
 		if err != nil {
-			log.C(ctx).Errorf("Failed to filter instances by selectors: %s", err)
+			log.C(ctx).Errorf("Failed to filter instances by selectors: %s", err.Error())
 			return "", err
 		}
 
@@ -140,9 +140,8 @@ func filterInstancesBySelectors(ctx context.Context, repository storage.Reposito
 
 	// Filter by label query:
 	labelsSelector, labelExists := selectors[instance_sharing.ReferenceLabelSelectorKey]
-	labelsSelectorArray := labelsSelector.Array()
-	hasLabelsSelectors := labelExists && len(labelsSelectorArray) > 0 && labelsSelectorArray[0].String() != ""
-	if hasLabelsSelectors {
+	if labelExists && len(labelsSelector.Array()) > 0 && labelsSelector.Array()[0].String() != "" {
+		labelsSelectorArray := labelsSelector.Array()
 		var criteria []query.Criterion
 		for index := 0; index < len(labelsSelectorArray); index++ {
 			queryToParse := labelsSelectorArray[index].String()
--- a/api/common/sharing/utils.go
+++ b/api/common/sharing/utils.go
@@ -228,25 +228,6 @@ func getInstanceByLabelSelector(ctx context.Context, repository storage.Reposito
 	return instancesList.ItemAt(0).(*types.ServiceInstance), nil
 }
 
-func matchLabels(instanceLabels types.Labels, selectorLabels types.Labels) (bool, error) {
-	if instanceLabels == nil {
-		return false, nil
-	}
-	for selectorLabelKey, selectorLabelArray := range selectorLabels {
-		instanceLabelArray, exists := instanceLabels[selectorLabelKey]
-		if exists && len(instanceLabelArray) > 0 {
-			for _, selectorLabelVal := range selectorLabelArray {
-				if !contains(instanceLabelArray, selectorLabelVal) {
-					return false, nil
-				}
-			}
-		} else {
-			return false, nil
-		}
-	}
-	return true, nil
-}
-
 func isSameServiceOffering(ctx context.Context, repository storage.Repository, offeringID string, planID string) (bool, error) {
 	count, err := repository.Count(ctx, types.ServicePlanType,
 		query.ByField(query.EqualsOperator, "service_offering_id", offeringID),
@@ -256,13 +237,4 @@ func isSameServiceOffering(ctx context.Context, repository storage.Repository, o
 		return false, util.HandleStorageError(err, types.ServicePlanType.String())
 	}
 	return count > 0, nil
-}
-
-func contains(s []string, e string) bool {
-	for _, a := range s {
-		if a == e {
-			return true
-		}
-	}
-	return false
 }
--- a/api/common/sharing/utils.go
+++ b/api/common/sharing/utils.go
@@ -110,13 +110,8 @@ func validateParameters(parameters map[string]gjson.Result) error {
 	name, nameExists := parameters[instance_sharing.ReferenceInstanceNameSelectorKey]
 	plan, planExists := parameters[instance_sharing.ReferencePlanNameSelectorKey]
 	labels, labelsExists := parameters[instance_sharing.ReferenceLabelSelectorKey]
-	selectorLabels := types.Labels{}
-	if labelsExists {
-		if err := util.BytesToObject([]byte(labels.Raw), &selectorLabels); err != nil {
-			return err
-		}
-	}
-	hasAtLeastOneSelector := selectorLabels != nil && len(selectorLabels) > 0 || nameExists && len(name.String()) > 0 || planExists && len(plan.String()) > 0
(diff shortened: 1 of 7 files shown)


schemas/predefined_plans.go
```diff
@@ -42,19 +45,50 @@ func BuildReferencePlanSchema() string {
...
+      },
+      "update": {
+        "parameters": {
+          "$schema": "http://json-schema.org/draft-04/schema#",
+          "type": "object",
+          "_show_form_view": false,
+          "_controlsOrder": [],
+          "additionalProperties": false,
+          "properties": {}
+        }
```
EpsilonTal, line 91: this code prevents changing the instance parameters on the update flow.

schemas/predefined_plans.go
```diff
@@ -42,19 +45,50 @@ func BuildReferencePlanSchema() string {
...
-              "description": "%[11]s",
-              "type": "string",
-              "minLength": 0,
-              "maxLength": 100
+              "type": "object",
+              "_controlsOrder": [
+                "%[9]s",
+                "%[12]s",
+                "%[15]s"
+              ],
```
EpsilonTal, line 53: this is the wrapper object of the selectors parameters

pkg/instance_sharing/types.go
```diff
@@ -7,16 +7,22 @@ const (
...
 
 	ReferencePlanNameSelectorKey         = "plan_name_selector"
 	ReferencePlanNameSelectorTitle       = "Find by Plan Name"
-	ReferencePlanNameSelectorDescription = "You can use the plan name to find the shared instance to which you want to create a reference."
+	ReferencePlanNameSelectorDescription = "Specify the plan name of the shared instance to which you want to create a reference."
 
-	ReferenceLabelSelectorKey = "instance_labels_selector"
+	ReferenceLabelSelectorKey         = "instance_label_selector"
+	ReferenceLabelSelectorTitle       = "Find by Label Query"
+	ReferenceLabelSelectorDescription = "You can use the labels query to find the shared instance to which you want to create a reference. For example: \\\"origin\\\": [\\\"eu\\\"] returns an instance whose origin is in the EU."
```
EpsilonTal, line 27: strings are not final, waiting for @i522006.

api/common/sharing/utils.go
```diff
@@ -37,13 +37,8 @@ func ExtractReferencedInstanceID(req *web.Request, repository storage.Repository
...
-		if err != nil {
-			log.C(ctx).Errorf("Failed to retrieve shared instances: %s", err)
-			return "", err
-		}
-
-		referencedInstance, err = filterInstancesBySelectors(ctx, repository, sharedInstances, parameters, smaap, referencePlan.ServiceOfferingID)
+		selectors := parameters[instance_sharing.SelectorsKey].Map()
+		referencedInstance, err = filterInstancesBySelectors(ctx, repository, referencePlan.ServiceOfferingID, tenantIdentifier, getTenantId(), selectors, smaap)
 		if err != nil {
 			log.C(ctx).Errorf("Failed to filter instances by selectors: %s", err)
```
sigalmaya, line 43: In case of error, it will print the address of the error and the its details. 
You have 2 options use
1.  err.Error() - then it will give you the string 
2.  use %v in format

api/common/sharing/utils.go
```diff
@@ -106,21 +101,23 @@ func validateParameters(parameters map[string]gjson.Result) error {
...
+		validPlan := planExists && plan.Type == gjson.String && len(plan.String()) > 0
+		validLabels := labelsExists && labels.String() != "" && len(labels.Array()) > 0
+		hasAtLeastOneSelector = validLabels || validName || validPlan
 	}
-	hasAtLeastOneSelector := selectorLabels != nil && len(selectorLabels) > 0 || nameExists && len(name.String()) > 0 || planExists && len(plan.String()) > 0
-	selectorsAndID := IDExists && len(ID.String()) > 0 && hasAtLeastOneSelector
-	emptyValues := len(ID.String()) == 0 && !hasAtLeastOneSelector
-	if selectorsAndID || emptyValues {
-		return util.HandleInstanceSharingError(util.ErrInvalidReferenceSelectors, "")
+	if (validID && hasAtLeastOneSelector) || (!validID && !hasAtLeastOneSelector) {
```
sigalmaya, line 119: if selectors doesn't exist you pass null  in validateParameters function, then  in line 40  you do the following:
selectors := parameters[instance_sharing.SelectorsKey].Map() 
in this case it will crash

api/common/sharing/utils.go
```diff
@@ -37,13 +37,8 @@ func ExtractReferencedInstanceID(req *web.Request, repository storage.Repository
...
 		}
 	} else {
-		sharedInstances, err := getSharedInstances(ctx, repository, referencePlan.ServiceOfferingID, tenantIdentifier, getTenantId())
-		if err != nil {
-			log.C(ctx).Errorf("Failed to retrieve shared instances: %s", err)
-			return "", err
-		}
-
-		referencedInstance, err = filterInstancesBySelectors(ctx, repository, sharedInstances, parameters, smaap, referencePlan.ServiceOfferingID)
+		selectors := parameters[instance_sharing.SelectorsKey].Map()
```
sigalmaya, line 40: see the comment belowm the parameters[instance_sharing.SelectorsKey] could be null, you return null in validateParameters function

api/common/sharing/utils.go
```diff
@@ -135,11 +132,47 @@ func getSharedInstances(ctx context.Context, repository storage.Repository, offe
...
+func filterInstancesBySelectors(ctx context.Context, repository storage.Repository, offeringID, tenantIdentifier, tenantValue string, selectors map[string]gjson.Result, smaap bool) (*types.ServiceInstance, error) {
 	var err error
 	var filteredInstances types.ServiceInstances
 
-	planSelectorVal, planSelectorExists := parameters[instance_sharing.ReferencePlanNameSelectorKey]
+	var sharedInstances types.ObjectList
+
+	// Filter by label query:
+	labelsSelector, labelExists := selectors[instance_sharing.ReferenceLabelSelectorKey]
+	labelsSelectorArray := labelsSelector.Array()
```
sigalmaya, line 143: labelsSelector.Array() - what happens if the value doesn't exist ? should not you  check  labelExists before doing labelSelector.Array()

api/common/sharing/utils.go
EpsilonTal: thanks, using err.Error() instead.

api/common/sharing/utils.go
```diff
@@ -135,11 +132,47 @@ func getSharedInstances(ctx context.Context, repository storage.Repository, offe
...
+	hasLabelsSelectors := labelExists && len(labelsSelectorArray) > 0 && labelsSelectorArray[0].String() != ""
+	if hasLabelsSelectors {
+		var criteria []query.Criterion
```
sigalmaya, line 146: criterias

api/common/sharing/utils.go
EpsilonTal: if `selectorsExists=false`, 
then `hasAtLeastOneSelector=false`
so either you are using the `referenced_instance_id` key and won't enter this selectors flow.
or you will fail on the condition of line 119: `validID && hasAtLeastOneSelector) || (!validID && !hasAtLeastOneSelector`

I cover the scenario you mention with tests:
Context: `empty selectors object`

api/common/sharing/utils.go
EpsilonTal: can't be null, as explained there in the comment.

api/common/sharing/utils.go
EpsilonTal: criteria are already plural
criterion is it's singular

api/common/sharing/utils.go
EpsilonTal: fixing

api/common/sharing/utils.go
sigalmaya: ok

sigalmaya (approved):

api/common/sharing/utils.go
sigalmaya: ok

idobergerfroind (approved):

PR #692, opened 2021-07-25
allow empty credentials

sigalmaya:
with mtls out of box allow no credentials

### DIFF
--- a/api/filters/checkBrokerCredentials.go
+++ b/api/filters/checkBrokerCredentials.go
@@ -2,6 +2,7 @@ package filters
 
 import (
 	"fmt"
+	"github.com/Peripli/service-manager/pkg/httpclient"
 	"net/http"
 
 	"github.com/Peripli/service-manager/pkg/util"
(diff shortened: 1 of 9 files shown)


api/filters/checkBrokerCredentials.go
```diff
@@ -28,7 +29,8 @@ func (*CheckBrokerCredentialsFilter) Run(req *web.Request, next web.Handler) (*w
 	basicFields := gjson.GetManyBytes(req.Body, fmt.Sprintf(basicCredentialsPath, "username"), fmt.Sprintf(basicCredentialsPath, "password"))
 	tlsFields := gjson.GetManyBytes(req.Body, fmt.Sprintf(tlsCredentialsPath, "client_certificate"), fmt.Sprintf(tlsCredentialsPath, "client_key"))
 
-	if brokerUrl.Exists() && credentialsMissing(basicFields, tlsFields) {
+	httpSettings := httpclient.GetHttpClientGlobalSettings()
+	if brokerUrl.Exists() && (len(httpSettings.ServerCertificate) == 0 && credentialsMissing(basicFields, tlsFields)) {
```
nirbenrey, line 33: httpSettings.ServerCertificate is sm certificate?

pkg/types/credentials.go
```diff
@@ -22,24 +22,24 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"github.com/Peripli/service-manager/pkg/httpclient"
 )
 
 // Basic basic credentials
 type Basic struct {
-	Username string `json:"username,omitempty"`
-	Password string `json:"password,omitempty"`
```
nirbenrey, line 30: dosen't it means that when we return the broker (eg. on GET v2/service_brokers) the user will get empty strings? on username and password?

pkg/types/credentials.go
```diff
@@ -22,24 +22,24 @@ import (
...
+	Username string `json:"username"`
+	Password string `json:"password"`
 }
 
 type TLS struct {
-	Certificate string `json:"client_certificate,omitempty"`
-	Key         string `json:"client_key,omitempty"`
```
nirbenrey, line 35: dosen't it means that when we return the broker (eg. on GET v2/service_brokers) the user will get empty strings on certificate and key?

pkg/types/credentials.go
```diff
@@ -58,24 +58,53 @@ func (c *Credentials) MarshalJSON() ([]byte, error) {
...
+			if c.TLS.Key != "" && c.TLS.Certificate != "" {
+				_, err := tls.X509KeyPair([]byte(c.TLS.Certificate), []byte(c.TLS.Key))
+				if err != nil {
+					return errors.New("invalidate TLS configuration: " + err.Error())
+				}
+			}
+
+		}
+	} else {
+		if c.Basic != nil {
```
nirbenrey, line 89: i think you missed some cases:
1. user is switching from basic to tls when sm dose not have tls
2. user is switching from tls to basic when sm dose not have tls

pkg/types/credentials.go
```diff
@@ -58,24 +58,53 @@ func (c *Credentials) MarshalJSON() ([]byte, error) {
...
+			}
+			if c.Basic.Password != "" && c.Basic.Username == "" {
+				return errors.New("missing broker username")
+			}
 		}
-		if c.Basic.Password == "" {
-			return errors.New("missing broker password")
+		if c.TLS != nil {
+			if c.TLS.Certificate != "" && c.TLS.Key == "" {
+				return errors.New("missing certificate")
```
nirbenrey, line 75: is this error reflected to the user? if so, is it UA approved?

test/broker_test/broker_test.go
```diff
@@ -1222,6 +1222,69 @@ var _ = test.DescribeTestsFor(test.TestCase{
...
+									}
+									reply := ctx.SMWithOAuth.PATCH(web.ServiceBrokersURL + "/" + brokerIDWithMTLS).WithJSON(updatedBrokerJSON).
+										Expect()
+									reply.Status(http.StatusOK)
+									assertInvocationCount(brokerServerWithSMCertficate.CatalogEndpointRequests, 1)
+									byID := query.ByField(query.EqualsOperator, "id", brokerIDWithMTLS)
+									dbObj, err := repository.Get(context.TODO(), types.ServiceBrokerType, byID)
+									Expect(err).ToNot(HaveOccurred())
+									brokerFromDB := dbObj.(*types.ServiceBroker)
+									Expect(brokerFromDB.Credentials.Basic.Username).To(BeEmpty())
```
nirbenrey, line 1282: you should also validate what is returned from GET request. specially the username and password fields should not return on the broker object.

PR #700, opened 2021-08-12
change sm provided credentials level

sigalmaya:
move sm_provided_tls_credentials to tls object

### DIFF
--- a/pkg/types/credentials.go
+++ b/pkg/types/credentials.go
@@ -50,9 +50,6 @@ func (c *Credentials) MarshalJSON() ([]byte, error) {
 	if toMarshal.Basic == nil || toMarshal.Basic.Username == "" || toMarshal.Basic.Password == "" {
 		toMarshal.Basic = nil
 	}
-	if toMarshal.TLS == nil || toMarshal.TLS.Certificate == "" || toMarshal.TLS.Key == "" {
-		toMarshal.TLS = nil
-	}
 
 	return json.Marshal(toMarshal)
 }
--- a/pkg/types/credentials.go
+++ b/pkg/types/credentials.go
@@ -56,7 +56,7 @@ func (c *Credentials) MarshalJSON() ([]byte, error) {
 
 // Validate implements InputValidator and verifies all mandatory fields are populated
 func (c *Credentials) Validate() error {
-	if !c.BasicExists() && (*c.TLS == TLS{}) {
+	if !c.BasicExists() && (!c.TLSExists()) {
 		return errors.New("missing broker credentials: SM provided mtls , basic or tls credentials are required")
 	}
 	if c.BasicExists() {
@@ -145,6 +145,10 @@ func GenerateCredentials() (*Credentials, error) {
 func (c *Credentials) CertificateExists() bool {
 	return c.TLS != nil && !(c.TLS.Certificate == "" && c.TLS.Key == "")
 }
+func (c *Credentials) TLSExists() bool {
+	return c.TLS != nil && *c.TLS!= TLS{}
+}
+
 
 func (c *Credentials) BasicExists() bool {
 	return c.Basic != nil && *c.Basic != Basic{}
--- a/pkg/types/credentials.go
+++ b/pkg/types/credentials.go
@@ -106,14 +106,13 @@ func (c *Credentials) validateTLS() error {
 }
 
 func (c *Credentials) validateBasic() error {
-	if c.Basic.Username == "" || c.Basic.Password == "" {
-		if c.Basic.Username == "" {
-			return errors.New("missing broker username")
-		}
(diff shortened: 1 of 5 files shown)


pkg/types/credentials.go
```diff
@@ -146,7 +142,7 @@ func GenerateCredentials() (*Credentials, error) {
 }
 
 func (c *Credentials) TLSExists() bool {
-	return c.TLS != nil && *c.TLS != TLS{}
+	return c.TLS != nil && !(c.TLS.Certificate == "" && c.TLS.Key == "")
```
pavelmaliy, line 145: change it to CertificateExists

pkg/types/credentials.go
```diff
@@ -50,16 +50,13 @@ func (c *Credentials) MarshalJSON() ([]byte, error) {
 	if toMarshal.Basic == nil || toMarshal.Basic.Username == "" || toMarshal.Basic.Password == "" {
 		toMarshal.Basic = nil
 	}
-	if toMarshal.TLS == nil || toMarshal.TLS.Certificate == "" || toMarshal.TLS.Key == "" {
-		toMarshal.TLS = nil
-	}
```
pavelmaliy, line 53: why you removed it?

pavelmaliy: maybe like this?
if toMarshal.TLS == nil || (!UseSMCertificate && (toMarshal.TLS.Certificate == "" || toMarshal.TLS.Key == "" )) {
              toMarshal.TLS = nil
}

pavelmaliy (dismissed):

pkg/types/credentials.go
evyaffe: !(c.TLS.Certificate == "" && c.TLS.Key == "") --> (c.TLS.Certificate != "" || c.TLS.Key != "")
Shouldn't it be (c.TLS.Certificate != "" && c.TLS.Key != "")?
Unless I got something wrong here...

pavelmaliy: you right they both can't be empty, good catch
c.TLS.Certificate != "" && c.TLS.Key != ""

pkg/types/credentials.go
sigalmaya: but it is useless,  because we ahve also "omitmpty " on fields

pkg/types/credentials.go
sigalmaya: Why it is wrong? we did it on purpose(for the update case) : if c.CertificateExists() {
		err := c.validateTLS() 
and then in validate tls we check each field 

func (c *Credentials) validateTLS() error {
	if c.TLS.Certificate == "" || c.TLS.Key == "" {
		return errors.New("tls public certificate and key should be provided")
	}

	if c.TLS.UseSMCertificate {
		return errors.New("only one of the options could be set, SM provided credentials or tls")
	}
	_, err := tls.X509KeyPair([]byte(c.TLS.Certificate), []byte(c.TLS.Key))
	if err != nil {
		return errors.New("invalidate TLS configuration: " + err.Error())
	}

	return nil
}

pkg/types/credentials.go
```diff
@@ -32,16 +32,16 @@ type Basic struct {
 }
 
 type TLS struct {
-	Certificate string `json:"client_certificate,omitempty"`
-	Key         string `json:"client_key,omitempty"`
+	Certificate      string `json:"client_certificate,omitempty"`
+	Key              string `json:"client_key,omitempty"`
+	UseSMCertificate bool   `json:"sm_provided_tls_credentials,omitempty"`
```
evyaffe, line 37: the naming here is a bit odd, why not leave it SMProvidedTLSCredentials?
later in broker ToObject you use 'tls := &types.TLS{UseSMCertificate: e.SMProvidedCredentials}', so you need to match the strings, either change in the broker object or keep this one as is. 

WDYT?

sigalmaya: Eyal  told us  to use UseSMCertificate after we already agreed on API with xrs.  What do you think is better to use "useSMCerificate" or "SMProvidedTLSCredentials" i will align

pkg/types/credentials.go
pavelmaliy: you right I forgot it followed by validate call, 
you wanted to assure here that update leaves inconsistent data ignore my comment

evyaffe: agree, resolving

pkg/types/credentials.go
sigalmaya: it is used by JSON library.

pkg/types/credentials.go
pavelmaliy: please change it to SMProvidedTLSCredentials it's confusing.

PR #703, opened 2021-09-22
add context signature to request body in osb provision and bind

nirbenrey:
## Motivation

broker developers would like to verify that the context is originated in SM

## Approach

SM will add a 'signature' field to bind/provision broker requests. The field value will be the hashed & signed value of the context.
By using it the service brokers will be able to verify that the context was originated in SM.

## Pull Request status
* [x] Initial implementation
* [x] Integration tests


api/osb/context_signature.go
```diff
@@ -0,0 +1,119 @@
...
+	return ContextSignaturePluginName
+}
+
+func (s *ContextSignaturePlugin) Provision(req *web.Request, next web.Handler) (*web.Response, error) {
+	return s.sign(req, next)
+}
+
+func (s *ContextSignaturePlugin) Bind(req *web.Request, next web.Handler) (*web.Response, error) {
+	return s.sign(req, next)
+}
```
eyalnathan, line 44: UpdateService? we pass a context in update calls as well

api/osb/osb_store_plugin.go
```diff
@@ -816,12 +823,38 @@ func (sp *storePlugin) storeInstance(ctx context.Context, storage storage.Reposi
...
+	if _, exists := ctxMap["signature"]; exists {
+		delete(ctxMap, "signature")
+		outputData, err := json.Marshal(ctxMap)
+		if err != nil {
+			log.C(ctx).Errorf("failed to marshal context: %v", err)
+			return nil, err
+		}
+		rawContext = outputData
+	}
+	return rawContext, nil
```
eyalnathan, line 841: -> return sjson.DeleteBytes(rawContext, "signature") ?

nirbenrey: yes, thank you!! :)

api/osb/context_signature.go
nirbenrey: done

PR #727, opened 2021-12-05
donֿֿt add basic header if there are no username and password

sigalmaya:


pkg/client/broker_client.go
```diff
@@ -59,7 +59,7 @@ func (bc *BrokerClient) authAndTlsDecorator(requestHandler util.DoRequestWithCli
 		client := http.DefaultClient
 		ctx := req.Context()
 		logger := log.C(ctx)
-		if bc.broker.Credentials.Basic != nil && bc.broker.Credentials.Basic.Username != "" && bc.broker.Credentials.Basic.Password != "" {
+		if bc.broker.Credentials.BasicNotEmpty() {
 			bc.addBasicAuth(req)
```
pavelmaliy, line 63: why you're not using BasicExists()?

sigalmaya: BasicExists check  basic.username  has value OR basic.password has a value. I want to check that basis.username AND  basic.password has value. I want to avoid bearer tokens as "Bearer myusername:" or "Bearer :mypassword" 
func (c *Credentials) BasicExists() bool {
	return c.Basic != nil && *c.Basic != Basic{}
}

pavelmaliy (approved):

PR #756, opened 2022-03-27
avoid failing the request if sign fail

I065450:
# Pull Request Template

## Prerequisites

## Motivation

## Approach

## Pull Request status

* [x] Initial implementation
* [ ] Refactoring
* [ ] Unit tests
* [ ] Integration tests

## Third-party code

## Work in Progress label

### DIFF
--- a/test/common/context_signature.go
+++ b/test/common/context_signature.go
@@ -192,7 +192,7 @@ func ProvisionInstanceAndVerifySignature(ctx *TestContext, brokerServer *BrokerS
 	return instanceID
 }
 
-func ProvisionInstance(ctx *TestContext, brokerServer *BrokerServer, provisionFunc func() string) string {
+func ProvisionInstanceWithoutSignature(ctx *TestContext, brokerServer *BrokerServer, provisionFunc func() string) string {
 	brokerServer.ServiceInstanceHandler = GetVerifyContextInvalidKeyHandlerFunc()
 
 	instanceID := provisionFunc()
--- a/test/common/context_signature.go
+++ b/test/common/context_signature.go
@@ -89,6 +89,29 @@ func GetVerifyContextHandlerFunc(publicKeyStr string) func(http.ResponseWriter, 
 	}
 }
 
+func GetVerifyContextInvalidKeyHandlerFunc() func(http.ResponseWriter, *http.Request) {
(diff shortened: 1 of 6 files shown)


test/common/context_signature.go
```diff
@@ -168,3 +191,18 @@ func ProvisionInstanceAndVerifySignature(ctx *TestContext, brokerServer *BrokerS
 
 	return instanceID
 }
+
+func ProvisionInstance(ctx *TestContext, brokerServer *BrokerServer, provisionFunc func() string) string {
```
nirbenrey, line 195: please rename to a more precise name. eg. ProvisionInstanceWithoutSignature

test/service_instance_and_binding_test/context_signature/context_signature_test.go
```diff
@@ -101,4 +101,26 @@ var _ = Describe("context signature verification tests", func() {
...
+		instanceID := "signed-ctx-instance-invalid"
+		var provisionFunc func() string
+		BeforeEach(func() {
+			ctx = common.NewTestContextBuilderWithSecurity().WithEnvPostExtensions(func(e env.Environment, servers map[string]common.FakeServer) {
+				e.Set("api.osb_rsa_private_key", "invalidKey")
+			}).Build()
+			ctxSetup()
+		})
+
+		It("request should be sent without the signature", func() {
```
nirbenrey, line 115: The "context" and "it"  should form a readable story. 
eg. 
context("when the private key is invalid", func(){
    it("should not fail the request"/"should send the request without the signature")
})
then when you read it in the logs you can easily understand what is being tested.
In contrast to "Invalid private key. it request should be sent without the signature"

test/service_instance_and_binding_test/context_signature/suit_test.go
```diff
@@ -47,6 +47,10 @@ var _ = BeforeSuite(func() {
 		e.Set("api.osb_rsa_public_key", publicKeyStr)
 		e.Set("api.osb_successor_rsa_public_key", publicSuccessorKeyStr)
 	}).Build()
+	ctxSetup()
+})
+
+func ctxSetup() {
```
nirbenrey, line 53: this function does not setup the ctx, it registers a service broker. please rename it accordingly.

PR #773, opened 2022-06-24
Return broker credentials in response with sensitive data hidden

evyaffe:
## Motivation

SM API returns broker response without credentials. Instead, return the credentials sanitized to get a better indication on how brokers authentication is configured.


pkg/types/service_broker.go
```diff
@@ -69,7 +72,43 @@ func (e *ServiceBroker) GetTLSConfig(logger *logrus.Entry) (*tls.Config, error)
 }
 
 func (e *ServiceBroker) Sanitize(context.Context) {
-	e.Credentials = nil
+	isSmTLS := false
```
nirbenrey, line 75: isServiceManagerTLS?

evyaffe: changed

nirbenrey (approved):

kerenlahav (approved):

PR #781, opened 2022-08-11
I320494  introduce redis

nirbenrey:
## Motivation

The Service Manager rate limiter is implemented in memory. This means that the actual limit is #SM-APPS multiplied by the configured limit.

## Approach
Use Redis as a distributed cache layer for the rate limiting.


api/rate_limiter.go
```diff
@@ -93,19 +96,35 @@ func parseRateLimiterConfiguration(input string) ([]RateLimiterConfiguration, er
...
+		}
+	}
+
 	configurations, err := parseRateLimiterConfiguration(options.APISettings.RateLimit)
 	if err != nil {
 		return nil, err
 	}
 	for _, configuration := range configurations {
+		store := redisStore
+		if store == nil {
```
evyaffe, line 121: this log will be logged X number of rate limiters, we can log it once above the loop

api/rate_limiter.go
```diff
@@ -93,19 +96,35 @@ func parseRateLimiterConfiguration(input string) ([]RateLimiterConfiguration, er
...
+			log.C(ctx).Error("redis client is not initialized. creating in memory store for rate limiting")
```
evyaffe, line 122: it's the same log for all rate limiters, let's log it once above the loop

evyaffe (approved):

